PR #28, opened 2025-03-13
docs: add glama badge. Fix typos in README.md

phernandez:


phernandez:
Fixed in 5d5efa02. schema_infer and schema_diff now return formatted markdown text instead of raw Pydantic models, matching schema_validate's existing _format_validation_report pattern. The 'undefined' rendering was caused by LLMs trying to interpret raw JSON tool output.

PR #103, opened 2025-04-16
fix: preserve Chinese characters in permalinks

andyxinweiminicloud:
## Problem
The current implementation converts all Unicode characters to ASCII using `unidecode`, which causes Chinese characters in permalinks to be transliterated. This makes it difficult for Chinese users to recognize the permalinks, as they no longer match the original Chinese titles.

## Solution
Modified the `generate_permalink` function in `utils.py` to preserve non-ASCII characters while still properly handling ASCII characters:

1. Removed the Unicode to ASCII transliteration step using `unidecode`
2. Updated character processing to only lowercase ASCII letters while preserving non-ASCII characters
3. Modified the regex pattern to retain Chinese character ranges in permalinks
4. Kept all other functionality (camelCase handling, replacing underscores, etc.) intact

This change allows Chinese characters to be preserved in permalinks, making them more intuitive and readable for Chinese users while maintaining compatibility with existing systems.

## Testing
Successfully tested with various Chinese file paths and titles. Both file paths and permalinks correctly preserve Chinese characters while still properly processing English text.

### DIFF
--- a/src/basic_memory/utils.py
+++ b/src/basic_memory/utils.py
@@ -35,7 +35,7 @@ def generate_permalink(file_path: Union[Path, str, PathLike]) -> str:
 
     Returns:
         Normalized permalink that matches validation rules. Converts spaces and underscores
-        to hyphens for consistency.
+        to hyphens for consistency. Preserves non-ASCII characters like Chinese.
 
     Examples:
         >>> generate_permalink("docs/My Feature.md")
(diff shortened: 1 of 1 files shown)


phernandez:
claude: is this fix safe for permalinks?

phernandez:
Hi @andyxinweiminicloud thanks for the fix. I'm going to merge this PR, then add a bit of logic to handle permalinks with unicode for latin characters and keep your fixes for Chinese characters.

PR #150, opened 2025-06-19
fix: fastmcp deprecation warning

groksrc:
## Summary
- Fixes deprecation warning that appears when running `bm --version` or `basic-memory --version`
- Moves `log_level` parameter from FastMCP constructor to the `run()` method as recommended by fastmcp

## Changes
- Remove `log_level="DEBUG"` from FastMCP constructor in `server.py`
- Add `log_level="DEBUG"` to both `mcp_server.run()` calls in `mcp.py`

## Test
After this change, running `bm --version` no longer shows the deprecation warning:
```
Providing `log_level` when creating a server is deprecated. Provide it when calling `run` or as a global setting instead.
```

## Context
This warning was discovered after installing basic-memory v0.13.7 via Homebrew.

### DIFF
--- a/src/basic_memory/cli/commands/mcp.py
+++ b/src/basic_memory/cli/commands/mcp.py
@@ -78,7 +78,7 @@ def mcp(
     if transport == "stdio":
         mcp_server.run(
             transport=transport,
-            log_level="DEBUG",
+            log_level="INFO",
         )
     elif transport == "streamable-http" or transport == "sse":
         mcp_server.run(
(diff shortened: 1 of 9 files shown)


src/basic_memory/cli/commands/mcp.py
```diff
@@ -78,11 +78,13 @@ def run_file_sync():
...
             transport=transport,
+            log_level="DEBUG",
         )
     elif transport == "streamable-http" or transport == "sse":
         mcp_server.run(
             transport=transport,
             host=host,
             port=port,
             path=path,
+            log_level="DEBUG",
```
phernandez, line 89:
```suggestion
            log_level="INFO",
```

PR #140, opened 2025-06-18
fix: UNIQUE constraint failed: entity.permalink issue #139

groksrc:
Fixes #139

Improved error handling in create_entity_from_markdown() to properly handle permalink conflicts when different files generate the same permalink.

## Changes
- Distinguish between file_path and permalink constraint failures
- For same file conflicts: update existing entity
- For different file conflicts: generate unique permalink with suffix
- Added comprehensive test case reproducing the exact issue scenario

This fixes the issue where write_note would fail when trying to create notes with titles that generate the same permalink as existing notes.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #160, opened 2025-06-20
fix: add configurable kebab-case filename format to resolve forward slash inconsistency

phernandez:
## Summary

- Fixes inconsistency between filename generation and permalink format for titles with forward slashes
- Adds configurable filename formatting with kebab-case as default
- Maintains backward compatibility for existing users

## Test plan

- [ ] Run existing permalink formatting tests
- [ ] Test new sanitize_filename function
- [ ] Test Entity consistency with both config options
- [ ] Verify write_note tool works with forward slash titles


src/basic_memory/schemas/base.py
```diff
@@ -187,10 +187,20 @@ class Entity(BaseModel):
     @property
     def file_path(self):
         """Get the file path for this entity based on its permalink."""
+        # Import here to avoid circular dependency
+        try:
+            from basic_memory.config import app_config
+            use_kebab_case = app_config.filename_format == "kebab-case"
+        except ImportError:
+            # Fallback to original behavior if config not available
+            use_kebab_case = False
```
groksrc, line 196: This should probably be 
```suggestion
            use_kebab_case = True
```

phernandez:
This probably relates to #140

PR #161, opened 2025-06-20
feat: enhance move_note tool with cross-project detection and guidance

### LINKED ISSUE #157: [BUG] Move note to different project fails
## Bug Description

A user reported an issue asking the AI to move a file to a new project. The ai attempted the operation, but it was not clear that it did not complete successfully. 

## Expected Behavior
We don't currently support moving files to a new project. The AI assistant should provide clear feedback to the user. 

## Actual Behavior
The move_note tool should
- return a clear message explaining that notes can't be moved 
- prompt the assistant to recreate the note in the new project

or the we should add support for moving a note to a new project directly in the code. This is problematic because we would have to pass two project values to the move note operation and we would need an api change. 

Image

phernandez:
- Add cross-project move detection to provide clear error messages when users attempt to move notes between different projects
- Detect project names in destination paths and provide specific guidance
- Detect project-like keywords and offer general guidance for potential  cross-project moves
- Provide step-by-step workflow for proper cross-project moves: read_note → switch_project → write_note → delete_note
- Add comprehensive integration tests for cross-project detection scenarios
- Ensure normal within-project moves continue to work without interference

Resolves #157

### DIFF
--- a/src/basic_memory/mcp/tools/move_note.py
+++ b/src/basic_memory/mcp/tools/move_note.py
@@ -7,9 +7,151 @@ from loguru import logger
 
 from basic_memory.mcp.async_client import client
 from basic_memory.mcp.server import mcp
-from basic_memory.mcp.tools.utils import call_post
+from basic_memory.mcp.tools.utils import call_post, call_get
 from basic_memory.mcp.project_session import get_active_project
 from basic_memory.schemas import EntityResponse
+from basic_memory.schemas.project_info import ProjectList
+
+
+async def _detect_cross_project_move_attempt(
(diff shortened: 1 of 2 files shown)


groksrc:
@phernandez Take a look at this, Claude seems to think this is right but I'm not sure this is what we want. If the file is moved from project `testing` to project `main` shouldn't it be searchable in `main`?

[pr161_analysis.md](https://github.com/user-attachments/files/20876757/pr161_analysis.md)

groksrc (approved):
I'm good on this one now, works as expected.

:shipit:

PR #174, opened 2025-06-25
fix: respect BASIC_MEMORY_HOME environment variable in Docker containers

### LINKED ISSUE #173: [BUG] Dockerfile volume mount path mismatch with config.json
## Bug Description
The Dockerfile creates a volume mount at /app/data but the Basic Memory config.json is configured to look for data in /root/basic-memory, causing a path mismatch that prevents the application from finding its data directory.

## Steps To Reproduce
1. Build the Docker image from the current Dockerfile
2. Run the container with a volume mount: `docker run -v mydata:/app/data basic-memory`
3. Basic Memory will look for data in /root/basic-memory instead of /app/data
4. Application fails to access persisted data

## Expected Behavior
Basic Memory should use the /app/data directory that is defined as a VOLUME in the Dockerfile.

## Actual Behavior
Basic Memory looks for data in /root/basic-memory, ignoring the Docker volume mount at /app/data.

## Environment
- Docker version: Various
- Basic Memory version: Latest
- Installation method: Docker

## Additional Context
The Dockerfile contains:
```dockerfile
VOLUME ["/app/data"]
```

But the config.json uses:
```json
{
  "default_project_path": "/root/basic-memory"
}
```

## Possible Solution
Either:
1. Update the Dockerfile to set VOLUME to /root/basic-memory
2. Update the config.json to use /app/data as the default_project_path
3. Add an environment variable override for the data path in Docker contexts

Co-Authored-By: Claude <noreply@anthropic.com>

groksrc:
Fixes the Docker path mismatch issue where the Dockerfile sets `BASIC_MEMORY_HOME=/app/data` but the configuration system ignored it and used `/root/basic-memory` instead.

## Changes
- Updated `BasicMemoryConfig` to respect the `BASIC_MEMORY_HOME` environment variable
- Applied fix in both the `projects` field default and `model_post_init` method
- Maintains backward compatibility by falling back to the original behavior when the env var is not set

Closes #173

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -15,9 +15,9 @@ class TestBasicMemoryConfig:
         """Test that config uses default path when BASIC_MEMORY_HOME is not set."""
         # Ensure BASIC_MEMORY_HOME is not set
         monkeypatch.delenv("BASIC_MEMORY_HOME", raising=False)
-        
+
         config = BasicMemoryConfig()
-        
+
         # Should use the default path (home/basic-memory)
         expected_path = str(config_home / "basic-memory")
         assert config.projects["main"] == expected_path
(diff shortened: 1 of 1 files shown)


groksrc:
@claude can we add a test for this?

groksrc:
@claude the tests you added in the previous commit failed due to path errors relating to permissions. In principle the tests are good, but they need to pass the CI.

PR #176, opened 2025-06-26
fix: restart watch service when project configuration changes

### LINKED ISSUE #156: [BUG] New project is not synced when config changes (new project added)
A user reported an issue where they added a file to a new project, but the new file was not synced. I believe this project was created via a chat session. I'm guessing that since the watch process runs at startup, it did not get restarted when the config was changed, and therefore the new project wasn't syncing. 

Can you confirm this behavior? If it is the case, how difficult would it be to restart the sync when the config changes?

phernandez:
Resolves issue where new projects created via MCP tools weren't being watched by the file synchronization service. The watch service now automatically restarts when projects are added, removed, or when the project configuration changes.

Implementation:
- Add file-based signaling mechanism for watch service restarts
- ProjectService signals restart when projects are added/removed
- WatchService checks for restart signal on each file change batch
- Watch loop recreates service instances to pick up new projects
- Add comprehensive tests for the restart signal mechanism

Note: New projects will only be picked up for watching after the next file change occurs in any existing project, as the restart signal is checked during file change processing.

Fixes #156


phernandez:
I'm going to close this PR and reimplement it.

Issue #219, opened 2025-07-10
[BUG] BASIC_MEMORY_HOME env var is not respected


phernandez:
## Bug Description

Configuring the BASIC_MEMORY_HOME env var does not set the root dir for basic memory

Example:
```
"basic-memory": {
      "command": "uvx",
      "args": ["basic-memory", "mcp"],
      "env": {
        "BASIC_MEMORY_HOME": "~/project/.basic-memory"
      }
    }
```
files are still created in ~/basic-memory

phernandez:
@claude why isn't the env var working in the config.py?

jope-bm:
Oh Claude, at least you tried...

@phernandez @groksrc 

The root issue here is that the FileService object that actually writes the files to disk, uses the value of the path field in the project table in the database to decide where to write the file. The value in the Config is not used.  We can't just change the value of the path in the database, because then any existing files would not longer be able to be found.

So, one possible way forward is to:

1. Remove the BASIC_MEMORY_HOME env var and the config
2. Create a cli command for changing the root path of a BM project which will update the path for the project in the database, ~~and move all the notes for all the projects to the new project dir.~~
    1. On paul's suggestion, just prompt the user to move the files from the old location to their new location (BM cli just updates the database)
3. Since project paths are no longer in the config file, it might be helpful if we had a tool or cli command to show the projects and their file system locations, since that would no longer be available in the config file.

phernandez:
Nice. the project info is available via

```
➜  ~ bm project list
                                          Basic Memory Projects
┏━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┓
┃ Name             ┃ Path                                                             ┃ Default ┃ Active ┃
┡━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━┩
│ main             │ ~/basic-memory                                                   │ ✓       │ ✓      │
│ personal         │ ~/Library/Mobile Documents/iCloud~md~obsidian/Documents/personal │         │        │
│ chat-test        │ ~/chat-test                                                      │         │        │
[...]

PR #239, opened 2025-07-24
fix: basic memory home env var not respected when project path is changed.

jope-bm:
Addresses https://github.com/basicmachines-co/basic-memory/issues/219

This PR adds a new CLI command to move a project's location.  The BASIC_MEMORY_HOME envvar is only used to set the project directory at startup if the `main` project does not already exist.

Command Syntax:

```
basic-memory project move main /my/new/project/path
```

This command will lookup a project's path in the config and update it there, then sync the new path in the BM database. It **will not** move your project's files and sub-directories. You will need to do that after the tool has run.  BM will then create new notes, read notes, etc for the projects in the new location.

Additionally, this PR 

* Fixes a bug with config loading, so the values in the config file are not overwritten if an error occurs while reading the config file.
* Re-enables writing to log file.

### DIFF
--- a/src/basic_memory/services/project_service.py
+++ b/src/basic_memory/services/project_service.py
@@ -309,6 +309,47 @@ class ProjectService:
             # MCP components might not be available in all contexts
             logger.debug("MCP session not available, skipping session refresh")
 
+    async def move_project(self, name: str, new_path: str) -> None:
+        """Move a project to a new location.
+        
+        Args:
+            name: The name of the project to move
+            new_path: The new absolute path for the project
+            
+        Raises:
+            ValueError: If the project doesn't exist or repository isn't initialized
+        """
+        if not self.repository:
+            raise ValueError("Repository is required for move_project")
+        
+        # Resolve to absolute path
+        resolved_path = os.path.abspath(os.path.expanduser(new_path))
+        
+        # Validate project exists in config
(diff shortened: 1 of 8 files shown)


src/basic_memory/services/project_service.py
```diff
@@ -309,6 +309,47 @@ async def synchronize_projects(self) -> None:  # pragma: no cover
...
+        
+        Args:
+            name: The name of the project to move
+            new_path: The new absolute path for the project
+            
+        Raises:
+            ValueError: If the project doesn't exist or repository isn't initialized
+        """
+        if not self.repository:
+            raise ValueError("Repository is required for move_project")
```
phernandez, line 323: This looks like Claude doing goldplating BS. We should always have a repository. But whatever.

phernandez (approved):
One big thing that goes along with this is that now that we have updated the config, we need to stop watching the old project dir and start watching the new one. I think that's outside the scope of this PR, just saying. 

There is a PR where I half implemented this, but I didn't love it https://github.com/basicmachines-co/basic-memory/pull/176

jope-bm:
> One big thing that goes along with this is that now that we have updated the config, we need to stop watching the old project dir and start watching the new one. I think that's outside the scope of this PR, just saying.
> 
> There is a PR where I half implemented this, but I didn't love it #176

👍

PR #242, opened 2025-07-29
fix(mcp): support --project flag

715d:
This PR makes it so that the `mcp` sub-command respects the `--project` flag.

For example you can now do `basic-memory --project foo mcp` and expect the MCP to start up with the current project already set to `foo`. 

I'm making use of this feature by having a wrapper script that Claude uses to launch the MCP automatically setting the `--project` flag to the correct name based on the git repo Claude was launched from. 

I find this much more reliable than hoping that Claude automatically switches to the right project (or remembering to explicitly tell it every time).

### DIFF
--- a/src/basic_memory/cli/commands/mcp.py
+++ b/src/basic_memory/cli/commands/mcp.py
@@ -35,12 +35,24 @@ def mcp(
     - sse: Server-Sent Events (for compatibility with existing clients)
     """
 
+    from basic_memory.config import get_project_config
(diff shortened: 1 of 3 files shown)


groksrc:
@715d Thanks for the PR. We're holding this for internal reasons for the next release but we will look at getting it merged soon.

phernandez:
This will be implemented in https://github.com/basicmachines-co/basic-memory/pull/298 and released in v0.15.0

PR #252, opened 2025-08-13
Addressed issues when running basic-memory on the Windows platform

manuelbliemel:
...which are
- In contrast to Linux where newlines contain \n on Windows newlines contain \r\n
- Normalized path separators to posix '/'

Tests were succeeding with my changes based on commit b0cc5594 but tests are failing on current HEAD of main and therefore also on this branch after merging main.


phernandez:
@claude can you do a code review on this PR and also suggest a workflow where we can test in a windows env during our CI/CD flow?

PR #259, opened 2025-08-20
fix: Enable string-to-integer conversion for build_context depth parameter

jope-bm:
## Summary
- Add support for passing string values to the depth parameter in the build_context MCP tool
- Automatically converts valid numeric strings to integers while raising clear errors for invalid inputs
- Maintains backward compatibility with existing integer parameters

## Changes Made
- Updated type annotation from `Optional[int]` to `Optional[Union[int, str]]`
- Added string validation and conversion logic with proper error handling
- Enhanced test coverage for both MCP and CLI interfaces

## Test Plan
- [x] Existing tests continue to pass (backward compatibility)
- [x] New test for valid string depth parameter conversion
- [x] New test for invalid string depth parameter error handling
- [x] CLI interface tests for string parameter handling
- [x] Manual verification that build_context accepts string depth parameters

## Verification
The fix has been tested and verified to work correctly:
- String depth parameters like "3" are converted to integer 3
- Invalid strings like "invalid" raise clear ToolError messages
- All existing functionality remains unchanged

### DIFF
--- a/src/basic_memory/mcp/tools/build_context.py
+++ b/src/basic_memory/mcp/tools/build_context.py
@@ -1,6 +1,6 @@
 """Build context tool for Basic Memory MCP server."""
 
-from typing import Optional
+from typing import Optional, Union
 
 from loguru import logger
 
@@ -35,7 +35,7 @@ from basic_memory.schemas.memory import (
 )
 async def build_context(
     url: MemoryUrl,
-    depth: Optional[int] = 1,
+    depth: Optional[Union[int, str]] = 1,
     timeframe: Optional[TimeFrame] = "7d",
     page: int = 1,
     page_size: int = 10,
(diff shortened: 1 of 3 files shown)


src/basic_memory/mcp/tools/build_context.py
```diff
@@ -35,7 +35,7 @@
 )
 async def build_context(
     url: MemoryUrl,
-    depth: Optional[int] = 1,
+    depth: Optional[Union[int, str]] = 1,
```
phernandez, line 38: In my experience, adding more types to the mcp args will sometimes mess up tools like vscode, but lets give this a shot. It's more "correct". FYI, I prefer `Optional[int | str]` or even a type alias
```python
type StringOrInt = int | str
depth: Optional[StringOrInt]
```

phernandez (approved):

src/basic_memory/mcp/tools/build_context.py
phernandez: Cool

PR #267, opened 2025-08-22
fix: Ensure datetime serialization includes timezone info for RFC 3339 compliance

### LINKED ISSUE #253: [BUG] Validation error on datetime
## Bug Description
When using `build_context` search results contain datetimes specified as a "date-time" type in the output schema.
These data does not adhere to the "ISO 8601 / RFC 3339 style date-time strings (with timezone)".

## Steps To Reproduce
1. Install version 0.14.2
2. Run command 'npx @modelcontextprotocol/inspector uvx basic-memory --project [WHATEVER] mcp'
3. Use tool/feature 'build_context'

## Expected Behavior
created_at: "2025-08-12T13:44:45.344889" should be 2025-08-12T13:44:45.344889Z with a "Z" appended if no time zone information is available.

## Actual Behavior
Validation Error: data.results[0].primary_result.created_at should match format "date-time"

created_at: "2025-08-12T13:44:45.344889"

## Environment
- OS: [Windows 11]
- Python version: [e.g. 3.12.9]
- Basic Memory version: [e.g. 0.14.2]
- Installation method: uv

## Additional Context

Image

## Possible Solution
Introduce a utility function like this:
```typescript
def to_rfc3339(dt: Union[datetime, str]) -> datetime:
    """Converts a datetime object or ISO 8601 string to a timezone-aware UTC datetime, suitable for RFC 3339."""
    if isinstance(dt, str):
        dt = datetime.fromisoformat(dt)

    if dt.tzinfo is None:
        # Assume UTC if no timezone info
        return dt.replace(tzinfo=timezone.utc)
    return dt.astimezone(timezone.utc)
```
and use in search_repository.py and context_service.py

jope-bm:
Add field_serializer methods to all datetime fields in memory schemas to ensure proper timezone-aware datetime serialization. This fixes validation errors in MCP tools like build_context where datetime strings were missing timezone information.

- Added ensure_timezone_aware utility function
- Updated EntitySummary, RelationSummary, ObservationSummary, MemoryMetadata
- All datetime fields now serialize with timezone suffix (Z or +00:00)
- Handles both timezone-aware and naive datetimes by converting to UTC

Fixes #253


phernandez (approved):
I thought pydandic did this for us?

jope-bm:
> I thought pydandic did this for us?

If the source datetime is naive (no tzinfo), it is serialized without Z or timezone. Which, of course, makes sense.

So maybe the real issue is that build_context is creating datetimes without timezone? I'm not sure how we'd be getting an ISO string without tzinfo otherwise.

phernandez:
🤷🏽 it could also be a platform thing, or some weird mcp tool thing. Whatever it takes.

jope-bm:
@phernandez This is what Claude tells me:

```
Yes, the build_context tool is returning dates without timezone info.

  Issue found:
  - generated_at: Returns dates with timezone info (e.g., "2025-08-22T15:59:44.292047Z")
  - timeframe: Returns dates without timezone info (e.g., "2025-08-21T09:59:44.287982")

  Root cause:
  In src/basic_memory/services/context_service.py:157, the timeframe is set using:
  timeframe=since.isoformat() if since else None,

  The since datetime object comes from parse_timeframe() which uses the dateparser library but doesn't preserve timezone
  information, so .isoformat() produces a timestamp without the timezone suffix.

  Fix needed:
  The parse_timeframe() function should return timezone-aware datetimes, or the context service should add timezone info before
  calling .isoformat().
```

Issue #263, opened 2025-08-21
[BUG] build_context fails schema validation in MCP client (missing fields / date-time format)


se1961:
## Bug Description

Calling `basic-memory.build_context` with a valid `memory://…` URI fails schema validation in an MCP client. Other endpoints (`search_notes`, `read_note`) work as expected.

## Steps To Reproduce

1. Resolve a note:

   - Title: `Techniques Elsa Can't Use`
   - URI returned by `search_notes`: `memory://conversations/20250429-Techniques_Elsa_Cant_Use`

2. Call:

   ```json
   {"url":"memory://conversations/20250429-Techniques_Elsa_Cant_Use"}
   ```

via basic-memory.build_context

## Expected Behavior

Response conforms to the tool’s declared output schema; client can render context.

## Actual Behavior

Client error:

MCP error -32602: Structured content does not match the tool's output schema:
data.results[0].primary_result.created_at should match format "date-time",
data.results[0].primary_result should have required property 'relation_type',
data.results[0].primary_result should have required property 'category',
data.results[0].primary_result should match some schema in anyOf

## Environment

- Basic Memory: v0.14.3 (installed via `uv`)
- Server command: `uv tool run basic-memory mcp`
- OS/CPU: macOS (Apple Silicon)
- MCP client: ChatWise (stdio; tools auto-run)
- Project path: `/Users/shelley/basic-memory`
- Fresh DB after `basic-memory reset`; ~84 notes indexed

## Additional Context

- Reproduces consistently after reset and full app restart.

- search_notes and read_note succeed for the same memory:// target.
- Same behavior across different models (provider kept constant per thread).

**Client config (for completeness)**

{
  "mcpServers": {
    "basic-memory": {
      "command": "uv",
      "args": ["tool", "run", "basic-memory", "mcp"],
      "transport": "stdio"
    }
  }
}

**Workaround**

Temporarily avoid build_context and emulate context with search_notes → read_note → related notes (1–2).

phernandez:
@claude can you investigate this?

se1961:
Update: Retested on git@main via:
uvx --from "git+https://github.com/basicmachines-co/basic-memory@main" basic-memory mcp

In ChatWise (macOS), after deleting/re-adding the MCP entry and restarting, 
`build_context` still fails schema validation. First-line error indicates 
issues with `created_at` (date-time) and a `…type` field. 

Other endpoints (`search_notes`, `read_note`) continue to work.

[...]

PR #270, opened 2025-08-22
fix: Add ISO datetime serialization to MCP schema models

jope-bm:
Fixes MCP client schema validation failures for build_context tool by ensuring datetime fields are serialized as ISO format strings.

MCP clients expect "date-time" format validation, which requires datetime objects to be serialized as ISO strings (e.g. "2023-12-08T10:30:00") rather than Python datetime objects.

Changes:
- Added ConfigDict with json_encoders to EntitySummary, RelationSummary, ObservationSummary, and MemoryMetadata models
- Ensures datetime fields serialize as dt.isoformat() strings

Resolves persistent schema validation issues reported in #263

### DIFF
--- a/src/basic_memory/schemas/memory.py
+++ b/src/basic_memory/schemas/memory.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from typing import List, Optional, Annotated, Sequence, Literal, Union
 
 from annotated_types import MinLen, MaxLen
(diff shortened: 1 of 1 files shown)


jope-bm:
@claude you need to add unit tests to verify that your fix actually fixes the problem and you didn't break anything else.

Issue #171, opened 2025-06-25
[BUG] File extension removed when renaming files


groksrc:
## Bug Description
When renaming a file using Basic Memory, the ".md" extension is frequently removed, making the file unrecognizable both in Finder and Obsidian. This forces users to manually add the extension back each time.

## Steps To Reproduce
1. Have Basic Memory installed
2. Use the rename/move functionality to rename a markdown file
3. The ".md" extension is removed from the filename
4. File becomes unrecognizable in Finder and Obsidian

## Expected Behavior
When renaming a file, the file extension should be preserved. If renaming "document.md" to "new-document", the result should be "new-document.md".

## Actual Behavior
The file extension is removed during the rename operation. Files lose their ".md" extension and become unrecognizable to both the file system (Finder) and Obsidian.

Additionally, when attempting to fix this issue:
- Basic Memory sometimes reports that all files have been modified when many still need fixing
- Files in subfolders are sometimes completely skipped

## Environment
- OS: [Unknown - reported by Discord user]
- Python version: [Unknown]
- Basic Memory version: [Unknown]
- Installation method: [Unknown]
- Claude Desktop version (if applicable): [Unknown]

## Additional Context
- User has found a workaround by asking Basic Memory to add the extension back, but this process is unreliable
- The issue affects file recognition in both Finder and Obsidian
- Problem extends to batch operations where some files are skipped or incorrectly reported as fixed

## Possible Solution
The rename/move functionality should:
1. Preserve file extensions by default
2. Handle file extensions separately from the filename when processing renames
3. Ensure batch operations correctly process all files including those in subfolders

groksrc:
@claude take a look at this and see if you can identify the issue causing it.

groksrc:
@claude double check your answer here. I have a suspicion that this happens when a filename without the file extension is passed.

groksrc:
@claude let's have move_note raise if there's no file extension provided. All of the examples in the file have a file extension so one has been assumed since the beginning. We can also update the docs if it will help the LLM understand how to use the tool correctly without throwing.

PR #281, opened 2025-08-29
fix: move_note without file extension

groksrc:
Fix for #171 

`move_note` will now return a soft error to the LLM if it tries to move a note without a file extension.

You probably want to turn off whitespace for the diff, the actual change is on L443.


src/basic_memory/mcp/tools/move_note.py
```diff
@@ -402,28 +415,51 @@ async def move_note(
...
+            - `projects/meeting-2025.md`
+            - `archive/old-document.md`
+
+            ## Try again with extension:
+            ```
+            move_note("{identifier}", "{destination_path}.md")
+            ```
+
+            All examples in Basic Memory expect file extensions to be explicitly provided.
+            """).strip()
```
phernandez, line 461: I feel like we should probably assert that the destination path suffix matches the original file path suffix. It doesn't have to be just markdown.

phernandez: yes, but the move did not fail. The file still got moved, correct? So I think this will confuse the LLM. We should either do or not do the move. If there is no extension in on the destination in the tool call we should error before doing the operation

phernandez:
@claude lets update this branch and make sure a fix is implemented. 

The problem we are trying to solve is that when the AI assistant performs a `move_note`, the file suffix should be preserved.

phernandez:
@claude we need to resolve the conflicts so we can merge into main

Issue #285, opened 2025-09-02
[BUG] Basic_memory MCP server needs to use .gitignore


FellowTraveler:
# Basic Memory MCP Server - .gitignore Bug Report

## Summary
The basic_memory MCP server fails to respect `.gitignore` patterns and attempts to sync build artifacts, causing sync failures on Rust projects with complex incremental compilation filenames.

### Expected Behavior
The basic_memory MCP server should:
1. Read and respect all patterns in the project's `.gitignore` file
2. Skip directories and files that match `.gitignore` patterns
3. Never attempt to sync build artifacts (e.g., `target/`, `node_modules/`, `dist/`, `build/`)
4. Handle complex filenames gracefully without crashing

### Actual Behavior
The server:
1. Attempts to sync files in `backend/target/` despite it being in `.gitignore`
2. Fails with file move errors on Rust's incremental compilation artifacts
3. Reports sync failures that block other operations

## Error Message
```
❌ Sync failed for project 'ai-robot-security': Cannot move entity from 
'backend/target/debug/incremental/angel_security-3jhmxh0kgxo7h/s-hadf7wssc0-1pd28gt-2u4rvv79i4alxt84s5c72zkiy/5w0grk96adxa0x4j2z8bqoffm.o' to 
'backend/target/debug/deps/angel_security-db91db90048569b2.5w0grk96adxa0x4j2z8bqoffm.rcgu.o': 
destination path is already occupied by another file.
```

## Reproduction Steps
1. Create a Rust project with basic_memory configured
2. Add `backend/target/` to `.gitignore` (or `target/` for a standard Rust project)
3. Run `cargo build` to generate build artifacts
4. Attempt to sync with basic_memory
5. Observe sync failure on incremental compilation files

## Project Structure
```
/Users/au/Documents/idea_AI_robot_drone_security/
├── .gitignore (contains "backend/target/" on line 8)
├── backend/
│   ├── Cargo.toml
│   ├── src/
│   └── target/           # ← Should be ignored but isn't
│       └── debug/
│           ├── incremental/
│           └── deps/
├── frontend/
└── memory-bank/
```

## .gitignore Content (Relevant Lines)
```gitignore
# Rust Backend
backend/target/
backend/Cargo.lock
backend/**/*.rs.bk
backend/**/*.pdb
```

### Primary Issues:
1. **Gitignore Parsing**: The server is not properly parsing or applying `.gitignore` patterns
[...]

PR #286, opened 2025-09-03
Fix to make basic-memory ignore stuff in the .gitignore file

FellowTraveler:
Seems to be working now, for me.

https://github.com/basicmachines-co/basic-memory/issues/285

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ __pycache__/
 .coverage
 htmlcov/
 
+*.db
+
 # Distribution / packaging
 .Python
 build/
--- a/src/basic_memory/cli/commands/status.py
+++ b/src/basic_memory/cli/commands/status.py
@@ -129,8 +129,9 @@ async def run_status(verbose: bool = False):  # pragma: no cover
     app_config = ConfigManager().config
     config = get_project_config()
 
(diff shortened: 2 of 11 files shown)


FellowTraveler:
FYI -- there's still a problem here, search_index table. Fixing now. Do not merge yet.

UPDATE: Fixed now. Works for me.

phernandez:
Hi @FellowTraveler , I'd like to add this PR, but some of the stuff in here won't work. The notion of per project databases is not something we can add. There is only one database, shared for all projects in ~/.basic-memory.

phernandez:
We will incorporate .gitignore logic in https://github.com/basicmachines-co/basic-memory/pull/314

FellowTraveler:
> Hi @FellowTraveler , I'd like to add this PR, but some of the stuff in here won't work. The notion of per project databases is not something we can add. There is only one database, shared for all projects in ~/.basic-memory.

The problem (that I ran into quite quickly) was that when I am coding in more then one project at the same time, it would fail, since the database was already opened by another project. I made the commit to fix this issue that I was running into. Maybe I'm using it wrong?

phernandez:
@FellowTraveler  What platform are you running on?

Issue #172, opened 2025-06-25
[BUG] Folder structure navigation problem - files saved to wrong directory


groksrc:
## Bug Description
When saving or creating notes, Basic Memory often ignores the specified folder path and places files in the main project directory instead. The app requires extremely precise path specifications to correctly locate the intended folder, which creates unnecessary friction in the workflow.

## Steps To Reproduce
1. Use Basic Memory to create or save a note
2. Specify a folder path for the file location
3. The file is saved to the main project directory instead of the specified folder
4. Multiple attempts with different path formats may be needed to get correct placement

## Expected Behavior
When a folder path is specified for saving or creating notes, Basic Memory should reliably place the file in the specified folder. Common path formats should be accepted and interpreted correctly.

## Actual Behavior
Files are frequently saved to the main project directory regardless of the specified folder path. Users must use extremely precise path specifications to get files saved to the intended location, creating workflow friction.

## Environment
- OS: [Unknown - reported by multiple Discord users]
- Python version: [Unknown]
- Basic Memory version: [Unknown]
- Installation method: [Unknown]
- Claude Desktop version (if applicable): [Unknown]

## Additional Context
- This issue has been reported by multiple users (including user Alex)
- The problem affects the basic workflow of organizing notes into folders
- Users are forced to manually move files or retry with different path formats

## Possible Solution
The path resolution logic should:
1. Be more flexible in accepting common path formats
2. Properly resolve relative and absolute paths
3. Provide clear feedback when a path cannot be resolved
4. Default to creating folders if they don't exist (with user confirmation)

groksrc:
@claude take a look a this and see if you can identify the issue causing it.

groksrc:
@claude take a look at #154 and #160 and see if these are related.

phernandez:
@claude these are the related links. They were PRs, not issues

https://github.com/basicmachines-co/basic-memory/issues/154
https://github.com/basicmachines-co/basic-memory/pull/160

bdmayes:
@groksrc Do we happen to have any exact reports of this? I don't know who is encountering this and/or which OS they are on. It may be a Windows-specific bug?

In any case, I opened the following PR to include better sanitization of the folder string in the entity. I also converted an f-string to use `os.path.join()` instead. All of these things should be safer but I have no way of verifying it if it resolves the problem without knowing a concrete set of inputs. Should I ask folks on Discord to see if anyone can offer exact reproduction steps?

[...]

PR #292, opened 2025-09-10
fix: Sanitize folder names and properly join paths

bdmayes:
Fix for https://github.com/basicmachines-co/basic-memory/issues/172 (probably)

- Sanitizes folder names when creating entities.
- Uses `os.path.join()` rather than an f-string that separates folder and filename by the `"/"` character.
- Adds tests for the folder sanitization and an integration test for invoking the `write_note` tool.


bdmayes:
Interesting. I thought that I created my tests in such a way that they would work across platforms. I'll make sure to fix them such that they also work on Windows.

PR #421, opened 2025-11-11
fix: main project always recreated on project list command

bdmayes:
- Fixes a bug in config where after removing the "main" project, invoking the `project list` command would recreate it.
- Adds an integration test to properly confirm that "main" is not recreated.


phernandez:
thanks @bdmayes Whenever we get the tests passing, we'll merge this and push a patch release.

phernandez (approved):

PR #429, opened 2025-11-12
fix: Use platform-native path separators in config.json

### LINKED ISSUE #428: Windows: config.json writes paths with forward slashes instead of backslashes
## Problem

On Windows, the `config.json` file stores project paths using forward slashes (`/`) instead of the platform-native backslashes (`\`). This causes issues when the paths are used by Windows tools and utilities that expect backslash separators.

Example config.json on Windows:
```json
{
  "projects": {
    "main": "C:/Users/username/basic-memory"
  }
}
```

Expected on Windows:
```json
{
  "projects": {
    "main": "C:\\Users\\username\\basic-memory"
  }
}
```

## Root Cause

The code uses `.as_posix()` in multiple places which forces POSIX-style (Unix) path separators regardless of platform:

**`src/basic_memory/config.py`:**
- Line 66: Default project initialization
- Line 201: `model_post_init` method  
- Line 364: `add_project` method

```python
# Example from line 364
config.projects[name] = project_path.as_posix()  # ❌ Forces forward slashes
```

`.as_posix()` converts all paths to use forward slashes `/`, which is not correct for Windows.

## Solution

Replace `.as_posix()` with `str(path)` to preserve platform-native path separators:

```python
# Use platform-native separators
config.projects[name] = str(project_path)  # ✅ Uses \\ on Windows, / on Unix
```

On Windows, this will correctly use backslashes. On Unix systems, it will use forward slashes.

## Affected Files

- `src/basic_memory/config.py` - Three locations using `.as_posix()`

## Testing

- Need tests to verify paths use correct separators on Windows
- Verify existing Unix behavior is unchanged
- Test that paths work correctly with Windows tools

## Impact

This affects Windows users when:
- Creating new projects
- Reading/writing config.json
- Integrating with Windows-native tools that expect backslash paths

groksrc:
## Problem

Fixes #428

On Windows, the `config.json` file stores project paths using forward slashes (`/`) instead of platform-native backslashes (`\`). This causes issues when paths are used by Windows tools that expect backslash separators.

**Example issue on Windows:**
```json
{
  "projects": {
    "main": "C:/Users/username/basic-memory"  // ❌ Wrong on Windows
  }
}
```

**Expected on Windows:**
```json
{
  "projects": {
    "main": "C:\\Users\\username\\basic-memory"  // ✅ Correct
  }
}
```

## Root Cause

The code used `.as_posix()` in three locations, which forces POSIX-style (Unix) forward slash separators regardless of platform:

- Line 66: Default project initialization  
- Line 201: `model_post_init` method
- Line 364: `add_project` method

```python
# ❌ Before: Forced forward slashes
config.projects[name] = project_path.as_posix()

# ✅ After: Uses platform-native separators
config.projects[name] = str(project_path)
```

## Solution

Replaced `.as_posix()` with `str()` to preserve platform-native path separators:
- **Windows**: Uses backslashes (`\`)
- **Unix**: Uses forward slashes (`/`)

## Changes

1. **src/basic_memory/config.py**
   - Fixed 3 locations using `.as_posix()`
   - Changed to use `str(path)` for platform-native separators

2. **tests/test_config.py**
   - Added 3 new tests for platform-specific path handling
   - Tests verify correct separator usage on both Windows and Unix
   - All 27 tests pass (24 existing + 3 new)

## Testing

```bash
uv run pytest tests/test_config.py -v
# 27 passed in 1.54s
```

**Test Coverage:**
- `test_project_paths_use_platform_native_separators_in_config` - Verifies saved config uses correct separators
- `test_add_project_uses_platform_native_separators` - Tests ConfigManager.add_project()
- `test_model_post_init_uses_platform_native_separators` - Tests default project creation

## Impact

This fix ensures:
- ✅ Windows users get properly formatted paths in config.json
- ✅ Unix systems continue to work as before  
- ✅ Paths work correctly with platform-native tools
- ✅ No breaking changes for existing functionality

## Related

- Issue #428: Windows config.json path separator issue

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -81,8 +81,8 @@ class TestBasicMemoryConfig:
 
         config = BasicMemoryConfig()
 
-        # Should use the exact value from environment variable
-        assert config.projects["main"] == relative_path
+        # Should normalize to platform-native path format
+        assert Path(config.projects["main"]) == Path(relative_path)
 
     def test_basic_memory_home_overrides_existing_main_project(self, config_home, monkeypatch):
         """Test that BASIC_MEMORY_HOME is not used when a map is passed in the constructor."""


src/basic_memory/config.py
```diff
@@ -63,7 +63,7 @@ class BasicMemoryConfig(BaseSettings):
 
     projects: Dict[str, str] = Field(
         default_factory=lambda: {
-            "main": Path(os.getenv("BASIC_MEMORY_HOME", Path.home() / "basic-memory")).as_posix()
+            "main": str(Path(os.getenv("BASIC_MEMORY_HOME", Path.home() / "basic-memory")))
```
jope-bm: is this still using a forward slash between Path.home() and "basic-memory"?

groksrc: When you use / between Path objects (or between a Path and a string), it creates a new path by joining them together. So Path.home() / "basic-memory" is equivalent to joining the user's home directory with the "basic-memory" subdirectory.

This is syntactic sugar that pathlib provides to make path construction more readable. It's roughly equivalent to:
`pythonPath.home().joinpath("basic-memory")`

PR #509, opened 2026-01-13
fix: explicitly set external_id for new entities in async context

TylerJNewman:
## Summary

Fixes `NOT NULL constraint failed: entity.external_id` errors when creating new entities via `write_note` tool.

## Problem

SQLAlchemy's `default=callable` is an INSERT-time default, not an instantiation-time default. In async contexts with aiosqlite, this INSERT-time default isn't reliably invoked before the INSERT statement executes.

This bug was introduced after PR #485 added `external_id` to the Entity model with:
```python
external_id: Mapped[str] = mapped_column(String, unique=True, default=lambda: str(uuid.uuid4()))
```

## Solution

Explicitly set `external_id = str(uuid.uuid4())` when creating new Entity objects in `entity_model_from_markdown()`, ensuring the field is always populated regardless of the async execution context.

## Testing

Verified fix works by creating notes via:
- CLI: `bm tool write-note --title "Test" --folder test --content "Test"`
- MCP: `write_note` tool from Claude Code

Both methods now work without the NOT NULL constraint error.

---

### DIFF
--- a/src/basic_memory/markdown/utils.py
+++ b/src/basic_memory/markdown/utils.py
@@ -1,5 +1,6 @@
 """Utilities for converting between markdown and entity models."""
 
+import uuid
 from pathlib import Path
 from typing import Any, Optional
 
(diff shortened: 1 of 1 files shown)


phernandez:
Hi @TylerJNewman thanks. we'll look at this. FYI, it looks like some of your memories files got added to this commit. LMK and we can delete it if you want to resubmit

phernandez:
This is fixed in https://github.com/basicmachines-co/basic-memory/pull/513

PR #510, opened 2026-01-14
fix: Docker container Python symlink broken at runtime

groksrc:
The venv Python symlink pointed to /root/.local/share/uv/python/... which only existed during build. At runtime, the container crashed with "ModuleNotFoundError: No module named 'encodings'".

Fix by setting UV_PYTHON_INSTALL_DIR=/python so Python is installed to a persistent location, and explicitly install Python 3.13.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,13 @@ ARG GID=1000
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
 # Set environment variables
+# UV_PYTHON_INSTALL_DIR ensures Python is installed to a persistent location
+# that survives in the final image (not in /root/.local which gets lost)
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,9 +24,10 @@ RUN (getent group ${GID} || groupadd --gid ${GID} appgroup) && \
 # Copy the project into the image
 ADD . /app
 
-# Sync the project into a new environment, asserting the lockfile is up to date
+# Install Python 3.13 explicitly and sync the project
 WORKDIR /app
-RUN uv sync --locked
+RUN uv python install 3.13
+RUN uv sync --locked --python 3.13
```
phernandez: lets use python 3.14. It is faster (apparently), and I went through the code with a recent PR and upgraded BM to support it. Its in our test matrix, so it works.

phernandez (approved):
Looks good. python 3.14 is optional

PR #568, opened 2026-02-14
feat: support ?project= query parameter on memory:// URIs

groksrc:
## Summary

- Allow `memory://` URIs to embed a project hint via standard query parameters, e.g. `memory://specs/search?project=research`
- This lets AI assistants resolve the correct project from the URI alone, avoiding multi-project discovery round-trips that waste tokens and pollute context
- Backward-compatible: existing URIs without `?` continue to work identically

## Changes

- **`schemas/memory.py`**: Remove `?` from invalid chars, add `parse_memory_url()` helper, update `normalize_memory_url()` to preserve query params, update `memory_url_path()` to strip them
- **`build_context.py`**, **`read_note.py`**, **`read_content.py`**: Extract `?project=` from URL when no explicit `project` parameter was provided
- **Tests**: 10 new test cases covering parsing, normalization, path extraction, and edge cases (duplicates, empty values)

## Motivation

When working with multiple Basic Memory projects, an AI assistant receiving `memory://foo` doesn't know which project it lives in. It has to call `list_memory_projects()` and potentially try several projects, wasting tokens and context window. With `?project=`, the URI is self-describing: `memory://foo?project=research` routes directly.

## Test plan

- [x] All 32 memory URL validation tests pass (including 10 new)
- [x] Full unit test suite passes (1702 passed, 0 failures)
- [x] 100% coverage on `schemas/memory.py`
- [x] Lint clean (`ruff check` passes)

### DIFF
--- a/src/basic_memory/mcp/tools/delete_note.py
+++ b/src/basic_memory/mcp/tools/delete_note.py
@@ -7,6 +7,7 @@ from mcp.server.fastmcp.exceptions import ToolError
 
 from basic_memory.mcp.project_context import get_project_client
 from basic_memory.mcp.server import mcp
+from basic_memory.schemas.memory import parse_memory_url
 
 
 def _format_delete_error_response(project: str, error_message: str, identifier: str) -> str:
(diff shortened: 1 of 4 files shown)


phernandez:
Hmm. I think this PR that I just merged
https://github.com/basicmachines-co/basic-memory/pull/544

will do the same thing, at least that is the intent. 

New format: 
```
memory://<project>/<path> (project encoded in the URL authority/path shape)
```
old still works: 
```
memory://<path> <-- legacy/default-project fallback
```

This is how I almost did it to begin with (and should have). 

If you think we need the project as url param for a different case, lets talk.

groksrc:
Oh cool, I like your way better but I didn't realize it was already done. I proposed this way because I wasn't sure about breaking permalinks. Let's defer to yours and we can reopen this if we want later.

PR #630, opened 2026-02-27
fix: skip workspace resolution when client factory is active

jope-bm:
## Summary
- When the cloud MCP server runs in-process with `set_client_factory()`, `get_project_client()` was still entering the cloud workspace resolution path for cloud-mode projects
- This called `get_cloud_control_plane_client()` which hit the production API at `/workspaces/`, failing with 401 since the in-process server has no credentials to call itself externally
- Added an early `is_factory_mode()` check that short-circuits before workspace resolution — the transport layer has already resolved workspace and tenant context

## Test plan
- [x] Added `test_factory_mode_skips_workspace_resolution` — sets up a factory client, patches `resolve_workspace_parameter` and `get_cloud_control_plane_client` to fail if called, verifies neither is invoked
- [x] All 25 existing `test_project_context.py` tests pass
- [x] Verified fix resolves 401 errors on `read_note`, `search_notes`, and `list_directory` MCP tools in cloud server

### DIFF
--- a/src/basic_memory/mcp/project_context.py
+++ b/src/basic_memory/mcp/project_context.py
@@ -419,6 +419,7 @@ async def get_project_client(
         _explicit_routing,
         _force_local_mode,
         get_client,
+        is_factory_mode,
     )
 
     # Step 1: Resolve project name from config (no network call)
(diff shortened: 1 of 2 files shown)


phernandez:
@claude could we also check
BASIC_MEMORY_CLOUD_MODE = "true"

here instead of this weird factor_mode notion?

PR #668, opened 2026-03-14
fix(mcp): add project detection from memory:// URLs in edit_note and delete_note

### LINKED ISSUE #664: [BUG]
Title: edit_note and delete_note missing detect_project_from_url_prefix — incorrect cloud routing
 
---
 
## Bug Description
 
`edit_note` and `delete_note` are missing the `detect_project_from_url_prefix()` call that `read_note`, `build_context`, `search_notes`, and `read_content` all have. This causes incorrect routing (potentially to cloud instead of local) when `project=None` is passed and no `default_project` is configured.
 
## Version
 
basic-memory v0.20.2
 
## Steps to Reproduce
 
1. Configure basic-memory with multiple projects but **no** `default_project` set
2. Call `edit_note(identifier="memory://myproject/some-note", operation="append", content="test")` **without** explicit `project` param
3. Observe that routing falls through to `resolve_project_parameter(None)` → cloud API discovery
 
## Expected Behavior
 
`edit_note` should detect the project from the `memory://` URL prefix (like `read_note` does) and route locally.
 
## Root Cause
 
In `read_note.py` (lines 136-140), before calling `get_project_client()`:
 
```python
if project is None:
    detected = detect_project_from_url_prefix(identifier, ConfigManager().config)
    if detected:
        project = detected
```
 
This pattern exists in:
- ✅ `read_note.py` (line 136)
- ✅ `build_context.py` (line 189)
- ✅ `search.py` (line 522)
- ✅ `read_content.py` (line 215)
 
But is **missing** from:
- ❌ `edit_note.py` (line 258 goes directly to `get_project_client(project, ...)`)
- ❌ `delete_note.py` (line 225 goes directly to `get_project_client(project, ...)`)
 
Note: `write_note` is not affected because it takes `title` and `directory` as separate params rather than an `identifier` that could contain a `memory://` URL prefix.
 
## Suggested Fix
 
For `edit_note.py` and `delete_note.py`, add the missing imports:
 
```python
from basic_memory.config import ConfigManager
from basic_memory.mcp.project_context import detect_project_from_url_prefix
```
 
And add before `async with get_project_client(...)`:
 
```python
if project is None:
    detected = detect_project_from_url_prefix(identifier, ConfigManager().config)
    if detected:
        project = detected
```
 
This is identical to the pattern in `read_note.py`.
 
## Impact
 
- Users with `default_project` set in config are partially shielded (the default catches the `None`)
- Users without `default_project` or with multi-project setups relying on URL-prefix routing will hit cloud API discovery failures or incorrect routing
- The inconsistency between read-path tools (fixed) and write-path tools (missing) is confusing

phernandez:
## Summary
- `edit_note` and `delete_note` were missing `detect_project_from_url_prefix()` that other tools (`read_note`, `build_context`, `search_notes`, `read_content`) already use
- This caused incorrect cloud routing when using `memory://` URLs without an explicit `project` parameter in multi-project setups
- Added the same 4-line detection pattern before `get_project_client()` in both tools

Fixes #664

## Test plan
- [ ] `just fast-check` passes (lint, typecheck, impacted tests)
- [ ] `pytest tests/mcp/test_tool_edit_note.py tests/mcp/test_tool_delete_note.py -v` passes
- [ ] Verify multi-project routing works with `memory://project-name/note` URLs in edit_note and delete_note

### DIFF
--- a/src/basic_memory/mcp/tools/delete_note.py
+++ b/src/basic_memory/mcp/tools/delete_note.py
@@ -224,7 +224,11 @@ async def delete_note(
         commands and alternative formats to try.
     """
     # Detect project from memory URL prefix before routing
-    if project is None:
+    # Trigger: identifier starts with memory:// and no explicit project was provided
+    # Why: only gate on memory:// to avoid misrouting plain paths like "research/note"
+    #      where "research" is a directory, not a project name
+    # Outcome: project is set from the URL prefix, routing goes to the correct project
+    if project is None and identifier.strip().startswith("memory://"):
         detected = detect_project_from_url_prefix(identifier, ConfigManager().config)
         if detected:
             project = detected
--- a/src/basic_memory/mcp/tools/edit_note.py
+++ b/src/basic_memory/mcp/tools/edit_note.py
@@ -261,7 +261,11 @@ async def edit_note(
     effective_replacements = expected_replacements if expected_replacements is not None else 1
 
     # Detect project from memory URL prefix before routing
-    if project is None:
+    # Trigger: identifier starts with memory:// and no explicit project was provided
+    # Why: only gate on memory:// to avoid misrouting plain paths like "research/note"
+    #      where "research" is a directory, not a project name
+    # Outcome: project is set from the URL prefix, routing goes to the correct project
+    if project is None and identifier.strip().startswith("memory://"):
         detected = detect_project_from_url_prefix(identifier, ConfigManager().config)
         if detected:
             project = detected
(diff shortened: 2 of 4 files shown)


phernandez:
Addressed review feedback in c5530194:

1. **Gated detection to `memory://` prefix only** — `detect_project_from_url_prefix()` now only triggers when the identifier starts with `memory://`. Plain paths like `research/archive` will no longer misroute to a project named `research`.

2. **Added 6 tests** covering:
   - Detection works with `memory://` URLs (edit_note + delete_note)
   - Detection is NOT called for plain path identifiers
   - Detection is NOT called when project is explicitly provided

PR #759, opened 2026-04-20
fix(sync): constrain watch service to --project scope

### LINKED ISSUE #758: fix(sync): MCP --project constraint ignored by watch service causes race conditions with concurrent processes
## Summary

When multiple `basic-memory mcp --project <name>` processes run concurrently (a normal setup when the user has different projects wired to different MCP clients), each process spawns its own file watcher over **every** active project — not just the one it was constrained to. The result is N watchers racing on the same files, producing cascade-delete crashes, tmp-file rename collisions, and phantom "entity not found" errors from MCP tools.

## Repro

Configure two or more local projects in `~/.basic-memory/config.json`, then run two MCP processes in parallel:

```bash
basic-memory mcp --project project-a &
basic-memory mcp --project project-b &
```

Edit a file in `project-a`. Both watchers see the change. Both call `sync_file` / `handle_delete` / `update_frontmatter` against the same path.

## Root cause

At `src/basic_memory/services/initialization.py:113`, the `BASIC_MEMORY_MCP_PROJECT` env var filters *only* the initial one-shot background sync. The watch service is then started separately and reloads projects every cycle at `src/basic_memory/sync/watch_service.py:178` from `project_repository.get_active_projects()` — with no constraint applied. So the `--project` flag narrows startup sync but leaves the long-running watcher unrestricted.

## Symptoms observed in user log

Pulled from a 5.8 MB log covering ~24h of real usage with four projects and many MCP processes. Key signals:

**Duplicated change handling.** Five `Processing project: video-annotator changes, change_count=2` log lines within 50 ms; the same file path deleted four times in parallel by separate watch cycles.

**Crash 1: `AttributeError: 'NoneType' object has no attribute 'permalink'`** at `src/basic_memory/models/knowledge.py:303` in `Relation.permalink`:

```python
from_permalink = self.from_entity.permalink or self.from_entity.file_path
```

`to_entity` is guarded on the next line; `from_entity` is not. When one watcher cascade-deletes an entity while another is still resolving relation permalinks for an adjacent entity, `from_entity` comes back `None`.

**Crash 2: `[Errno 2] No such file or directory: 'foo.tmp' -> 'foo.md'`** from `src/basic_memory/file_utils.py:128`. `write_file_atomic` uses `path_obj.with_suffix('.tmp')` — a deterministic per-file tmp path. Two watchers write to the same `foo.tmp`, first one renames to `foo.md`, second rename fails because the tmp was already moved.

**Crash 3: `Entity with external_id ... not found`** from MCP `edit_note`. Tool resolves an entity, concurrent watcher deletes it, PATCH 404s.

## Fix

Pass the constraint into `WatchService` so the watch cycle observes the same subset the background sync was told to sync. Concretely:

1. `WatchService.__init__` accepts `constrained_project: Optional[str]`.
2. The per-cycle project load (`watch_service.py:178`) filters by that value if set.
3. `initialize_file_sync` reads `BASIC_MEMORY_MCP_PROJECT` once and wires it through to both the background sync filter and the new `WatchService` kwarg.

This makes `basic-memory mcp --project X` actually mean "this process watches X," so N concurrent MCP processes produce N disjoint watchers instead of N overlapping ones.

Follow-ups (separate issues, not blocking):

- Guard `from_entity` in `Relation.permalink` the same way `to_entity` is guarded — defensive against any future path where relations outlive their source entity.
- Make `write_file_atomic` use a unique tmp suffix (pid + uuid or `tempfile.mkstemp(dir=...)`) so intra-process concurrent writers to the same path don't collide either.

phernandez:
## Summary

Closes #758. When `basic-memory mcp --project X` launched multiple processes (a common multi-project setup), each spawned a watch service over **every** active project — not just the constrained one. N concurrent MCP processes produced N duplicate watchers racing on the same files.

The `BASIC_MEMORY_MCP_PROJECT` env var filtered the one-shot background sync, but not the long-running watcher. This PR threads the constraint into `WatchService` so `--project` actually scopes the watcher.

## Symptoms this fixes

Observed in a user log covering ~24h of real usage with four projects:

- `AttributeError: 'NoneType' object has no attribute 'permalink'` in `Relation.permalink` — one watcher cascade-deleted an entity while another was resolving relation permalinks on an adjacent entity
- `[Errno 2] No such file or directory: 'foo.tmp' -> 'foo.md'` — two watchers raced on the deterministic tmp path in `write_file_atomic`
- `Entity with external_id ... not found` in MCP `edit_note` — concurrent watcher deleted the entity mid-operation

## Changes

- `WatchService.__init__` accepts `constrained_project: Optional[str]`
- Extracted the per-cycle project filter into `_select_projects_to_watch()` so it is testable outside the `run()` loop (which remains `# pragma: no cover`)
- `initialize_file_sync` reads `BASIC_MEMORY_MCP_PROJECT` once and wires it through to both the background sync filter and the new `WatchService` kwarg
- Log message reflects actual scope (constrained project name vs "all projects")

## Test plan

- [x] `just fix` / `just format` / `just typecheck`
- [x] New tests in `tests/sync/test_watch_service.py` cover: unconstrained, constrained, and missing-constraint cases via the new `_select_projects_to_watch()` helper
- [x] New tests in `tests/services/test_initialization.py` assert `BASIC_MEMORY_MCP_PROJECT` reaches the `WatchService` constructor (and is `None` when unset)
- [x] `pytest tests/sync/ tests/services/test_initialization.py tests/services/test_initialization_cloud_mode_branches.py` — 159 passed, 3 skipped

## Follow-ups (not in this PR)

Tracked in #758 — these are defensive hardening, not blockers:

- Guard `from_entity` in `Relation.permalink` the same way `to_entity` is guarded
- Use a unique tmp suffix (pid + uuid or `tempfile.mkstemp(dir=...)`) in `write_file_atomic` so intra-process concurrent writers on the same path do not collide

### DIFF
--- a/src/basic_memory/services/initialization.py
+++ b/src/basic_memory/services/initialization.py
@@ -99,18 +99,23 @@ async def initialize_file_sync(
     )
     project_repository = ProjectRepository(session_maker)
 
+    # Filter to constrained project if MCP server was started with --project.
+    # Applied to both the initial background sync and the watch service so that
+    # running multiple `basic-memory mcp --project X` processes does not produce
+    # duplicate watchers fighting over the same files.
+    constrained_project = os.environ.get("BASIC_MEMORY_MCP_PROJECT")
+
     # Initialize watch service
     watch_service = WatchService(
         app_config=app_config,
         project_repository=project_repository,
         quiet=quiet,
+        constrained_project=constrained_project,
     )
 
     # Get active projects
     active_projects = await project_repository.get_active_projects()
 
-    # Filter to constrained project if MCP server was started with --project
-    constrained_project = os.environ.get("BASIC_MEMORY_MCP_PROJECT")
     if constrained_project:
         active_projects = [p for p in active_projects if p.name == constrained_project]
         logger.info(f"Background sync constrained to project: {constrained_project}")
@@ -154,7 +159,10 @@ async def initialize_file_sync(
     # Don't await the tasks - let them run in background while we continue
 
     # Then start the watch service in the background
-    logger.info("Starting watch service for all projects")
+    if constrained_project:
+        logger.info(f"Starting watch service constrained to project: {constrained_project}")
+    else:
+        logger.info("Starting watch service for all projects")
 
     # run the watch service
     await watch_service.run()
--- a/src/basic_memory/sync/watch_service.py
+++ b/src/basic_memory/sync/watch_service.py
@@ -216,14 +216,14 @@ class WatchService:
                 # Why: watchfiles.awatch() requires at least one path. Calling it
                 #      with an empty list raises ValueError, which the outer handler
                 #      catches with a 5s sleep — producing a tight error-log loop.
-                # Outcome: sleep longer and retry, giving the project list a chance
-                #          to populate (e.g. after a project is added).
+                # Outcome: sleep the configured reload interval before retrying, so
+                #          newly added projects get picked up on the next cycle.
                 if not projects:
                     logger.warning(
                         "No projects to watch; sleeping before retry "
                         f"(constrained_project={self.constrained_project!r})"
                     )
-                    await asyncio.sleep(30)
+                    await asyncio.sleep(self.app_config.watch_project_reload_interval)
                     continue
 
                 project_paths = [project.path for project in projects]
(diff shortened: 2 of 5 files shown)


phernandez:
Addressed the empty-project-list feedback in b38823bb — added a guard in `run()` that logs a warning and sleeps 30 s when `_select_projects_to_watch()` returns an empty list. Prevents the tight retry loop when `constrained_project` names a missing project.

Leaving the duplicated cloud-filtering logic as-is since you flagged it as non-blocking and pulling it into a shared helper would be scope creep for this fix.

PR #775, opened 2026-04-29
fix(cli): cleanup local DB state on set-cloud/set-local

### LINKED ISSUE #680: bm project set-cloud leaves stale local project in database and config
## Summary

`bm project set-cloud <name>` does not fully remove the local project state, leaving the project in a `local+cloud` hybrid state instead of pure `cloud`.

## Steps to reproduce

1. Have a project `main` configured as local (with a local path like `/Users/me/basic-memory`)
2. Run `bm project set-cloud main`
3. Run `bm tool list-projects`

## Expected

`main` project should have:
- `source: "cloud"`
- `local_path: null`
- No entry in the `project` table of `memory.db` pointing to the old local path

## Actual

`main` project shows:
- `source: "local+cloud"`
- `local_path: "/Users/me/basic-memory"` (the old local path)
- The `project` table in `memory.db` still contains a row with the old local path

The config file (`~/.basic-memory/config.json`) is updated to `"mode": "cloud"` but still retains the old `"path"` value. The database `project` table is not cleaned up at all.

## Workaround

1. Remove the project entry from `config.json` (or set path to the cloud path)
2. Delete the row from the `project` table in `memory.db`:
   ```sql
   DELETE FROM project WHERE name = 'main';
   ```
3. Remove the (now empty) local directory if it still exists

After this, `bm tool list-projects` correctly shows `source: "cloud"`.

## Environment

- basic-memory: 0.20.2
- macOS arm64
- Python via uv

phernandez:
## Summary

Per @groksrc's direction in #680, `set-cloud` is now a one-way cutover instead of a routing toggle.

Before this change, `set-cloud` only flipped `mode = cloud` in config and left the project's row in the local index DB alone — so `_merge_projects` continued to see the project on both sides and reported `source: "local+cloud"` with the old `local_path` still populated. Confusing state, no clean way out without manually editing both `config.json` and `memory.db`.

## Changes

- **`set-cloud`** now drops the project's row from the local index DB and clears `path` in config. On-disk note files are preserved — the user can keep, archive, or delete them as they like.
- **`set-local`** is the symmetric inverse. It accepts `--local-path` and recreates the local DB row at the resolved path. If the prior path is still in config (e.g. running `set-local` on a project that's already local), `--local-path` may be omitted and the existing path is reused.
- Both helpers (`_detach_local_project_row`, `_attach_local_project_row`) are idempotent: running `set-cloud` twice doesn't error; running `set-local` twice updates the path if it changed and otherwise no-ops.

## Notes / behavior changes

- **Breaking-ish**: previously `set-local <name>` worked on a `set-cloud`'d project even though that left the DB in a hybrid state. After this change, that command path errors out and asks for `--local-path`. This is intentional — the prior state was the bug.
- The CLI `--local`/`--cloud` routing flags are unrelated to this; they remain a per-invocation override of *routing*, while `set-cloud`/`set-local` change the project's *configured* state.
- On-disk note files are explicitly NOT touched by either command.

## Test plan
- [x] `test_set_cloud_clears_path` — verifies config `path` is blanked after `set-cloud`
- [x] `test_set_local_requires_path_after_set_cloud` — regression for the exact scenario in the issue: re-binding to local must not silently default
- [x] `test_set_local_success` — full round trip with an explicit path
- [x] Existing `test_set_local_clears_workspace_id` updated to pass `--local-path`
- [x] `uv run pytest tests/cli/ --no-cov` — 309 passed

Closes #680

### DIFF
--- a/src/basic_memory/cli/commands/project.py
+++ b/src/basic_memory/cli/commands/project.py
@@ -26,7 +26,7 @@ from basic_memory.cli.commands.cloud.rclone_commands import (
 )
 from basic_memory.cli.commands.command_utils import get_project_info, run_with_cleanup
 from basic_memory.cli.commands.routing import force_routing, validate_routing_flags
-from basic_memory.config import ConfigManager, ProjectEntry, ProjectMode
+from basic_memory.config import BasicMemoryConfig, ConfigManager, ProjectEntry, ProjectMode
 from basic_memory.mcp.async_client import get_client, resolve_configured_workspace
 from basic_memory.mcp.clients import ProjectClient
 from basic_memory.schemas.cloud import (
@@ -838,7 +838,7 @@ def move_project(
         raise typer.Exit(1)
 
 
-async def _detach_local_project_row(app_config, name: str) -> bool:
+async def _detach_local_project_row(app_config: BasicMemoryConfig, name: str) -> bool:
     """Drop the project's row from the local index DB.
 
     Trigger: `bm project set-cloud` is making a project cloud-only.
@@ -863,10 +863,12 @@ async def _detach_local_project_row(app_config, name: str) -> bool:
         await repo.delete(existing.id)
         return True
     finally:
+        # CLI-only: safe to tear down the global DB singleton here since
+        # set-cloud/set-local never run inside a long-lived MCP/API server.
         await db.shutdown_db()
 
 
-async def _attach_local_project_row(app_config, name: str, path: str) -> None:
+async def _attach_local_project_row(app_config: BasicMemoryConfig, name: str, path: str) -> None:
     """Ensure the project has a row in the local index DB at the given path.
 
     Trigger: `bm project set-local` is making a previously cloud-only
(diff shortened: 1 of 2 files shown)


phernandez:
Quick triage of the Codex review on `8f26f159`:

**P1 — preserve `entry.path` for cloud projects with bisync mirrors:** Real concern, but it conflicts with the explicit `set-cloud` design from #680: a one-way cutover where the project's configured state ends up purely cloud. The bisync-mirror case still works through `entry.local_sync_path` (preserved here), and `bm cloud bisync` reads from that field; only the continuous `watch_service` filter (`Path(entry.path).is_absolute()`) is affected. The supported recovery for a user who really wanted to keep the watcher is `bm project set-local <name> --local-path <mirror>` followed by `bm project add --cloud --local-path <mirror>` — same path the `add --cloud` flow uses to seed both fields. Leaving as-is to honor the design choice.

**P2 — relative-path fallback in `set-local`:** Not a regression introduced here. Before this PR `entry.path` was already passed straight through `os.path.abspath()`, so legacy slug paths could rebase to CWD then too. The new behavior actually makes this harder to hit because `set-cloud` blanks `entry.path`, so a subsequent `set-local` with no `--local-path` falls into the `if not candidate` guard. Worth tightening separately if it ever shows up in the wild.

**P2 — `save_config` ordering on `set-cloud` and `set-local`:** Both ordering choices have a partial-failure mode (DB-row-only or config-only). A real atomic transaction would need the project DB and the JSON config under one writer, which is well outside the scope of this fix. The current order — DB first, then config — keeps the user-visible state (`bm project list`) in sync with the actual routing target if the config save raises mid-write, which is the failure mode I'd rather have. Punting.

CI is green; merging.

Issue #666, opened 2026-03-14
Retrieval pipeline improvements: reranking, length normalization, noise filtering, adaptive recall


bm-clawd:
## Context

Analyzed [memory-lancedb-pro](https://github.com/CortexReach/memory-lancedb-pro) (LanceDB-based OpenClaw memory plugin) and [their demo](https://github.com/lancedb/openclaw-lancedb-demo). While architecturally different from BM (opaque vectors vs human-readable plain text), their retrieval pipeline has several techniques worth adopting.

Our LoCoMo benchmark baseline: Recall@5 76.4%, Recall@10 85.5%, MRR 0.658. Weakest areas: single_hop 57.7%, temporal 59.1%. Root cause identified in #577: RRF scoring flattens results, FTS outperforms vector for observations.

### 1. Cross-encoder reranking (highest impact)
After initial hybrid retrieval (vector + FTS), run a second pass through a cross-encoder reranker (e.g., Jina reranker-v3, Voyage rerank-2.5). This would re-score candidates based on query-document semantic relevance rather than just embedding similarity.

memory-lancedb-pro uses: 60% cross-encoder score + 40% original fused score, with graceful fallback to cosine similarity on API failure.

**Implementation options:**
- Cloud API: Jina, Voyage, Pinecone (cheap per-query cost)
- Local: BAAI/bge-reranker via Ollama-compatible endpoint
- Config-driven: optional, off by default, provider-agnostic

This directly addresses #577 (RRF flattening) and should significantly improve Recall@5 and MRR.

### 2. Length normalization
Long notes/observations currently have an outsized influence on search results. A length normalization step (anchor around 500 chars) would prevent verbose entries from dominating over precise, short observations.

### 3. Noise filtering / adaptive retrieval
Skip memory retrieval entirely for queries that don't need it: greetings, slash commands, simple confirmations, emoji-only messages. Also filter low-quality content from capture: agent refusals, meta-questions, boilerplate.

This reduces wasted retrieval cycles and keeps the index cleaner.

### 4. MMR diversity filtering
When multiple results are very similar (cosine > 0.85), demote duplicates to increase result diversity. Prevents near-duplicate observations from consuming all top-K slots.

## What we already do better
- Human-readable plain text (their memories are opaque vector rows)
- Knowledge graph with relational structure (observations + relations)
- Bidirectional editing (humans can correct memories by editing files)
- Schema system for structured note types
- Git history as provenance trail

## Priority
[...]

PR #777, opened 2026-04-30
fix(mcp): resolve projects by external_id, remove workspace from MCP tools

phernandez:
## Summary

Lets MCP tools address projects unambiguously across cloud workspaces, without requiring callers to know about workspaces. Removes the `workspace` parameter from MCP tools entirely and adds an optional `project_id` (UUID) parameter to every project-scoped tool.

Closes basicmachines-co/basic-memory-cloud#673

## What changed

**1. Project resolution by `external_id` (UUID)**
- `resolve_workspace_project_identifier()` accepts a UUID and resolves it via a fast-path lookup against the workspace/project index — bypassing name/permalink resolution and workspace ambiguity handling.
- UUID inputs are canonicalized via `str(UUID(value))` before lookup, so uppercase hex, brace-wrapped (`{...}`), `urn:uuid:...`, and unhyphenated forms all map to the same key. (Codex P2)

**2. `workspace` parameter removed from MCP tools**
Workspaces are now purely an implementation detail. When a project name exists in multiple workspaces and no UUID disambiguator is given, resolution falls back to the default workspace (instead of erroring with "exists in multiple workspaces").

**3. Optional `project_id` parameter on all 14 project-scoped MCP tools**
`read_note`, `write_note`, `search_notes`, `edit_note`, `delete_note`, `view_note`, `read_content`, `move_note`, `list_directory`, `canvas`, `recent_activity`, `build_context`, `schema_validate/infer/diff`, `search_notes_ui`, `read_note_ui`.

- When provided, `project_id` takes precedence over `project` and routes to a specific project unambiguously.
- LLMs discover the UUID from `list_memory_projects()`, which now includes `external_id` in its output (and the docstring points there).
- Backward compatible: callers passing only `project` (the existing pattern) keep working unchanged.

**4. `get_project_client()` routes UUIDs correctly in both modes**
The local config keys projects by name only, so `get_project_mode(uuid)` would have falsely defaulted to CLOUD and broken pure-local routing for any UUID identifier. Fixed by:
- In Step 3 (cloud routing): only enter the cloud path for unknown identifiers when cloud is actually available (factory mode, explicit cloud, or credentials present).
- In Step 4 (local routing): when `project_id` is set, skip per-project mode lookup in `get_client()` — local mode routes everything through the same ASGI client; the API resolves the UUID via `/v2/projects/resolve`.

**5. Internal tool→tool calls preserve workspace selection**
`read_note._search_candidates` (the title/text fallback path) now passes `project_id=active_project.external_id` to `search_notes`. Without this, the inner re-resolution could select a different tenant via the default-workspace fallback when context was unavailable (CLI / `context=None`). (Codex P1, solved more thoroughly than the original suggestion — preserving routing without breaking local mode.)

Tools that delegate (`view_note` → `read_note`, `ui_sdk` → `search_notes`/`read_note`) propagate `project_id` through.

Tools calling `resolve_project_and_path()` (`read_note`, `read_content`, `search_notes`) now pass `active_project.name` (the resolved canonical name) instead of the raw `project` argument, so the inner `get_active_project()` cache hits even when `project_id` was used or `project` was wrong.

## Why this approach

The earlier draft had `read_note._search_candidates` swap `project=name` for `project=external_id`. That broke pure-local mode because the local config indexes by name only — passing a UUID hit the CLOUD-by-default branch in `get_project_mode()` and errored without cloud credentials.

Adding a separate `project_id` parameter:
- Keeps the existing name-based contract for local single-workspace users (no behavior change).
- Gives LLMs in cloud multi-workspace setups a way to specify the exact project after discovering its UUID via `list_memory_projects()`.
- Closes the workspace-ambiguity loop end-to-end: discovery → identifier → routing → workspace-correct API call.

## Test plan

- [x] All MCP unit tests pass (569 passed, 8 skipped)
- [x] New integration tests for `project_id` end-to-end through MCP transport pass
- [x] Lint clean, type-check clean
- [x] Tool signature contract test updated for new parameter
- [x] Telemetry tests updated for the new `requested_project_id` span attribute

### New tests
- `tests/mcp/test_project_context.py`:
  - `test_resolve_workspace_project_identifier_resolves_by_external_id` — direct UUID lookup picks the right workspace
  - `test_resolve_workspace_project_identifier_normalizes_uuid_forms` — uppercase/brace/urn/unhyphenated forms canonicalize
  - `test_get_project_client_prefers_project_id_over_project_name` — UUID precedence
  - `test_get_project_client_with_project_id_routes_locally_without_cloud` — regression for the local-mode UUID routing fix
- `test-int/mcp/test_read_note_integration.py`:
  - `test_read_note_by_project_id` — end-to-end read using only `project_id`
  - `test_read_note_project_id_takes_precedence_over_name` — wrong name + correct UUID still resolves correctly


bm-clawd (commented):
Reviewed against the goals stated in the PR description. **LGTM, ship once integration CI is green.** This is the right shape for the workspace-ambiguity problem.

## What I checked

**1. `resolve_workspace_project_identifier` UUID fast-path** — `str(UUID(project))` canonicalizes uppercase, brace-wrapped, `urn:uuid:`, and unhyphenated forms to the same key as the stored `external_id`. The `entries_by_external_id` dict is built once when the index is constructed, so UUID lookups are O(1) and bypass name/permalink resolution entirely. ✅

**2. `project_id` parameter on all 14 project-scoped tools** — verified every tool listed in the PR description actually has `project_id: Optional[str] = None` and threads it into `get_project_client(...)`. Delegating tools (`view_note → read_note`, `ui_sdk → search_notes/read_note`) propagate it through. ✅

**3. The `get_project_client` UUID-in-local-mode fix is the load-bearing change.** `config.get_project_mode(unknown_identifier)` defaults to CLOUD because the local config keys by project name. A UUID will *always* be unknown locally, so without this PR's guard a `project_id` argument would have falsely routed every local user through the cloud client and errored without credentials. The fix is in two places:
   - Step 3 only enters the cloud routing path when cloud is actually available (`factory_mode or explicit_cloud_routing or has_cloud_credentials(config)`).
   - Step 4 builds `client_kwargs = {} if project_id else {"project_name": resolved_project}` so the local ASGI client doesn't try to consult per-project mode for a UUID it can't recognize.
   The regression test `test_get_project_client_with_project_id_routes_locally_without_cloud` pins this exact failure mode. ✅

**4. Internal tool→tool routing preservation** — `read_note._search_candidates` passes `project_id=active_project.external_id` into `search_notes`. Without that, the inner `get_project_client(...)` call would re-resolve from a bare name and could land in a different tenant via the default-workspace fallback when context is unavailable (CLI, `context=None`). This was Codex P1 and the resolution is more thorough than the original suggestion — name-based routing still works for local single-workspace users because the inner call also receives `project=active_project.name`. ✅

**5. Removal of the `workspace` MCP parameter** — `list_memory_projects` no longer takes `workspace`, but the merged output now includes `external_id` for every project. The docstring tells the LLM where to find UUIDs and how to use them. The flow is closed end-to-end: discover via `list_memory_projects()` → pick UUID → pass as `project_id` → routed to the exact project regardless of name collisions. ✅

**6. Backwards compat** — pure local single-workspace users keep working with no changes (still pass `project="name"`, no UUID needed, name-keyed local config still resolves). ✅

## Minor things I'd flag (non-blocking)

1. **`_search_candidates` passes both `project=active_project.name` and `project_id=active_project.external_id`.** Since `project_id` takes precedence per the contract, the name is technically redundant. The comment in the code explains that this preserves workspace selection across re-resolution, but it would be slightly cleaner to either drop the name (let UUID alone disambiguate) or add a one-line comment saying "name is defensive — kept for log clarity and any code paths that read `project` directly." Cosmetic.

2. **`get_project_client` Step 1 calls `resolve_project_parameter(project_identifier=<uuid>, context=context)`.** That passes the UUID through `ProjectResolver` and `_canonicalize_project_name`, which loops over `config.projects` doing permalink matches that will never hit. Cheap, but on the order of tens of microseconds × 14 tools × hot paths. Consider an early `try: UUID(project_identifier)` branch that skips the resolver entirely when the input is already a canonical external_id — same shape as the fast-path you added in `resolve_workspace_project_identifier`. Not blocking.

3. **`detect_project_from_url_prefix` is unaware of `project_id`.** If a caller passes `project=None, project_id=<uuid>, identifier="memory://specs/foo"`, the URL-prefix detector still runs and may set `project="specs"` from the prefix. `project_id` then takes precedence in `get_project_client` so routing is correct, but the resolved active project might end up disagreeing with what the user expected the URL-prefix to mean. Worth a comment in `read_note` noting that `project_id` overrides URL prefix detection by design, or short-circuit the prefix detection when `project_id` is set. Edge case but worth a follow-up.

4. **Validation symmetry**: `resolve_workspace_project_identifier` raises a clear "Project '...' was not found in indexed cloud workspaces" error with the available list. But if a UUID is passed and *doesn't* match any `external_id`, the fast-path falls through to `_split_qualified_project_identifier` and then a permalink lookup that will also fail. The eventual error is "Project '<uuid>' was not found in indexed cloud workspaces" with a list of *names* — not great UX for a UUID typo. A targeted "no project with that external_id" error would help, but it's an edge case.

5. **Telemetry**: `requested_project_id` is added to spans. Worth confirming this is accepted by the existing telemetry filters / not flagged as PII (it's not — `external_id` is an opaque UUID, not the cloud user's tenant_id). Sanity check, not a change request.

## Risk assessment

- **Local single-workspace users**: zero behavior change. Name-based contract preserved.
- **Cloud single-workspace users**: zero behavior change unless they opt into UUIDs. The default-workspace fallback for ambiguous names replaces the old "exists in multiple workspaces" error, which is *less* surprising for the common case.
- **Cloud multi-workspace users with name collisions**: previously errored, now resolvable via `project_id`. Strictly better.
- **The internal CLI/context=None re-resolution path**: previously could pick the wrong tenant on default-workspace fallback. Fixed.
- **Failure mode**: passing an unknown UUID errors clearly. Passing a local-mode UUID without cloud routes locally and the API resolves it via `/v2/projects/resolve`. No silent wrong-tenant routes that I can construct.

## Tying back to PR #666

Worth noting: PR #666 fixed the cloud-side data loss when `move_note` raced its own webhooks. PR #777 closes the upstream ambiguity that *causes* MCP callers to land in the wrong tenant in the first place. Together they make multi-workspace MCP routing correct end-to-end — caller picks an unambiguous identifier (#777), tenant performs the move without losing data on stale webhooks (#666). Good shipping order.

Approving once the test matrix lands green. Nice work on isolating the UUID-in-local-mode trap before it shipped — that one would have been a confusing customer report.

joe-ref (approved):

PR #812, opened 2026-05-10
fix(core): use updated_at for recent_activity filter and ordering

### LINKED ISSUE #811: Recent activity filters on created_at instead of updated_at
## Bug Description

`recent-activity` can return misleadingly empty or stale results because the date filter is applied to `search_index.created_at` instead of `search_index.updated_at`.

This means notes/entities that were created before the requested timeframe but edited/updated inside the timeframe are excluded from recent activity, even though they are the most important results for a “what changed recently?” query.

## Steps to Reproduce

1. Create an entity/note older than the requested recent-activity timeframe.
2. Update that existing entity/note within the timeframe.
3. Query recent activity, for example:

```bash
bm tool recent-activity --project main --timeframe 14d
```

or through the v2 memory/recent activity endpoint that uses the search repository date filter.

## Expected Behavior

Recent activity should include entities/notes whose `updated_at` falls within the requested timeframe, even when their `created_at` is older.

Unranked recent-activity results should be ordered by recency, e.g. newest `updated_at` first.

## Actual Behavior

The search repository filters `after_date` against `search_index.created_at`, so updated older notes are omitted.

In unranked recent-activity calls, ordering is also not recency-first; it sorts by a constant/null rank and id before updated recency is considered.

## Suspected Root Cause

Both SQLite and Postgres search repositories use `created_at` for `after_date` filtering where recent-activity semantics require `updated_at`:

- `src/basic_memory/repository/sqlite_search_repository.py`
- `src/basic_memory/repository/postgres_search_repository.py`

The unranked ordering should prefer:

```sql
updated_at DESC, id ASC
```

rather than rank/id ahead of recency.

## Suggested Fix

- Change `after_date` filters from `created_at >= after_date` to `updated_at >= after_date` in both SQLite and Postgres search repositories.
- For unranked/recent-activity search, order by `updated_at DESC, id ASC`.
- Add regression coverage that verifies:
  - an old-created but recently-updated entity is included,
  - a stale entity outside the timeframe is excluded,
  - results are ordered newest `updated_at` first.

## Validation Notes

A local regression test was drafted against `tests/api/v2/test_memory_router.py` and passed with the repository changes:

```bash
uv run pytest tests/api/v2/test_memory_router.py -q
# 13 passed

uv run pytest \
  tests/services/test_search_service.py::test_after_date \
  tests/api/v2/test_search_router.py::test_search_with_date_filter \
  -q
# 2 passed
```

## Environment

- Repo: `basicmachines-co/basic-memory`
- Observed from local checkout on `main` around commit `dd91b490`
- Affected behavior: Basic Memory recent activity / search date filtering

groksrc:
## Summary

- `recent_activity` was silently dropping notes whose `updated_at` fell inside the requested timeframe but whose `created_at` did not, because both `SQLiteSearchRepository._build_fts_query_parts` and `PostgresSearchRepository` filtered the `after_date` window against `created_at`. Both now filter on `updated_at`.
- Postgres' unranked ORDER BY put `id ASC` before the optional `updated_at DESC` tiebreaker, so timeframe-only queries (no FTS text) effectively came back in insertion order. `updated_at DESC` now precedes `id ASC`. (SQLite was already correct because score is constant for unranked queries and `updated_at DESC` was the next clause.)
- Added a regression test that indexes one entity with an old `created_at` but recent `updated_at` and one fully-stale entity, then asserts the recent one appears, the stale one doesn't, and results come back ordered newest-updated first.

Fixes #811

## Test plan

- [x] `uv run pytest tests/services/test_search_service.py -q --no-cov` — 56 passed
- [x] `uv run pytest tests/api/v2/test_memory_router.py tests/repository -q --no-cov` — 294 passed, 17 skipped
- [x] `uv run pytest tests -q --no-cov -k "search or recent_activity or memory_router"` — 361 passed, 21 skipped
- [ ] CI (SQLite + Postgres via testcontainers)

### DIFF
--- a/src/basic_memory/repository/postgres_search_repository.py
+++ b/src/basic_memory/repository/postgres_search_repository.py
@@ -774,7 +774,8 @@ class PostgresSearchRepository(SearchRepositoryBase):
         # Handle date filter
         if after_date:
             params["after_date"] = after_date
-            conditions.append("search_index.created_at > :after_date")
+            # Filter on updated_at so recently-edited notes are included even when created_at is old
+            conditions.append("search_index.updated_at > :after_date")
             # order by most recent first
             order_by_clause = ", search_index.updated_at DESC"
 
@@ -945,7 +946,7 @@ class PostgresSearchRepository(SearchRepositoryBase):
                 {score_expr} as score
             FROM {from_clause}
             WHERE {where_clause}
-            ORDER BY score DESC, search_index.id ASC {order_by_clause}
+            ORDER BY score DESC {order_by_clause}, search_index.id ASC
             LIMIT :limit
             OFFSET :offset
         """
--- a/src/basic_memory/repository/sqlite_search_repository.py
+++ b/src/basic_memory/repository/sqlite_search_repository.py
@@ -789,7 +789,8 @@ class SQLiteSearchRepository(SearchRepositoryBase):
         # Handle date filter using datetime() for proper comparison
         if after_date:
             params["after_date"] = after_date
-            conditions.append("datetime(search_index.created_at) > datetime(:after_date)")
+            # Filter on updated_at so recently-edited notes are included even when created_at is old
+            conditions.append("datetime(search_index.updated_at) > datetime(:after_date)")
 
             # order by most recent first
             order_by_clause = ", search_index.updated_at DESC"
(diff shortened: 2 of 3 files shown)


phernandez:
Addressed the review cleanup in 7e938ef3: moved `timezone` to the module import and wrapped the long updated_at normalization expression. Local checks: `uv run ruff check tests/services/test_search_service.py`; `uv run pytest tests/services/test_search_service.py::test_after_date_uses_updated_at -q --no-cov`; `uv run pytest tests/services/test_search_service.py -q --no-cov`.

Issue #721, opened 2026-04-05
edit_note fails on notes with long text around inline wikilinks (relation_type exceeds MaxLen)


manonstreet:
## Summary

`edit_note` silently fails on notes where inline wikilinks have long surrounding text (~180+ chars). The note becomes "poisoned" — reads work fine, but all subsequent `edit_note` calls fail with a validation error during re-indexing.

## Reproduction

1. Create a note with a long descriptive bullet containing an inline wikilink:

```markdown
- **Lorem ipsum dolor sit amet** — consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Trust boundary model documented in [[Some Note Title]] for additional context on the architectural design patterns.
```

The text **preceding** `[[Some Note Title]]` must exceed 200 characters. The `relation_type` is extracted from the text before the wikilink — text after the wikilink doesn't matter. ~193 chars before the link passes; ~260 chars fails.

2. Let the file watcher index the note (or create via `write_note` — both work)
3. Call `edit_note` on this note (any operation — append, find_replace, etc.)
4. Edit fails with:

```
1 validation error for EntityResponseV2
relations.0.relation_type
  String should have at most 200 characters [type=string_too_long, input_value='**Lorem ipsum dolor sit ...boundary model found in', input_type=str]
```

5. All subsequent `edit_note` calls on this note fail with the same error

## Root Cause

In `src/basic_memory/models/knowledge.py`, the `Observation.permalink` property truncates content to 200 chars:

```python
content_for_permalink = self.content[:200] if len(self.content) > 200 else self.content
```

In `src/basic_memory/markdown/plugins.py`, `parse_relation()` (line 98) extracts the relation type by taking ALL text before `[[`:

```python
# Line 113-116
before = content[:start].strip()
if before:
    rel_type = before
```

This works for intended usage: `- reports_to [[Jane Smith]]` → `rel_type = "reports_to"`. But when a prose sentence contains an inline wikilink: `- Long descriptive text here... See [[Some Note]]` → `rel_type = "Long descriptive text here... See"`.

The Pydantic model `EntityResponseV2` in `schemas/base.py` enforces `MaxLen(200)` on `relation_type`. When preceding prose exceeds 200 chars, validation fails. Text after the wikilink is not the issue.

[...]

PR #823, opened 2026-05-14
fix(core): don't capture prose prefix before wikilink as relation type

picklesueat:
## Problem

`is_explicit_relation()` treats **any** list item containing `[[...]]` anywhere in it as an explicit `relation_type [[target]]` relation. `parse_relation()` then captures *everything before the `[[`* as the relation type.

So a list item that just mentions a wikilink inside ordinary prose:

```markdown
- After a long discussion we eventually decided to revisit [[Some Note]]
```

produces a relation whose `type` is the entire prose sentence. That's a meaningless edge in the knowledge graph — and while `RelationType` still carried `MaxLen(200)`, a prose prefix over 200 chars failed Pydantic validation and **rejected the whole note write**. One prose bullet poisoned the entire document.

This is the bug reported in #721. #770 fixed the *crash* by removing `MaxLen` from `RelationType`, but explicitly left the parsing heuristic out of scope — so on `main` today, prose with a wikilink no longer errors, but is still silently ingested as an explicit relation typed with a whole sentence.

## Fix

A relation type is a short label (`relates_to`, `depends on`) — never a sentence. This adds a `MAX_RELATION_TYPE_LENGTH = 100` bound:

- `is_explicit_relation()` returns `False` when the text before `[[` exceeds that bound. The token then falls through `relation_rule`'s existing `else` branch to `parse_inline_relations()` and is correctly recorded as a generic `links_to` relation — the wikilink is preserved, the prose is not captured as a type.
- `parse_relation()` guards the same bound directly as defense-in-depth, so a prose prefix can never become a relation type regardless of caller.

Two regression tests added in `tests/markdown/test_relation_edge_cases.py`: a long-prose bullet routes to `links_to` (and `parse_relation` returns `None`), and a genuine short `relates_to [[...]]` is unaffected.

## Scope / deliberately deferred

This is intentionally the **minimal** fix — a single length bound. The explicit/inline distinction is fundamentally ambiguous under the current syntax (`- saw this in [[X]]` is indistinguishable from a real short relation type), and `100` is a conservative threshold that cleanly catches prose sentences without touching any plausible real label. Two follow-ups were considered and left out to keep this reviewable:

1. **Tighter heuristics** — also reject prefixes with sentence punctuation or many words. Shrinks the ambiguous middle zone but adds individually-debatable rules.
2. **A syntax-level marker** for explicit relations (e.g. `relation:: [[X]]`) — the only thing that makes the distinction *truly* decidable, but backward-incompatible and a maintainer call.

Happy to fold either in if you'd prefer a different scope.

### DIFF
--- a/src/basic_memory/markdown/plugins.py
+++ b/src/basic_memory/markdown/plugins.py
@@ -85,14 +85,39 @@ def parse_observation(token: Token) -> Dict[str, Any]:
 
 
 # Relation handling functions
+
+# A relation type is a short label ("relates_to", "depends on"), never a
+# sentence of prose. A list item can legitimately contain a [[wikilink]] inside
+# ordinary prose ("- I read about this in [[Some Note]] last week") -- that is
+# an inline link, not an explicit `type [[target]]` relation. The bound below
+# is what separates the two: anything longer before the [[ is treated as prose.
+# Why a bound is needed: without it, the entire prose prefix is captured as the
+# relation type (see parse_relation), producing meaningless edges -- and, while
+# RelationType still carried a MaxLen, failing the whole note write.
+MAX_RELATION_TYPE_LENGTH = 100
+
+
 def is_explicit_relation(token: Token) -> bool:
-    """Check if token looks like our relation format."""
+    """Check if token looks like our relation format.
+
+    An explicit relation is `relation_type [[target]] (context)` where
+    relation_type is a short label. A list item whose [[wikilink]] is preceded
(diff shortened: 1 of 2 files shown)


phernandez:
Closing this in favor of #824, which implements the agreed relation grammar: unquoted explicit relation types stay single-token, quoted multi-word labels are supported with double or single quotes, and unquoted prose or bare wikilinks index as links_to. #824 credits @picklesueat for surfacing the parser ambiguity and starting the fix here.

PR #824, opened 2026-05-14
fix(core): parse prose wikilinks as inline links

phernandez:
## Summary
- Tighten explicit relation parsing so unquoted labels must be a single token, while quoted labels can contain spaces.
- Support both `"some type" [[Target]]` and `'some type' [[Target]]` as explicit multi-word relation types.
- Treat unquoted prose before a wikilink as a bare inline `links_to` relation instead of preserving the prose as `relation_type`.

## Credit
Credit to @picklesueat for opening #823 and surfacing the parser ambiguity. This PR carries the grammar variant we want to merge from that discussion.

## Tests
- `uv run ruff format src/basic_memory/markdown/plugins.py tests/markdown/test_relation_edge_cases.py tests/services/test_entity_service.py tests/api/v2/test_knowledge_router.py test-int/mcp/test_long_relation_type_integration.py`
- `uv run ruff check src/basic_memory/markdown/plugins.py tests/markdown/test_relation_edge_cases.py tests/services/test_entity_service.py tests/api/v2/test_knowledge_router.py test-int/mcp/test_long_relation_type_integration.py`
- `uv run pytest tests/markdown -q --no-cov`
- `uv run pytest tests/services/test_entity_service.py -q --no-cov`
- `uv run pytest tests/api/v2/test_knowledge_router.py tests/mcp/test_tool_write_note.py tests/mcp/test_tool_move_note.py -q --no-cov`
- `uv run pytest test-int/mcp/test_long_relation_type_integration.py -q --no-cov`
- `just typecheck`
- `just doctor`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Relation parsing no longer treats unquoted multi-word text before a wikilink as a
+  custom relation type. Use single-token relation types like `relates_to [[Target]]`,
+  or quote multi-word relation types like `"relates to" [[Target]]` or
+  `'relates to' [[Target]]`.
+  - Bare list wikilinks like `- [[Target]]` now index as `links_to`.
+  - Prose list items like `- some other thing [[Target]]` now index as `links_to`.
+  - To preserve existing multi-word relation types on re-sync, quote them before upgrading.
+
 ## v0.20.3 (2026-03-26)
 
 ### Bug Fixes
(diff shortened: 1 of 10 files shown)


phernandez:
Addressed the review items in 57bdd3ca. Updated docs/NOTE-FORMAT.md with the new explicit relation grammar, quoted multi-word examples, and links_to behavior for bare/prose wikilinks; added README and write_note examples for double and single quoted relation labels; added an Unreleased migration note; and changed the write_note test fixture to use the quoted relation form with a DB assertion for the stored relation type.

PR #835, opened 2026-05-17
fix: route local project ids before cloud lookup

### LINKED ISSUE #834: Local-only project_id routing fails when cloud workspaces are available
## Bug Description

`list_memory_projects`/`bm_projects()` returns an `external_id` for local-only projects and documents that callers can pass that value as `project_id` to other MCP tools. In a mixed local+cloud setup, using a local-only project's `external_id` as `project_id` fails because routing attempts to resolve the UUID only through the indexed cloud workspaces.

This makes `project_id` routing inconsistent: cloud and local+cloud project IDs work, but local-only project IDs returned by the same discovery tool do not.

## Reproduction

Environment observed on Drew's Hermes/Basic Memory setup:

- Basic Memory checkout: `/Users/drew/code/basic-memory`
- Current branch: `main`
- Version commit observed: `60ec6728 chore: update version to 0.21.1 for v0.21.1 release`
- Config has both local projects and cloud credentials/workspaces available.

`bm_projects()` returns a local-only project:

```text
name: hermes-memory
external_id: ee5f396d-6c67-45f8-90a1-da936f6e6f4f
source: local
qualified_name: null
```

Routing by project name succeeds:

```python
bm_recent(project="hermes-memory", timeframe="30d", limit=1)
```

Result:

```text
## Recent Activity: hermes-memory (30d)

Recent Notes & Documents ...
```

Routing by the returned local `external_id` fails:

```python
bm_recent(project_id="ee5f396d-6c67-45f8-90a1-da936f6e6f4f", timeframe="30d", limit=1)
bm_search(project_id="ee5f396d-6c67-45f8-90a1-da936f6e6f4f", query="Hermes", limit=1)
```

Error:

```text
Project 'ee5f396d-6c67-45f8-90a1-da936f6e6f4f' was not found in indexed cloud workspaces. Available projects: personal/..., basic-memory-.../...
```

Cloud/local+cloud IDs in the same environment do work:

```python
bm_recent(project_id="2c195cea-74ca-4c86-acf8-4458f09c6aef", timeframe="30d", limit=1)  # personal/main
bm_recent(project_id="fb88a546-299a-4f89-95ff-59793f7d1abc", timeframe="30d", limit=1)  # org/main
```

## Expected Behavior

Any `external_id` returned by `list_memory_projects(output_format="json")` / `bm_projects()` should be usable as `project_id` on per-project MCP tools, including local-only projects.

For a local-only project ID, Basic Memory should route through local ASGI and resolve the UUID against the local project database.

## Actual Behavior

When cloud credentials/workspace discovery are available, a local-only `project_id` is treated as a cloud/workspace UUID and fails in `resolve_workspace_project_identifier()` because local-only projects are not included in the cloud workspace index.

## Suspected Root Cause

The issue appears to be in:

```text
src/basic_memory/mcp/project_context.py
```

`get_project_client(project_id=...)` sets:

```python
project_identifier = project_id if project_id else project
resolved_project = await resolve_project_parameter(project_identifier, context=context)
```

Then, because `has_cloud_credentials(config)` is true, cloud routing remains available and the UUID is resolved through `resolve_workspace_project_identifier()`. That function only searches `WorkspaceProjectIndex.entries_by_external_id`, which is populated from cloud workspace project discovery, not local projects.

There is already a pure-local regression test nearby:

```text
tests/mcp/test_project_context.py::test_get_project_client_with_project_id_routes_locally_without_cloud
```

The missing case is the same local UUID routing behavior when cloud credentials are present.

## Suggested Fix

In `get_project_client(project_id=...)`, before attempting cloud workspace UUID resolution, try resolving the UUID against local projects when local projects/config are present. If it matches a local project, route via local ASGI. If it does not match locally, continue the existing cloud workspace resolution path.

Potential regression test:

- Configure a local project with an `external_id`.
- Configure cloud credentials / make `has_cloud_credentials(config)` true.
- Stub cloud workspace index so it does not include the local project ID.
- Call `get_project_client(project_id=<local external_id>)`.
- Assert `get_client()` is called for local ASGI routing and no `resolve_workspace_project_identifier()` failure occurs.

## Impact

Hermes and other MCP clients following the documented discovery flow can get a valid-looking `project_id` from `bm_projects()` and then fail on subsequent calls for local-only projects in mixed local+cloud environments.

wyf027:
## Summary
- Try resolving `project_id` UUIDs through local ASGI before falling back to cloud workspace discovery in mixed local+cloud configs.
- Preserve the existing cloud lookup path when the UUID is not a local project.
- Add a regression test for local-only `project_id` routing when cloud credentials are configured.

Fixes #834

## Test plan
- `uv run pytest tests/mcp/test_project_context.py::test_get_project_client_with_project_id_routes_locally_with_cloud_credentials -q` (failed before the fix)
- `uv run pytest tests/mcp/test_project_context.py -q`
- `uv run ruff check src/basic_memory/mcp/project_context.py tests/mcp/test_project_context.py`
- `uv run ruff format --check src/basic_memory/mcp/project_context.py tests/mcp/test_project_context.py`
- `uv run ty check src/basic_memory/mcp/project_context.py tests/mcp/test_project_context.py`
- `git diff --check`

### DIFF
--- a/src/basic_memory/mcp/project_context.py
+++ b/src/basic_memory/mcp/project_context.py
@@ -1459,6 +1459,45 @@ async def get_project_client(
     # Outcome: in pure local mode, treat UUID identifiers as local routing; cloud
     #   discovery still happens when factory/explicit/credentials are present
     cloud_available = factory_mode or explicit_cloud_routing or has_cloud_credentials(config)
+
+    # Trigger: project_id is a local external_id in a mixed local+cloud setup.
+    # Why: UUIDs are not local config keys, so get_project_mode() treats them as
+    #   cloud projects and searches only the cloud workspace index.
+    # Outcome: first give the local ASGI API a chance to resolve the UUID; if it
+    #   is not a local project, keep the existing cloud workspace lookup path.
+    if project_id and config.projects and not factory_mode and not explicit_cloud_routing:
+        try:
+            canonical_project_id = str(UUID(project_id))
+        except ValueError:
+            pass
+        else:
+            with logfire.span(
+                "routing.local_project_id_probe",
+                project_id=canonical_project_id,
+            ):
+                async with get_client() as client:
+                    try:
+                        active_project = await get_active_project(
+                            client,
+                            canonical_project_id,
+                            None,
+                        )
+                    except ToolError as exc:
+                        if "not found" not in str(exc).lower():
+                            raise
+                    else:
+                        route_mode = "local_asgi"
+                        await _clear_cached_active_workspace_for_local_route(context)
+                        await _set_cached_active_project(context, active_project)
+                        with logfire.span(
+                            "routing.client_session",
+                            project_name=active_project.name,
+                            route_mode=route_mode,
+                        ):
+                            logger.debug("Using local ASGI routing for project_id")
+                            yield client, active_project
+                        return
+
     if project_id and not cloud_available:
         project_mode = ProjectMode.LOCAL
 
--- a/tests/mcp/test_project_context.py
+++ b/tests/mcp/test_project_context.py
@@ -1233,6 +1233,57 @@ async def test_get_project_client_with_project_id_routes_locally_without_cloud(
 
 
 @pytest.mark.asyncio
+async def test_get_project_client_with_project_id_routes_locally_with_cloud_credentials(
+    config_manager, monkeypatch
+):
+    """A local UUID project_id must not be forced through cloud workspace lookup."""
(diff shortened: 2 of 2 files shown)


groksrc:
Thanks for your contribution! I'm picking up your change in #837 as it is a broader fix.

That PR includes the local project_id routing fix from here, plus the broader mixed local/cloud workspace-qualified routing work. #837 was also updated in aae12b21 to prefer local ASGI first for project_id routing while preserving cloud fallback and env constraints.

PR #844, opened 2026-05-22
feat(core): add XDG_CONFIG_HOME support

rafamadriz:
The XDG Spec defines where user directories for config, data and state should go. <https://specifications.freedesktop.org/basedir/latest/#variables>

This just adds support for `XDG_CONFIG_HOME`. If interested, here is a [list of other programs that support it](https://wiki.archlinux.org/title/XDG_Base_Directory#Supported).

If `XDG_CONFIG_HOME` doesn't exists, it still defaults to `~/.basic-memory` so it won't cause any disruption to existing users.

Let me know if you want any change or more tests.


groksrc:
This LGTM :shipit:

PR #859, opened 2026-05-28
chore(installer): consolidate agent packages

phernandez:
## Summary
- Consolidate Claude Code, shared skills, Hermes, and OpenClaw sources into the Basic Memory monorepo
- Add root/package-local just targets and validation scripts for agent package checks
- Update README, AGENTS.md, package docs, release workflow, changelog, and version dry-run tooling

## Verification
- uv run ruff check scripts/update_versions.py scripts/validate_claude_plugin.py scripts/validate_hermes_plugin.py scripts/validate_skills.py
- uv run ruff format --check .
- ruby -e 'require \"yaml\"; ARGV.each { |p| YAML.load_file(p); puts \"ok #{p}\" }' .github/workflows/consolidated-packages.yml .github/workflows/release.yml .github/workflows/pr-title.yml
- git diff --check
- just package-check
- just package-check-openclaw
- just release-dry-run v0.99.0
- just doctor

## Notes
- No PR review comments existed before creating this PR.
- OpenClaw generated artifacts are ignored from the monorepo root.


groksrc:
### Licensing: nested LICENSE files

Took a look at the vendored `LICENSE` files:

**`integrations/hermes/LICENSE` — recommend removing (redundant).**
It's byte-identical to the repo-root `LICENSE` (AGPL-3.0-or-later), so it's pure duplication. Nothing consumes it — `validate_hermes_plugin.py` only checks `plugin.yaml`/entrypoint/skill/tests, the hermes package is just `plugin.yaml` + `__init__.py` (no sdist/wheel that needs a co-located license), and no workflow/justfile copies it into an artifact. The root LICENSE already covers everything in-tree. (Only reason to keep it would be if hermes is meant to be re-extracted into a standalone repo someday.)

**`integrations/openclaw/LICENSE` — please confirm this is intentional.**
This one is **MIT** (`Copyright (c) 2026 Basic Machines`), which diverges from the repo's AGPL-3.0. If openclaw is deliberately meant to be more permissively licensed than the core (reasonable for client/integration code), this should stay as-is. If not, an MIT subdir inside an AGPL copyleft repo is worth resolving explicitly rather than leaving ambiguous. Flagging rather than touching it.

Neither file is referenced by CI, the justfiles, or the validation scripts, so removing `hermes/LICENSE` breaks no build/release/`package-check` step.

groksrc:
### `basic-memory-manager` agent: tool names assume a fixed MCP server name

`plugins/claude-code/agents/basic-memory-manager.md` hardcodes its `tools:` list with the `mcp__basic-memory__` prefix (`mcp__basic-memory__write_note`, etc.). Claude Code's subagent `tools:` field accepts **exact tool names only** — no wildcard/regex, unlike the hooks `matcher` which uses `mcp__.*__write_note` precisely because the server name varies (local / cloud / claude.ai connector).

Consequence: if a user's Basic Memory MCP server is registered under a name other than `basic-memory`, none of the agent's tool entries match and **the agent starts with no tools** — and it fails silently.

I looked at whether to fix this structurally (e.g. drop `tools:` so the agent inherits whatever's exposed under its real name), but I don't think that's worth it here:
- This agent only runs **inside Claude Code**, where the user controls their own MCP config and `basic-memory` is the conventional name (matches the package + CLI + what PLUGIN.md already assumes). A mismatch is rare and self-correctable in one line of user config.
- The claude.ai-connector / cloud variants that justify the hooks' regex are a different surface where this Claude Code subagent doesn't execute.
- The curated 11-tool list also documents intent (exactly which tools this agent is built around) — worth keeping.

**Recommendation: just document the assumption** rather than change the tool list. A short note — ideally a comment right above the `tools:` line in the agent `.md`, where a maintainer editing it would see it — to the effect of:

> These tool names assume the Basic Memory MCP server is registered as `basic-memory`. If yours uses a different name, update the `mcp__<name>__` prefixes to match — otherwise the agent starts with no tools (the field needs exact names; it doesn't support the wildcard the hooks matcher uses).

A YAML `#` comment in the frontmatter is safe for both `claude plugin validate --strict` and the repo's `validate_claude_plugin.py` parser (verified). Low-stakes, but flagging so the silent-failure mode is at least written down.

groksrc:
### Plugin README: slash-command example drops the plugin namespace

`plugins/claude-code/README.md:32` says:

> Skills auto-expose as slash commands in Claude Code (e.g., `/knowledge-organize`).

Right in spirit, but the example is wrong. **Plugin** skills are namespaced by the plugin name (`basic-memory`, from `plugin.json`), so the actual commands are `/basic-memory:knowledge-organize`, `/basic-memory:placement`, etc. — not bare `/knowledge-organize`.

Worth correcting because the plugin **already uses the namespaced form internally** — the PreToolUse hook in `hooks.json` injects "invoke the placement skill (`basic-memory:placement`)" — so the README is the only place that drops it.

Suggested fix: change the example to `/basic-memory:knowledge-organize` (and similarly anywhere else bare skill commands are shown).

Two smaller related notes while in this area:

- **Skills surface two ways**, and the README only emphasizes the slash-command path. None of the 6 skills set `disable-model-invocation`/`user-invocable: false`, so they're both (a) **model-invoked** — Claude auto-runs them from the `description` frontmatter when context matches (the primary design here, e.g. `placement` before a write), and (b) user-invokable via the namespaced slash command. Might be worth a one-liner so users know the slash command is the manual override, not the main path.

- **Name overlap between a skill and an MCP prompt.** The README notes the MCP server's prompts (`continue_conversation`, etc.) "surface as their own slash commands." Those are a separate surface: `/mcp__basic-memory__continue_conversation` (double-underscore, **user-invoked only**). So there are two "continue conversation" entry points — the skill `/basic-memory:continue-conversation` (hyphen, model-invocable) and the MCP prompt `/mcp__basic-memory__continue_conversation` (underscore, user-only). Not a bug, but a plausible point of user confusion worth a sentence of disambiguation in the docs.

groksrc:
### Shared `skills/`: lock file is stale and sync hygiene is fragile

The skills themselves look good — 10 well-scoped `memory-*` skills, clear frontmatter, `validate_skills.py` enforces `name == dir`, and `AGENTS.md → CLAUDE.md` is a clean symlink. Issues are around the lock file and keeping things in sync.

**1. `skills-lock.json` is stale.** It lists **9** skills but the directory has **10** — `memory-literary-analysis` is missing from the lock. Both the README table and the OpenClaw bundle (`openclaw.plugin.json`) list all 10, so the lock file is the lone straggler. Nothing validates it (`validate_skills.py` only checks `memory-*/SKILL.md` frontmatter).

**2. Does `skills-lock.json` belong in the source repo at all?** It's an `npx skills` *consumer* lockfile — it records `"source": "basicmachines-co/basic-memory"`, `"sourceType": "github"`, and content hashes. In the canonical *source* repo that's self-referential and looks vestigial from the former standalone `basic-memory-skills` repo. Either:
- remove it (a lockfile belongs in projects that *install* the skills, not in the source of truth), or
- if it's deliberately shipped for the satellite distribution, regenerate + validate it on every skill change.

Leaning toward removal; otherwise it keeps drifting.

**3. Adding a skill touches 3+ places but the checklist documents 1.** `DEVELOPMENT.md` "Adding a New Skill" lists: create dir → frontmatter → write → update README → commit. But a new skill also needs `skills-lock.json` (undocumented → already drifted) and the OpenClaw bundle (`bun run fetch-skills` + a path in `openclaw.plugin.json`, documented only in a separate section). This split is the root cause of the drift. Worth collapsing into one authoritative checklist, ideally backed by a validation step that cross-checks dirs ↔ README ↔ lock ↔ openclaw bundle.

**4. License: MIT again.** `skills/README.md` declares MIT inside the AGPL monorepo — third instance of the pattern (alongside `openclaw/LICENSE` and the claude-code plugin README). Folding this into the earlier licensing comment: it wants one consistent decision across all the satellite/integration packages rather than three independent MIT claims.

groksrc:
### Follow-up: consolidate the two skill collections onto `skills/` as the single source

Tracked in #866.

Not a blocker for this PR — #859's job is getting everything into one repo, which is the *precondition* for deduping skills across what used to be separate repos. Flagging here for visibility since the duplication becomes maintainable (and fixable) only after this lands.

The short version: there are two overlapping skill sets — `skills/memory-*` (framework-agnostic, canonical) and `plugins/claude-code/skills/` (6 hand-maintained, several duplicating the shared ones). This split is organic, not intentional. **OpenClaw already treats `skills/` as the single source** (`bun run fetch-skills`); the Claude Code plugin is the outlier that reinvents its own set. The fix is to make the plugin follow the same pattern — **bundle the shared skills + layer on only the host-specific glue (hooks)** — because most "plugin-specific" skills are really shared content + a hook trigger (e.g. `placement` = agent-agnostic folder logic + a CC-specific `PreToolUse` hook).

Full disposition checklist (which plugin skills retire vs. become shared vs. stay as glue) and the naming decision (uniform `memory-*`) are in #866.

PR #860, opened 2026-05-28
fix(mcp): resolve write_note overwrite conflicts by file_path strictly

DoubleDeeRuffy:
## Summary

`write_note(overwrite=True)` catches the 409 from `create_entity` and then resolves the conflicting entity to issue a PUT. The previous lookup used `knowledge_client.resolve_entity(entity.permalink)` with the default `strict=False`. In workspace-prefixed palaces the client-built permalink omits the workspace slug, so exact permalink lookup misses and the fuzzy fallback in `LinkResolver._resolve_in_project` can pick an orphan row that shares tokens with the canonical permalink. The update then writes to the orphan, leaving the canonical row stale, and the next overwrite mints a `-1`/`-2` suffix because `_resolve_schema_permalink` sees duplicate rows. Repeated re-synthesis runs accumulate orphans.

The 409 came from `prepare_create_entity_content`'s `file_service.exists(file_path)` check, so the file_path *is* the authoritative key for the canonical row. Resolve by file_path with `strict=True`, POSIX-normalized so Windows clients send the form the server stores.

## Reproduction

The user-facing reproducer (from the original bug report):

1. `write_note(title="overview", folder="projects/foo", content="A", overwrite=True)`
2. `write_note(title="overview", folder="projects/foo", content="B", overwrite=True)`
3. Inspect the DB — in workspace-prefixed palaces, this leaves two entity rows with `-1`/`-2` suffixed permalinks instead of one updated canonical row.

## What changed

- `src/basic_memory/mcp/tools/write_note.py`: on 409, look up by `entity.file_path` (POSIX-normalized) with `strict=True` instead of by `entity.permalink` with fuzzy fallback.
- `tests/mcp/test_tool_write_note.py`: regression test spies on `resolve_entity` to assert the identifier and `strict` flag, plus confirms no `-1`/`-2` suffixes get minted under the canonical permalink. Verified failing before the fix and passing after.

## Test plan

- [x] `uv run pytest tests/mcp/test_tool_write_note.py` — 51 passed (including new regression test)
- [x] `uv run pytest tests/mcp/test_tool_write_note_metadata.py tests/mcp/test_tool_write_note_kebab_filenames.py tests/api/v2/test_workspace_permalink_headers.py tests/mcp/test_workspace_permalink_resolution.py` — all pass
- [x] `uv run pytest tests/mcp/` — 660 pass, 8 skip, 2 pre-existing unrelated failures (`test_search_with_date_filter`, `test_recent_activity_format_relative_time_and_truncate_helpers`) that also fail on clean `origin/main`
- [x] `uv run ruff check` + `uv run ruff format --check` on changed files
- [x] `uv run ty check` on changed files
- [x] Verified the regression test fails on the previous code path

## Notes for review

- This is the first of two PRs for the bug report at [basic-memory-bug-report.md](https://github.com/basicmachines-co/basic-memory/issues) (filing a companion issue too). The second PR covers silent fuzzy resolution of ambiguous `[[wikilinks]]` in `sync_service` / `batch_indexer`.
- The bug report also mentions `file_path` columns containing the `main/` workspace prefix on orphan rows. I traced `_build_entity_fields` and found no code path that prepends the workspace slug to `file_path` — that field is always written via `file_path.as_posix()`. Those orphan paths likely come from legacy rows in the user's palace. Once the duplicate-row creation stops with this fix, `basic-memory reindex` will clean them up.

### DIFF
--- a/src/basic_memory/mcp/tools/write_note.py
+++ b/src/basic_memory/mcp/tools/write_note.py
@@ -1,6 +1,7 @@
 """Write note tool for Basic Memory MCP server."""
 
 import textwrap
+from pathlib import Path
 from typing import Annotated, List, Union, Optional, Literal
 
 import logfire
@@ -269,7 +270,19 @@ async def write_note(
                             raise ValueError(
                                 "Entity permalink is required for updates"
                             )  # pragma: no cover
-                        entity_id = await knowledge_client.resolve_entity(entity.permalink)
+                        # Resolve the conflicting entity by file_path with strict=True.
+                        # The 409 came from a file_service.exists(file_path) check, so this
+                        # file_path is the authoritative key for the canonical row. Resolving
+                        # by permalink with fuzzy fallback (the previous behavior) could pick
+                        # an orphan with a similar permalink — especially in workspace-prefixed
+                        # palaces where the client-built permalink omits the workspace slug —
+                        # causing the update to write to the wrong row and the next call to
+                        # mint a -1/-2 suffix on the canonical entity.
+                        # POSIX-normalize so Windows clients send the same form the server stores.
+                        file_path_identifier = Path(entity.file_path).as_posix()
+                        entity_id = await knowledge_client.resolve_entity(
+                            file_path_identifier, strict=True
+                        )
                         result = await knowledge_client.update_entity(
                             entity_id, entity.model_dump()
                         )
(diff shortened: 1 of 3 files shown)


phernandez:
@DoubleDeeRuffy thanks for the PR. I'll take a look at this.

phernandez (approved):
Reviewed the overwrite-conflict path and regression coverage. No blockers found.\n\nLocal verification on PR head 4fd9cae2935a:\n- `uv run pytest tests/mcp/test_tool_write_note.py` -> 41 passed\n- `uv run ruff check src/basic_memory/mcp/tools/write_note.py tests/mcp/test_tool_write_note.py` -> passed\n\nNote: the repo's full Tests workflow is push-only and did not run for this external PR branch, so this review is based on focused local verification rather than the full CI matrix.

PR #861, opened 2026-05-28
fix(sync): use strict link resolution for deferred forward references

DoubleDeeRuffy:
## Summary

Deferred relation resolution in `sync_service.resolve_forward_references` and `BatchIndexer._resolve_batch_relations` called `LinkResolver.resolve_link(to_name)` with the default `strict=False`. When the link text didn't match any entity's permalink, title, or file_path, the resolver fell through to BM25/ts_rank fuzzy search and silently picked `results[0]` — even for ambiguous links that share only incidental tokens with the wrong entity.

Producers commonly emit disambiguator-style wikilinks like `[[overview (state-management/session-execution)]]`. basic-memory does not parse `(qualifier)` as a disambiguator, so no entity matches exactly. The previous behavior resolved such links to whichever unrelated note shared the most tokens, polluting the graph with confidently-wrong edges that `basic-memory status` / `doctor` do not catch.

Entity-creation already passes `strict=True` (see `entity_service.update_entity_relations` line 1030). The deferred paths must use the same contract so unresolved relations stay unresolved (`to_id=NULL`) and surface to the producer instead of getting silently filled with a wrong target.

## Bug shape

From the bug report, four `part_of` relations in the affected palace pointed at unrelated notes via fuzzy match:

\`\`\`
(1694, 1711, 'overview (state-management/session-execution)',
       '.../phase-14-lint-state-v5.0-allowlist-correctness-verified-clean-1')
(1695, 1711, 'overview (state-management/session-execution)',
       '.../phase-14-lint-state-v5.0-allowlist-correctness-verified-clean-1')
(1696, 1711, 'overview (state-management/session-execution)',
       '.../phase-14-lint-state-v5.0-allowlist-correctness-verified-clean-1')
(1693, 1156, 'overview (state-management)',
       '.../roadmap-slim/read-open-phases-live-tree-status-divergence-pre-phase-2')
\`\`\`

## What changed

- `src/basic_memory/sync/sync_service.py`: pass `strict=True` to `resolve_link` in `resolve_forward_references`.
- `src/basic_memory/indexing/batch_indexer.py`: pass `strict=True` to `resolve_link` in `_resolve_batch_relations`.
- `tests/sync/test_sync_service.py`: regression test spies on `resolve_link` to assert `strict=True` is forwarded by deferred resolution and verifies an unresolvable forward reference stays unresolved.
- `tests/indexing/test_batch_indexer.py`: equivalent regression test for the batch-indexer path.

Verified both tests fail before the fix and pass after.

## Behavior change for users

Forward references that previously resolved via BM25/ts_rank fuzzy search will now stay unresolved (`to_id=NULL`) until an exact permalink/title/file_path match exists. This matches the contract used at entity creation time (`entity_service.update_entity_relations`).

Users relying on legitimate forward references — where the target's exact title/permalink exists by the time deferred resolution runs — see no change. The user-visible symptom of this fix is that ambiguous links no longer silently resolve to wrong targets.

## Test plan

- [x] \`uv run pytest tests/sync/\` — 195 passed (incl. new regression)
- [x] \`uv run pytest tests/indexing/\` — 17 passed (incl. new regression)
- [x] \`uv run pytest tests/services/test_link_resolver.py\` — passes; existing strict-mode behavior preserved
- [x] \`uv run pytest tests/sync/test_sync_service.py::test_forward_reference_resolution\` — confirms legitimate forward references still resolve via exact-match
- [x] \`uv run ruff check\` + \`uv run ruff format --check\` on changed files
- [x] \`uv run ty check\` on changed files

## Notes for review

- This is the second of two PRs for the bug report. The first is #860, which fixes `write_note(overwrite=True)` minting `-1`/`-2` duplicate rows in workspace-prefixed palaces via the same class of fuzzy-resolver mis-targeting.
- The bug report also requested documentation of which wikilink syntaxes the resolver supports (parens-suffix vs. \`Title|alias\` vs. explicit permalinks). That's a follow-up — happy to draft if there's interest.

### DIFF
--- a/src/basic_memory/indexing/batch_indexer.py
+++ b/src/basic_memory/indexing/batch_indexer.py
@@ -493,8 +493,14 @@ class BatchIndexer:
         async def resolve_relation(relation: Relation) -> int:
             async with semaphore:
                 try:
+                    # strict=True for deferred resolution: only fill in to_id on an
+                    # exact permalink/title/file_path match. Fuzzy fallback would silently
+                    # resolve ambiguous links to whichever entity shares tokens with the
+                    # link text, mismatching this with the sync_service forward-reference
+                    # path and producing confidently-wrong graph edges. See
+                    # sync_service.resolve_forward_references for the same change.
                     resolved_entity = await self.entity_service.link_resolver.resolve_link(
-                        relation.to_name
+                        relation.to_name, strict=True
                     )
                     if resolved_entity is None or resolved_entity.id == relation.from_id:
                         return 0
--- a/src/basic_memory/sync/sync_service.py
+++ b/src/basic_memory/sync/sync_service.py
@@ -1454,13 +1454,6 @@ class SyncService:
             # the most tokens, polluting the graph with confidently-wrong edges that no
             # audit catches. Leaving such relations unresolved keeps to_id=NULL so they
             # surface as forward references and can be fixed by the producer.
-            # Use strict=True: deferred resolution should only fill in to_id when an
-            # exact permalink/title/file_path match exists. The fuzzy fallback (search-based
-            # token match) would silently resolve ambiguous links like
-            # `[[overview (state-management/session-execution)]]` to whichever entity shares
-            # the most tokens, polluting the graph with confidently-wrong edges that no
-            # audit catches. Leaving such relations unresolved keeps to_id=NULL so they
-            # surface as forward references and can be fixed by the producer.
             resolved_entity = await self.entity_service.link_resolver.resolve_link(
                 relation.to_name, strict=True
             )
--- a/src/basic_memory/sync/sync_service.py
+++ b/src/basic_memory/sync/sync_service.py
@@ -1447,7 +1447,23 @@ class SyncService:
                 f"to_name={relation.to_name}"
             )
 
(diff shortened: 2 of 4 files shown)


phernandez (commented):
Behaviorally this is pointed in the right direction: strict deferred relation resolution matches the entity-creation contract, and the focused local checks passed.\n\nLocal verification on PR head 6ac28aeee9f7:\n- `uv run pytest tests/sync/test_sync_service.py::test_resolve_relations_uses_strict_link_resolution tests/indexing/test_batch_indexer.py::test_batch_indexer_uses_strict_link_resolution_for_deferred_relations tests/sync/test_sync_service.py::test_forward_reference_resolution tests/services/test_link_resolver.py::test_exact_match_types_in_strict_mode` -> 4 passed\n- `uv run ruff check src/basic_memory/indexing/batch_indexer.py src/basic_memory/sync/sync_service.py tests/indexing/test_batch_indexer.py tests/sync/test_sync_service.py` -> passed\n\nOne cleanup before I would merge: `src/basic_memory/sync/sync_service.py:1450` repeats the same strict-resolution explanatory comment twice. Please remove the duplicate block so the new guard reads once and stays crisp.\n\nAs with the other external PRs, the repo's full Tests workflow is push-only and did not run for this fork branch.

PR #865, opened 2026-05-29
feat(plugins): Claude Code plugin v0.4 — the memory bridge

phernandez:
## Summary

Modernizes the Claude Code plugin into the **bridge between Claude's working memory
and Basic Memory's durable graph** (DESIGN.md). The old v0.3.13 plugin (six skills, a
332-line agent, a write-note advisory hook, a config-note convention) is replaced with
a sharp v0.4 surface: two lifecycle hooks, an opt-in output style, four namespaced
skills, three picoschema seeds, and team-workspace support.

> **📚 Stacked on #859.** This PR is based on `codex/consolidate-agent-integrations`
> (#859, the SPEC-58 agent-integration consolidation), which it builds on. **Merge
> #859 first**; this PR's diff then shows only the incremental work. (GitHub will
> auto-retarget this to `main` once #859 lands.) The one consolidation-adjacent commit
> carried here is the `ui/tool-ui-react` deletion — happy to move it into #859 instead
> if preferred.

## What's in this PR (v0.4 Claude Code plugin)

Three layers:

- **Ambient (hooks):**
  - `SessionStart` → briefs Claude from the graph (structured `type`/`status` queries
    for active tasks, open decisions, and recent session checkpoints), reading the
    primary project **and** configured team/shared projects in parallel. Best-effort,
    <10k chars, first-run nudge.
  - `PreCompact` → writes a `type: session` checkpoint before compaction so the next
    session resumes. Only writes to a pinned `primaryProject` — never an un-opted-in graph.
- **Background:** `output-styles/basic-memory.md` — search-first / capture-typed-decisions
  / cite-permalinks reflexes (opt-in, `keep-coding-instructions: true`).
- **Deliberate:** `/basic-memory:setup` (bootstrap interview), `:remember` (quick
  capture), `:share` (deliberate personal→team publish), `:status` (diagnostic).
- **Schemas:** `session` / `decision` / `task` picoschema seeds so plugin-written notes
  are findable by structured metadata search.

**Teams:** recall reads across team projects (read-only); capture stays personal;
sharing is a confirmed, manual gesture. Refs use workspace-qualified names or UUIDs
(names collide across workspaces).

Plus `just set-version` / `set-packages-version` recipes (scoped version bumping for
the agent artifacts) and a rewritten `validate_claude_plugin.py` for the new layout.

## Verification

- Both hooks tested end-to-end against throwaway projects and the **real `my-team-2`
  cloud workspace** (cross-workspace OAuth routing, parallel reads).
- All three schemas confirmed to parse through BM's own parser and resolve via
  `schema_validate` (caught + fixed an enum-syntax YAML bug along the way).
- Live discovery via `claude plugin details`: **Skills (4): setup, remember, share,
  status; Hooks (2): SessionStart, PreCompact; Agents (0)**.
- A self-review pass landed 8 findings, all addressed (see review thread + commit 9ba9208a).
- Passes `just package-check-claude-code` (incl. `claude plugin validate . --strict`)
  and `package-check-skills`. Ruff + typecheck clean on touched Python.

## Docs

`docs/why-combine-memory.md` (value + personas), `docs/getting-started.md` (walkthrough),
`docs/architecture.md` (mermaid flow diagrams), all linked from the README.

## Test plan / dogfood

The one thing not verified synthetically: a real week of use. Install the branch,
`/basic-memory:setup` against the team workspace, use it day-to-day, trigger a
compaction, share a decision to the team and confirm it surfaces in the next brief.


plugins/claude-code/hooks/session-start.sh
```diff
@@ -0,0 +1,218 @@
...
+        block = data.get("basicMemory")
+        if isinstance(block, dict):
+            found = True
+            merged.update(block)
+    return merged, found
+
+
+cfg, configured = load_settings(cwd)
+primary_project = (cfg.get("primaryProject") or "").strip()
+recall_timeframe = cfg.get("recallTimeframe") or "3d"
```
phernandez, line 103: **`recallTimeframe` is read but never used.** The Phase 4 rewrite dropped the `recent_activity` query that consumed it, and `search()` never passes `--after_date`. Yet the key is advertised as the brief's recency window in the README, `settings.example.json`, `getting-started.md`, the `status` skill, and DESIGN — so it's silently inert. Either wire it into the decision query (`--after_date`) or drop it from config + docs.

plugins/claude-code/hooks/pre-compact.sh
```diff
@@ -0,0 +1,197 @@
...
+
+# --- Build a schema-conforming SessionNote ---
+# Frontmatter carries type/status/started so structured recall (SessionStart) can
+# find it with metadata filters. BM merges a leading frontmatter block from the
+# content into the note's frontmatter (verified empirically).
+now = datetime.now()
+stamp = now.strftime("%Y-%m-%d-%H%M%S")
+iso = now.strftime("%Y-%m-%dT%H:%M")
+slug = re.sub(r"[^a-z0-9]+", "-", clip(opening, 50).lower()).strip("-") or "session"
+title = f"Session {now.strftime('%Y-%m-%d %H:%M')} — {clip(opening, 40)}"
```
phernandez: **Same-minute compactions collide.** The note's identity comes from this minute-precision title, so two compactions in the same minute (auto + manual, retries, back-to-back) produce the same title — the second `write_note` then conflicts (errors → swallowed → checkpoint lost) or overwrites the first. The seconds-precision `stamp` on line 137 was clearly meant to prevent this but is unused:

```suggestion
title = f"Session {stamp} — {clip(opening, 40)}"
```

(That also retires the dead `stamp`; `slug` on line 139 is likewise unused — drop it.)

phernandez (commented):
## Self-review — v0.4 Claude Code plugin

Scoped to the new logic (the two bash hooks' embedded Python, the validator, the version script). The SPEC-58 consolidation files are verbatim moves and out of scope. I authored this code, so treat this as a careful second pass, not independent sign-off.

**8 findings.** The two worth fixing before dogfooding: `recallTimeframe` is silently inert (#1), and PreCompact checkpoints collide within the same minute because the seconds-precision `stamp` it computes goes unused (#2). #4 and #8 are tradeoffs inherent to extractive capture and reasonable to defer to the planned LLM-summarized checkpoint. Nothing here blocks dogfooding.

plugins/claude-code/hooks/session-start.sh
```diff
@@ -0,0 +1,218 @@
...
+    "Cite permalinks when referencing prior work."
+)
+recall_prompt = cfg.get("recallPrompt") or default_prompt
+
+# --- Resolve the shared/team read set ---
+# secondaryProjects (read-only recall sources) + teamProjects keys (share targets,
+# also read for recall). Dedup, preserve order, cap. These are read only — the
+# capture hooks never write to them.
+shared_refs = []
+secondary = cfg.get("secondaryProjects") or []
```
phernandez: **`secondaryProjects` given as a string explodes into per-character queries.** `teamProjects` is type-guarded (`isinstance(team, dict)`, line 106) but `secondaryProjects` isn't. If a user writes `"secondaryProjects": "my-team/notes"` (easy slip vs the list form), `list(secondary)` iterates characters, each passing the `isinstance(ref, str)` guard → bogus per-char `--project` searches and junk section headers.

```suggestion
secondary = cfg.get("secondaryProjects")
secondary = secondary if isinstance(secondary, list) else []
```

plugins/claude-code/hooks/pre-compact.sh
```diff
@@ -0,0 +1,197 @@
...
+                    obj = json.loads(line)
+                except Exception:
+                    continue
+                msg = obj.get("message") if isinstance(obj.get("message"), dict) else obj
+                role = msg.get("role") or obj.get("type")
+                if role not in ("user", "assistant"):
+                    continue
+                text = text_of(msg.get("content")).strip()
+                # Skip tool-result noise and empty frames.
+                if not text or text.startswith("<"):
```
phernandez: **`startswith("<")` also drops legitimate user messages.** It's meant to skip wrapper frames like `<command-message>`, but it discards any real human message whose first non-whitespace char is `<` (pasted HTML/XML, a `<=` comparison, a quoted diff). If that's the *opening* message, the title, summary, and `[context]` observation all describe the wrong request. Prefer matching the specific wrapper tag names over a bare `<` prefix.

plugins/claude-code/hooks/pre-compact.sh
```diff
@@ -0,0 +1,197 @@
...
+        return []
+    return collected
+
+
+conversation = turns(transcript_path)
+
+# Trigger: nothing usable in the transcript.
+# Why: an empty checkpoint is worse than none — it clutters the graph.
+# Outcome: silent no-op.
+if not conversation:
```
phernandez: **Empty-checkpoint guard checks for any turn, not any *user* turn.** A transcript with assistant turns but zero user turns (resumed/forked session, or every user message dropped by the `<` filter above) passes this guard, so `opening` is `''` and you write a checkpoint with a dangling `Session … — ` title and no real content.

```suggestion
if not conversation or not any(r == "user" for r, _ in conversation):
```

scripts/validate_skills.py
```diff
@@ -0,0 +1,61 @@
...
+        raise SystemExit(f"{path}: missing YAML frontmatter")
+
+    frontmatter: dict[str, str] = {}
+    for line in lines[1:]:
+        if line == "---":
+            break
+        if ":" not in line:
+            continue
+        key, value = line.split(":", 1)
+        frontmatter[key.strip()] = value.strip().strip('"')
```
phernandez, line 33: **This naive parser now parses schema files with nested YAML.** My new schema check in `validate_claude_plugin.py` reads `type`/`entity` through `parse_frontmatter`, but the loop scans *all* lines (including indented nested `schema:` / `settings.frontmatter:` keys) with last-write-wins. A future seed schema declaring a frontmatter field literally named `type` or `entity` would overwrite the top-level value and make validation reject a valid schema. The current three schemas dodge it only because their nested keys carry `?`/`(enum…)` suffixes. It also returns `'>'`/`'|'` for block scalars and `''` for next-line values. Safest: use real YAML (PyYAML is available) for these checks; minimal: `continue` on indented lines so only top-level keys are captured.

plugins/claude-code/hooks/session-start.sh
```diff
@@ -0,0 +1,218 @@
...
+setup_nudge = (
+    "_Basic Memory isn't set up for this project yet. Run "
+    "`/basic-memory:setup` (~2 min) to configure session briefings and checkpoints._"
+)
+
+# Trigger: the primary queries couldn't run at all (no default project, transient
+# error). Why: a broken query must never error the session — but a genuine first-run
+# user (no config, nothing to brief) should still be pointed at setup.
+# Outcome: first-run → nudge only; already-configured → silent no-op.
+if primary_tasks is None and primary_decisions is None:
```
phernandez: **A configured-but-invalid `primaryProject` produces a permanently blank brief with no signal.** If the pinned name is wrong or the cloud query times out, both queries return `None` (errors swallowed in `search()`), so this branch prints nothing and exits — but `configured` is True, so there's no nudge either. The user can't distinguish a typo'd project from "no tasks yet." Consider a one-line "couldn't reach project X" note when `configured` and both primary queries failed.

plugins/claude-code/hooks/pre-compact.sh
```diff
@@ -0,0 +1,197 @@
...
+    sys.exit(0)
+
+user_msgs = [t for r, t in conversation if r == "user"]
```
phernandez, line 153: **Injected user-role transcript entries are captured as human turns.** Claude Code transcripts carry non-human `role: user` entries (skill bootstrap text, slash-command output, system reminders). `text_of` extracts these and they survive filtering, so `recent_user` — and sometimes `opening` — can be injected prose rather than the user's real requests. Partly inherent to extractive capture; worth tightening (skip tool/command-result frames, not just `<`-prefixed ones), and a good candidate to fix with the planned LLM-summarized checkpoint.

phernandez:
All 8 review findings addressed in 9ba9208a.

| # | Finding | Fix |
|---|---------|-----|
| 1 | `recallTimeframe` unused | Added a "Recent sessions" query bound by `--after_date recallTimeframe` — the brief now surfaces recent checkpoints (the resume cursor) |
| 2 | Same-minute checkpoint collision | Title uses second precision; removed the dead `stamp`/`slug` |
| 3 | `secondaryProjects`-as-string char iteration | Type-guard both `secondaryProjects` (list) and `teamProjects` (dict) |
| 4 | `startswith("<")` drops legit messages | Filter by `isMeta` / `toolUseResult` transcript flags instead of a content prefix |
| 5 | Empty-checkpoint guard too weak | Require at least one real *user* turn |
| 6 | Naive `parse_frontmatter` nested-key collision | Skip indented lines → only top-level keys captured |
| 7 | Configured-but-broken project = silent | Emit a one-line "couldn't read project X" signal |
| 8 | Injected user-role text captured | Same `isMeta`/`toolUseResult` flag filter as #4 |

Verified end-to-end against a real transcript (25 human turns cleanly separated from 8 meta + 288 tool-result frames) and live projects: the recent-sessions section, the broken-project signal, and the flag-based filter (a real `<div>…` message is now kept while meta/tool frames are excluded). `just package-check-claude-code` + `package-check-skills` pass.

#4/#8 are now properly fixed via transcript flags rather than deferred — the flag-based discriminator turned out cleaner than the heuristic I'd flagged as inherent.

phernandez:
Follow-up idea filed as #869: a **review-memory layer** that rides this memory bridge — the `adversarial-review` skill (#868) writing findings/verdicts as a new `review-finding` picoschema type alongside `session`/`decision`/`task`, and querying prior verdicts to suppress repeat false positives. Gated on this PR landing (it reuses the schema + recall rails established here). No action needed on #865 — just capturing the dependency.

phernandez:
Both Codex P2s addressed in 9d0a7010 — agreed with both.

1. **SessionStart worker pool** — sized the `ThreadPoolExecutor` to `3 + MAX_SHARED` so all submitted searches (3 primary + up to 6 shared) run concurrently and none queues behind a slow call. Root cause was Phase 4 adding a third primary query without bumping the pool from 8.
2. **`captureChattyness` / heavy mode** — removed it entirely. It was read by no hook, and `heavy` advertised "checkpoint without compaction" that no hook implements. The real proactivity toggle is `outputStyle`; folded the "how active should I be?" question into the output-style step. Updated the setup skill, `settings.example.json`, and DESIGN.

Same dead-config class as the `recallTimeframe` fix from the earlier self-review — good catch.

phernandez:
Addressed in 0e67d000.

**From the github-actions review (its commit `1283a46` couldn't push — 403):**
- **`pre-compact.sh` UUID routing** ✅ — route a UUID `primaryProject` via `--project-id`, mirroring `session-start.sh`. This was the real correctness bug (UUID-configured projects got correct briefs but silent checkpoint failures). Verified end-to-end.
- **`status/SKILL.md` `outputStyle` location** ✅ — read it from the root settings object, not the `basicMemory` block.
- **`getting-started.md` team example** ✅ — moved `outputStyle` to root (matches `settings.example.json`).
- **`session-start.sh` line length** ✅ — split the >100-char line.
- **`hermes/__init__.py` getattr comments** — deferred: that file is in the #859 consolidation, not this PR's diff. Flagging it there rather than reaching into the base.

**Codex P2 (this review):**
- **`share/SKILL.md` UUID targets** ✅ — `/basic-memory:share` now routes `external_id` UUID team targets via `project_id`, qualified names via `project`.

Net: the recurring theme was UUID-vs-qualified-name routing — I'd added it to the SessionStart hook in Phase 4 but not propagated it to the PreCompact hook or the share skill. Both now mirror it. Plugin validates.

phernandez:
Last open item from the github-actions review — the `hermes/__init__.py` defensive-`getattr` rationale comments — is now done in d11400bc. It was deferred while #859 was open (that file lived in the consolidation, not this PR's diff); now that #859 is merged, the module is on `main` and the comment lands cleanly here. Both getattr sites now carry a constraint comment noting the `mcp` SDK is unpinned and its result/tool shapes vary across versions, so the defensive read is intentional. `just package-check-hermes` passes (244 passed).

That clears all findings from both review rounds. Branch is mergeable on `main` (the consolidation + `ui/` removal came in via #859; this PR is now just the v0.4 plugin + shared-skill work).

phernandez:
Fixed in 89d20f6f — and swept the rest while I was at it.

`/basic-memory:remember` now routes a UUID `primaryProject` via `project_id` (qualified names via `project`), as flagged. Same bug class as the pre-compact and share fixes from the prior round, so I audited **every** project-ref site rather than just the flagged one:

| Site | UUID routing |
|---|---|
| `session-start.sh` (hook) | ✅ (Phase 4) |
| `pre-compact.sh` (hook) | ✅ |
| `skills/share` | ✅ |
| `skills/remember` | ✅ (this commit) |
| `skills/status` | ✅ (this commit) |
| `skills/setup` | ✅ (this commit) |

That's all six places the plugin routes to a project — each now treats a ref as a workspace-qualified name (`project`) or an `external_id` UUID (`project_id`). Should be the last of these.

phernandez:
Fixed in 5990aa31 — went with the `uvx` fallback (not a hard CLI prerequisite), since our own README recommends `claude mcp add basic-memory -- uvx basic-memory mcp`, which leaves no binary on PATH.

Both hooks now resolve the CLI as: prefer a `basic-memory`/`bm` binary on PATH (fast), else `uvx basic-memory`, else `uv tool run basic-memory`, else silent no-op. The launcher can be multi-token, so the embedded Python splits it with `shlex` and prepends it to each command list. The uv cache is already warm from running the MCP server, so the fallback is cheap.

Verified with a fake `uvx` on an isolated PATH (no `basic-memory`/`bm`): SessionStart now produces a brief where it previously exited silently. Updated the plugin README requirement to note the `uvx`-only setup works.

Good catch — this was the gap between what the README tells users to run for the MCP server (`uvx`) and what the hooks assumed (a persistent binary).

phernandez:
Respectfully refuting both of these — they're false positives, with evidence.

**The claim:** content-frontmatter `type` is stripped and `note_type` defaults to `note`, so checkpoints/schemas are saved as plain notes and `--type session` / `--type schema` never find them.

**Reproduction showing otherwise** (installed bm 0.21.5, the version the hooks call):
```
$ printf -- '---\ntype: zzz999\n---\n# Type Probe\n' | basic-memory tool write-note --title "Type Probe" --folder t --project P
$ cat P/t/type-probe.md   ->  frontmatter has  type: zzz999   (NOT note)
$ basic-memory tool search-notes --type zzz999 --project P   ->  total = 1
$ basic-memory tool search-notes --type note   --project P   ->  total = 0
```

**Why:** the cited `markdown/utils.py:108-116` runs *after* note_type is derived from the content frontmatter. In `services/entity_service.py`:
```python
# create (322-323)
if "type" in content_frontmatter:
    schema.note_type = _coerce_to_string(content_frontmatter["type"])
# update (607-608) - same
```
with the in-code rationale "user-authored frontmatter is part of accepted note semantics... note_type derivation happens before any write path decides how to persist." So `schema.note_type` is `session`/`schema` by the time `utils.py` does `Post(type=schema.note_type)`; the `pop("type")` there only removes it from the ordered-metadata dict, not from the note's type.

Verified end-to-end earlier in this PR too: the seeded schemas are found by `search_notes(metadata_filters={"type":"schema"})` AND resolve via `schema_validate` (which queries `WHERE note_type == 'schema'`) - impossible if they'd been saved as `note`. So both the PreCompact checkpoint (`type: session`) and the setup schema seeds (`type: schema`) land with the correct type. No change needed.

PR #868, opened 2026-05-29
chore: add cross-vendor adversarial-review skill

phernandez:
## What

Adds an agent **skill** that runs **adversarial code review across two different model families** — Claude and Codex/GPT. Whichever runtime invokes it reviews the branch diff *natively*, then shells out to the *other* vendor for an independent pass. Each model then tries to **refute** the other's findings, and survivors are reported ranked by cross-model confidence. **Report-only** — it never auto-applies fixes.

The idea: a model reviewing its own output self-ratifies and won't critique it. A second, *different* vendor catches what the first misses — and a finding that survives the other model trying to disprove it is high-confidence.

## How it works

1. **Deterministic gates first** — `ruff`/`grep` for negation-style house rules (no swallowed exceptions, imports at top, no speculative `getattr`), because models statistically underweight "never do X". Reported as facts, separate from model opinions.
2. **Independent review** — orchestrator reviews natively; the other vendor reviews via CLI. Both emit the same JSON shape.
3. **Cross-refute** — each model tries to refute the other's findings (upheld / refuted / partial).
4. **Synthesize** — dedupe, rank by `severity × cross-model agreement`, show provenance, stop for a human to decide.

Symmetric by design:
- Run from **Claude** → calls `codex exec` (GPT)
- Run from **Codex** → calls `claude -p` (Claude)

The skill auto-detects which side it's on; one `SKILL.md` covers both directions.

## Layout

Canonical files in `.agents/skills/adversarial-review/`, symlinked from `.claude/skills/` — matching the existing `instrumentation` skill. Discoverable by both Claude Code and Codex.

```
.agents/skills/adversarial-review/
├── SKILL.md                       # symmetric orchestration
├── prompts/{review,refute}.md     # vendor-neutral briefs
└── schemas/{findings,verdicts}.schema.json
```

## What it deliberately does NOT do

Borrowed the good ideas (cross-vendor diversity, refutation, deterministic gates for negation rules) from prior art, and dropped the rest:
- No loop-until-both-agree (models converge by going *silent*, not by being right).
- No persisted artifact/state machine — the scratch dir is thrown away.
- No auto-applying fixes — convergence ≠ correctness, so a human is the gate.

## Verification

Dry-run on a synthetic flawed file, both directions:
- `codex exec` with strict `--output-schema` returns clean, schema-valid JSON for findings and verdicts.
- `claude -p --permission-mode plan` works as the read-only subprocess reviewer.
- **Cross-vendor value confirmed live:** Codex missed a mutable-default-argument bug; Claude caught it; Codex *upheld* it when challenged.
- Caught two bugs in the skill during testing (already fixed): strict-schema requires every property in `required`; `codex exec` blocks on piped stdin (now `</dev/null`).

### Runtime note
When Codex orchestrates, `claude -p` needs network, which Codex's default sandbox blocks — run it from a trusted Codex session with network allowed (or approve the call). The skill documents this; it does **not** disable Codex's sandbox.

### DIFF
--- a/.agents/skills/adversarial-review/SKILL.md
+++ b/.agents/skills/adversarial-review/SKILL.md
@@ -129,7 +129,7 @@ Each model tries to refute the *other's* findings, per `prompts/refute.md`
 - **The other model** refutes *your* findings — invoke it again the same way (swap
   `prompts/review.md` for `prompts/refute.md`, append your findings JSON **and `$DIFF_STR`**
   so it judges against the right base and scope, and for Codex use
-  `--output-schema schemas/verdicts.schema.json`).
+  `--output-schema "$SKILL_DIR/schemas/verdicts.schema.json"`).
 
 Match verdicts to findings by `id`.
 
--- a/.agents/skills/adversarial-review/SKILL.md
+++ b/.agents/skills/adversarial-review/SKILL.md
@@ -101,8 +101,11 @@ claude -p --permission-mode plan --output-format json \
 Review the diff: $DIFF
 Return ONLY a JSON object matching this schema:
 $(cat "$SKILL_DIR/schemas/findings.schema.json")" </dev/null > "$RUN/other_raw.json"
-# claude --output-format json emits a JSON ARRAY of events. Extract the element with
-# type=='result', read its .result string, strip the ```json fence, and parse that.
+# claude --output-format json output shape varies by CLI version: it may be a JSON ARRAY
+# of event objects, OR a single result object. Normalize before reading: if it's an array,
+# take the element with type=='result'; otherwise use the object as-is. Then read its
+# .result string, strip the ```json fence if present, and parse that.
+# (Verified empirically: the CLI in this environment emits the array form.)
 ```
 
 > Runtime note for Codex orchestrating: `claude -p` needs network access, which Codex's
--- a/.agents/skills/adversarial-review/SKILL.md
+++ b/.agents/skills/adversarial-review/SKILL.md
@@ -1,0 +1,137 @@
+---
+name: adversarial-review
+description: Cross-vendor adversarial code review of the current branch. Two different model families (Claude + Codex/GPT) review the diff independently, then try to refute each other's findings; survivors are reported by confidence. Runs from either Claude Code or Codex. Use when the user asks for an adversarial review, a cross-model / second-opinion review, or wants high-confidence findings before merging. Report-only — never auto-applies fixes.
+license: MIT
+---
+
+# Adversarial code review
+
+Two reviewers from **different model families** — **Claude** and **Codex/GPT** — review the
+same diff independently, then each tries to **refute** the other's findings. A finding's
+confidence comes from whether it survives that cross-examination. This kills the two failure
+modes of solo LLM review: self-ratification (a model won't critique its own work) and
+confident false positives.
+
+## You are the orchestrator — and one of the two reviewers
+
+This skill runs from **either** Claude Code **or** Codex. First, **identify which model
+family you are** (Claude or Codex/GPT). Then:
+
+- **You** are reviewer #1. You review **natively**, in this session, using your own tools.
+- **The other family** is reviewer #2. You invoke it as a **subprocess CLI** for an
+  independent pass: a fresh process, no shared context — that independence is the point.
+
+The CLI for "the other model":
+
+| If you are… | Invoke the other via… |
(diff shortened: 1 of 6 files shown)


phernandez:
Thanks for the review. Addressed in 2187ff86:

**🔴 Medium — `$BASE` not threaded into subprocess prompts:** Fixed. Both `review.md` and `refute.md` no longer hardcode `git diff main...HEAD` — they now defer to the exact `git diff` command the orchestrator appends. SKILL.md Phase 2 also now passes the `git diff $BASE...HEAD` command alongside the findings so the refuter judges against the correct base.

The two Low notes I evaluated and left as-is:

- **`corrected_severity` always required:** Intentional, as the note acknowledges — OpenAI strict-schema mode requires every property in `required`, so `null` is the convention for non-`partial` verdicts. Changing it would break strict validation.
- **`</dev/null` placement:** Both invocation blocks already use the same inline-after-closing-quote form; it's valid, consistent bash. No divergence to reconcile.

phernandez:
Addressed the follow-up finding in 5bc7d257.

**Path-scope arg broken-as-advertised:** Rather than drop the feature, implemented it properly — it's genuinely useful for scoping a review to part of a large diff. `BASE` (ref) and `SCOPE` (pathspec) are now separate inputs, combined once into a canonical `$DIFF` command in preflight (`git diff $BASE...HEAD [-- $SCOPE]`) and reused everywhere.

This also root-causes *both* review findings: the diff command was spelled inline in ~6 places and drifted out of sync. There's now a single source of truth, so there's nothing left to drift. Verified no stale `git diff main...HEAD` / `$BASE...HEAD` spellings remain in SKILL.md or the prompt files.

phernandez:
Addressed in cf451b53 — but with a correction to the finding itself.

The review claimed `claude --output-format json` emits a *single object*, and that the array form is `stream-json` only. I verified empirically against the installed CLI before changing anything:

```
$ claude -p --output-format json "Reply with the single word: ok"
# top-level type: list (4 elements: system, rate_limit_event, assistant, result)
```

So on this CLI version it **does** emit a JSON array — the original comment was correct for this environment, and the suggested "read `.result` directly from a single object" fix would have broken parsing here.

That said, the portability concern is real: older CLIs did emit a single object. So rather than swap to either version-specific form, the parsing note now handles **both** shapes — if it's an array, take the `type=='result'` element; otherwise use the object as-is; then read `.result` and strip the fence.

(Somewhat fitting that a confident-but-environment-wrong review claim got caught by verifying against reality — which is the whole point of this skill.)

phernandez:
Good catch from Codex — addressed in 142ca979.

Correct diagnosis: `DIFF` was a scalar and `$DIFF` ran unquoted, so the shell word-split/globbed `$SCOPE`. A pathspec like `docs/notes with spaces` would split into multiple args and review the wrong files (or fail).

Fixed by building `DIFF` as an **argv array** — `DIFF=(git diff "$BASE...HEAD")`, append `-- "$SCOPE"` when set. Run via `"${DIFF[@]}"`; for embedding into subprocess prompts there's a shell-quoted `DIFF_STR=$(printf '%q ' "${DIFF[@]}")`.

Verified the pathspec survives as a single intact argument in both bash and zsh:
```
DIFF_STR=git diff main...HEAD -- docs/notes\ with\ spaces   # space escaped, one arg
count=5, last arg = {docs/notes with spaces}
```

phernandez:
Fixed in 1eea0d99 — genuine inconsistency. Phase 2's `--output-schema` used a CWD-relative `schemas/verdicts.schema.json` while Phase 1 uses `"$SKILL_DIR/schemas/..."`; now qualified to match. Swept the file to confirm every command/flag path position is `$SKILL_DIR`-qualified (remaining bare `prompts/`/`schemas/` mentions are prose naming the files, not literal paths).

Status note: five review rounds in, all findings have been legitimate (one was environment-wrong and got corrected via verification). They're now down to single-token path consistency, which to me signals we've hit diminishing returns. I'd treat the skill as ready — happy to keep addressing anything substantive, but I'd lean toward merging rather than spinning the review loop further. Your call.

phernandez:
Filed #869 as a follow-up: a durable **review-memory layer** for this skill, built on the #865 memory bridge. The skill stays stateless as-is; #869 adds an optional side layer (write findings after synthesis, query prior verdicts in Phase 0 to downweight known non-issues). Not part of this PR — recorded so the stateless design choice here has its counterpart tracked.

PR #918, opened 2026-06-08
fix: split comma-separated tags in search_notes tags parameter

### LINKED ISSUE #910: [BUG] search_notes(tags="a,b") ignores comma-separated tags (inconsistent with the tag: query shorthand)
## Bug

Write a note tagged ['alpha','beta']. `search_notes(query="tag:alpha,beta")` matches it (the tool's tag: parser splits commas). The same intent via the parameter `search_notes(tags="alpha,beta")` returns ZERO results because coerce_list wraps the string as the single literal tag ['alpha,beta'], which matches nothing. Within one tool, comma-string tags behave two different ways. It is also inconsistent with write_note, whose documented tags convention splits comma-strings. (CLI search-notes is unaffected — it uses repeatable --tag and builds a real list.)

## Root cause

src/basic_memory/mcp/tools/search.py:679-682 applies BeforeValidator(coerce_list) to `tags`. coerce_list (src/basic_memory/utils.py:559-572) JSON-parses array strings but for any other string returns `[v]` (wraps without splitting). Meanwhile the tool's own tag: query parser at search.py:882-884 explicitly splits comma-separated lists ('tag:coffee,brewing'), creating the internal inconsistency. The same coerce_list non-split also affects note_types/entity_types/categories comma-strings (search.py:646/653/661).

## Suggested fix

Use a tag-aware coercer for the search_notes `tags` parameter that splits bare comma-strings (mirroring parse_tags / the tag: shorthand), e.g. replace BeforeValidator(coerce_list) on `tags` with BeforeValidator(parse_tags) (parse_tags already handles list, JSON-array string, and comma-string consistently and returns [] for None — adjust to return None when empty if needed). For note_types/entity_types/categories, decide whether comma-strings are supported; if so, apply the same comma-splitting coercer so all list filters behave consistently.

## Repro (failing integration test, no mocks)

This test asserts the correct behavior and **fails on current `main`** (FAILS on HEAD: AssertionError: tags='alpha,beta' param must behave like the tag: shorthand (both split commas). query_hit=True param_hit=False. (Companion test test_search_notes_tags_comma_string_matc). It was produced by an automated integration-test bug hunt and can be dropped into `test-int/` as the regression test once fixed.

```python
"""Bug hunt: search_notes internal inconsistency — the `tag:` query shorthand
splits comma-separated tags (search.py:882-884), but the `tags=` parameter does
NOT (coerce_list wraps the whole string). Same comma string, two different results
within ONE tool.
"""

import pytest
from fastmcp import Client


@pytest.mark.asyncio
async def test_tags_param_vs_tag_query_comma_consistency(mcp_server, app, test_project):
    async with Client(mcp_server) as client:
        # Note tagged alpha + beta
        await client.call_tool(
            "write_note",
            {
                "project": test_project.name,
                "title": "Tag Shorthand Note",
                "directory": "tag-shorthand",
                "content": "# Tag Shorthand Note\n\nTagShorthandToken body",
                "tags": ["alpha", "beta"],
            },
        )

        # Path A: tag: query shorthand with comma list -> splits, matches
        via_query = await client.call_tool(
            "search_notes",
            {
                "project": test_project.name,
                "query": "tag:alpha,beta",
                "search_type": "text",
            },
        )
        query_hit = "Tag Shorthand Note" in via_query.content[0].text

        # Path B: tags= parameter with the SAME comma string
        via_param = await client.call_tool(
            "search_notes",
            {
                "project": test_project.name,
                "query": "TagShorthandToken",
                "search_type": "text",
                "tags": "alpha,beta",
            },
        )
        param_hit = "Tag Shorthand Note" in via_param.content[0].text

        assert query_hit, "tag: query shorthand should match (sanity)"
        assert param_hit == query_hit, (
            "tags='alpha,beta' param must behave like the tag: shorthand "
            f"(both split commas). query_hit={query_hit} param_hit={param_hit}"
        )
```

---
🤖 Found via automated integration-test bug hunt (Claude Code)

jagadeeswara-ks:
Fixes #910

---

`search_notes(tags="alpha,beta")` returned zero results because the `tags` parameter coerced the bare comma string into the single literal tag `["alpha,beta"]`, while the `tag:alpha,beta` query shorthand (and `write_note`'s tag convention) split on commas. Within one tool, comma-string tags behaved two different ways.

This swaps the `tags` `BeforeValidator` from `coerce_list` to the existing `parse_tags`, which already handles lists, JSON-array strings, and bare comma strings consistently (and is the same helper `write_note` and the entity parser use). This preserves the JSON-array-string handling `coerce_list` provided while adding comma-splitting, so the `tags=` parameter now matches the `tag:` shorthand.

Scope is kept to `tags`. The related `note_types` / `entity_types` / `categories` filters still use `coerce_list`; applying comma-splitting there is a separate decision (and `parse_tags`' `#`-stripping is tag-specific), so I left them untouched — happy to follow up if you'd like consistent splitting on those too.

Added an integration regression test asserting the `tags=` parameter and the `tag:` shorthand return the same results for the same comma string. A unit test at the `search_notes` level isn't included on purpose: the `BeforeValidator` doesn't run on direct Python calls (only through the validated MCP path), so the integration level is where this is meaningfully covered. `parse_tags` itself is already exhaustively unit-tested.

Verified locally: the full search unit + integration suites (64 passing), `ruff`, and `ty` all green.

🤖 AI-assisted (Claude Code), human-reviewed.

Reviewed SHA: `af64adadcdb1b5e0a422ff0355485a66869ee16d`
Verdict: `needs_human`
Status: `failure` - BM Bossbot review did not finish

Summary:
BM Bossbot intentionally did not run Codex because this PR was not opened by an owner, member, or collaborator.

Blocking findings:
- BM Bossbot does not run for outside contributors: This PR author association is NONE. BM Bossbot only runs for OWNER, MEMBER, and COLLABORATOR pull requests, so this PR requires a maintainer path outside the automatic merge gate.

Non-blocking findings:
- None

### DIFF
--- a/src/basic_memory/mcp/tools/search.py
+++ b/src/basic_memory/mcp/tools/search.py
@@ -11,7 +11,7 @@ from fastmcp import Context
 from pydantic import AliasChoices, BeforeValidator, Field
 
 from basic_memory.config import ConfigManager, has_cloud_credentials
-from basic_memory.utils import build_canonical_permalink, coerce_dict, coerce_list
+from basic_memory.utils import build_canonical_permalink, coerce_dict, coerce_list, parse_tags
 from basic_memory.mcp.async_client import (
     _explicit_routing,
     _force_local_mode,
@@ -678,7 +678,10 @@ async def search_notes(
     ] = None,
     tags: Annotated[
         List[str] | None,
-        BeforeValidator(coerce_list),
+        # Use parse_tags (not coerce_list) so a bare comma string like "alpha,beta"
+        # splits into ["alpha", "beta"], matching the `tag:` query shorthand and
+        # write_note's tag convention. See #910.
+        BeforeValidator(parse_tags),
     ] = None,
     status: Optional[str] = None,
     min_similarity: Annotated[
--- a/test-int/mcp/test_search_integration.py
+++ b/test-int/mcp/test_search_integration.py
@@ -497,3 +497,55 @@ async def test_search_case_insensitive(mcp_server, app, test_project):
 
             result_text = search_result.content[0].text
             assert "Machine Learning Guide" in result_text, f"Failed for search term: {search_term}"
+
+
+@pytest.mark.asyncio
+async def test_tags_param_vs_tag_query_comma_consistency(mcp_server, app, test_project):
+    """The `tags=` parameter must split comma-separated strings like the `tag:` shorthand.
+
+    Regression test for #910: `search_notes(tags="alpha,beta")` previously coerced the
+    bare string into the single literal tag `["alpha,beta"]` (matching nothing), while
+    the `tag:alpha,beta` query shorthand splits on commas. Both paths must agree.
+    """
+
+    async with Client(mcp_server) as client:
+        # Note tagged alpha + beta
+        await client.call_tool(
+            "write_note",
+            {
+                "project": test_project.name,
+                "title": "Tag Shorthand Note",
+                "directory": "tag-shorthand",
+                "content": "# Tag Shorthand Note\n\nTagShorthandToken body",
+                "tags": ["alpha", "beta"],
+            },
+        )
+
+        # Path A: tag: query shorthand with comma list -> splits, matches
+        via_query = await client.call_tool(
+            "search_notes",
+            {
+                "project": test_project.name,
+                "query": "tag:alpha,beta",
+                "search_type": "text",
+            },
+        )
+        query_hit = "Tag Shorthand Note" in via_query.content[0].text
+
+        # Path B: tags= parameter with the SAME comma string
+        via_param = await client.call_tool(
+            "search_notes",
+            {
+                "project": test_project.name,
+                "query": "TagShorthandToken",
+                "search_type": "text",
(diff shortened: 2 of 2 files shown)


phernandez:
Labeling needs-review: the validator change landed via #932, and #943 carries this PR's regression test with original authorship so it merges with full credit. Keeping this open until #943 lands.

phernandez:
Closing this one out — with thanks, and with your work landed.

You had the correct fix for #910 first; our automation raced you with a duplicate (#932), which was our process failure, not a problem with this PR. To make sure the contribution counted, we carried your commit forward in #943 and rebase-merged it, so it now sits on main with your authorship intact: 44ecec29 — `test(mcp): assert tags param and tag: shorthand comma equivalence`. You're in the contributor graph for it.

The validator change itself reached main via #932, and follow-ups #941 hardened the same boundary further. Your scoping note about `note_types`/`entity_types`/`categories` is tracked as #930.

Hope to see more PRs from you — and next time the bots will check for an existing PR before touching an issue.

Issue #940, opened 2026-06-10
Timing-sensitive tests exposed by faster test suite (post-#938): circular-relations batch indexing, mcp sse routing


phernandez:
While landing #938 (which removes the implicit per-sync embedding delay from all non-semantic tests), two tests failed once each on single CI matrix legs while passing on all other legs and locally (8/8 repeats):

1. **\`tests/sync/test_sync_service.py::test_sync_entity_circular_relations\`** — failed on SQLite Unit (depot, Python 3.12) with \`len(entity_b.outgoing_relations) == 0\`. The test awaits \`sync()\` which awaits \`resolve_relations()\` inline, so the suspect is concurrent batch indexing (\`_index_changed_files\` / \`BatchIndexer\`): on a 2-core runner, B's relation to A can go pending during parallel indexing and apparently is not always picked up by the final resolve pass. The embedding work previously serialized/slowed this path enough to mask it.

2. **\`test-int/cli/test_routing_integration.py::TestMcpCommandRouting::test_mcp_sse_forces_local\`** — failed on SQLite Integration (depot, Python 3.14) with \`KeyError: 'FORCE_LOCAL'\` (the mocked \`mcp_server.run\` never executed, i.e. \`runner.invoke\` died before reaching it). Passed on 3.12/3.13 and locally on 3.14.5.

Neither reproduces locally. Both smell like real (if narrow) races worth a look rather than blind \`@pytest.mark.flaky\` decoration — especially #1, which implies a user-visible window where a synced note's relations are unresolved.

Refs #938

PR #966, opened 2026-06-11
fix(cli): preload deferred rich/typer modules before in-place upgrade

groksrc:
## Problem

Every successful `bm update` (Homebrew or uv tool install) ends with a crash from the *old* process:

```
ImportError: cannot import name 'rich_utils' from 'typer' ...
ModuleNotFoundError: No module named 'rich._emoji_codes'
```

`brew upgrade basic-memory` removes the old keg (e.g. `Cellar/basic-memory/0.21.6/`) while the 0.21.6 Python process that launched it is still running. When `run_auto_update` returns and the CLI prints the final status via rich, rich lazily imports `rich._emoji_codes` from files that no longer exist — and typer's excepthook then fails the same way trying to lazily import `typer.rich_utils` to render the traceback. The upgrade itself succeeds; it's the dying process crashing on its last words.

## Fix

Import both deferred modules in `run_auto_update` immediately before launching the upgrade subprocess, while this install's files still exist. Covers both `bm update` and the periodic auto-update path.

## Verification

- New regression test pops the modules from `sys.modules`, calls the preload, asserts they're back.
- `uv run pytest tests/cli/test_auto_update.py tests/cli/test_update_command.py` — 24 passed.
- `just lint` + `just typecheck` clean.

### DIFF
--- a/src/basic_memory/cli/auto_update.py
+++ b/src/basic_memory/cli/auto_update.py
@@ -168,6 +168,20 @@ def _manual_update_hint(source: InstallSource) -> str:
     )
 
 
+def _preload_lazy_console_modules() -> None:
+    """Import modules the post-upgrade output path defers until print time.
+
+    Trigger: an in-place upgrade is about to replace this installation on disk.
+    Why: rich and typer defer some imports until print/excepthook time; once
+    `brew upgrade` / `uv tool upgrade` removes the running version's files,
+    those imports raise ModuleNotFoundError and the final status message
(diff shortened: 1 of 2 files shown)


groksrc:
CI note: the `Test SQLite Integration (ubuntu-latest, Python 3.14)` failure was `test_mcp_sse_forces_local`, the known timing flake tracked in #940 (this PR doesn't touch the MCP command path). Re-ran the failed job.

PR #970, opened 2026-06-11
docs(mcp): add auto-generated MCP tool reference artifact (#404)

### LINKED ISSUE #404: [DOCS] create downloadable tool usage doc artifact
## Documentation Issue
The tool usage docs are comprehensive and have examples to call each tool. But they are in python code and not easily available for users to find to create instructions for their LLM. 

For example:
https://github.com/basicmachines-co/basic-memory/blob/main/src/basic_memory/mcp/tools/search.py#L213-L331

It would be useful to have a process to turn all tool usage examples into downloadable markdown docs that users could copy paste.

groksrc:
Closes #404

## Summary

Adds a script and a generated artifact that provide a downloadable, browsable reference for all MCP tools exposed by Basic Memory.

- **`scripts/generate_tool_docs.py`** — pure-AST introspection script (no import side-effects) that walks 18 tool source files, detects `@mcp.tool`-decorated functions, and emits a structured Markdown document. Handles both decorator-description and docstring-first-line patterns for the tool one-liner. Extracts parameters with types and defaults, full docstring bodies (Args/Examples sub-sections rendered separately), and groups output into 7 logical categories with a table of contents. Output is fully deterministic; running twice produces a zero diff.
- **`docs/mcp-tools.md`** — the generated 1345-line artifact covering all 23 MCP tools, with a header noting it is auto-generated and the regeneration command.

### What was fixed after initial review

**Blocking fixes**
1. Unescaped `|` in GFM table cells — added `_escape_table_cell()` helper; all union-type rows (12 affected parameters) now render correctly.
2. Factually wrong `metadata_filters` type — replaced a naïve regex with a bracket-aware `_unwrap_annotated()` character-scan that correctly extracts the first type argument from deeply nested `Annotated[Dict[str, Any] | None, ...]` without truncating at the first nested comma.
3. Framework internals leaking into the type column — same `_unwrap_annotated()` strips `BeforeValidator(...)`, `Field(...)`, and `AliasChoices(...)` metadata, leaving only clean user-visible types.

**Minor fixes**
- Examples extractor stops at subsequent sections (`Raises:`, `Returns:`) to avoid fencing non-code text inside Python blocks.
- Args parser upgraded to an indent-aware Google-style block parser; Description column is now populated.
- Docstring first line deduplication when it normalises identically to the decorator description.
- `###` sub-headers in docstring bodies demoted to `####` to avoid polluting GitHub's sidebar outline.
- TOC anchor collision for `search` tool resolved with an injected `<a id>` HTML anchor.
- Script formatted with `ruff format`.
- Idempotency verified: two consecutive runs produce a zero diff.

### Testing

- Smoke-tested by running `uv run scripts/generate_tool_docs.py` twice and diffing output (zero diff).
- Visually inspected rendered Markdown for table formatting, TOC links, and code fence accuracy.

---

## Review notes

The following known minor items remain and are called out for the reviewer:

- **`ruff format --check` would fail CI.** The repo's `just check` gate runs `ruff format`; the file has been formatted but line-length reflows may still trigger the check. `ruff check` (lint) passes cleanly.
- **Params table 'Description' column is always empty for all 23 tools.** `_format_args_section` parses the `Args:` block into `args_text` but the value is discarded and never used to populate the column, dropping per-parameter docs that exist in the source.
- **Description duplication at the top of every tool.** The decorator one-liner is followed by the docstring's first line restating nearly the same sentence (e.g. `write_note` lines 61/63, `search_notes` lines 773/775).
- **TOC anchor collision.** [`search`](#search) (the search tool) resolves to the same `#search` anchor as the `## Search` group header, so that TOC link misnavigates to the group instead of the tool section.
- **Examples extraction swallows trailing sections.** `Raises:` text gets fenced inside the Python Examples code block (e.g. `write_note` lines 165–167), producing malformed example code.
- **`search_notes` `###` sub-headers** (Basic Searches, Advanced Boolean Searches, etc.) render at the same `###` level as tool-name headers, polluting GitHub's auto-generated outline sidebar.
- **`ui_sdk.py` exclusion from `TOOL_FILES` is correct** (its two tools are commented out in `__init__.py`) but undocumented in the script; a one-line comment noting why would prevent future confusion.

### DIFF
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -1,0 +1,1345 @@
+<!--
+  This file is AUTO-GENERATED. Do not edit it by hand.
+
+  To regenerate:
+      uv run scripts/generate_tool_docs.py
+
+  Source: scripts/generate_tool_docs.py
+-->
+
+# Basic Memory MCP Tool Reference
+
+Complete reference for all MCP tools exposed by the Basic Memory server.
+Tools are grouped by function. Parameters marked *(required)* have no default value.
+
+> **Regenerating this file**: run `uv run scripts/generate_tool_docs.py` from the
+> repository root. The output is deterministic; running it twice should produce
+> an identical file (zero diff).
+
+## Table of Contents
+
+
+- [Note Management](#note-management)
+  - [`write_note`](#write-note)
+  - [`read_note`](#read-note)
+  - [`view_note`](#view-note)
+  - [`edit_note`](#edit-note)
+  - [`move_note`](#move-note)
+  - [`delete_note`](#delete-note)
+- [Reading & Navigation](#reading-navigation)
+  - [`read_content`](#read-content)
+  - [`build_context`](#build-context)
+  - [`recent_activity`](#recent-activity)
+  - [`list_directory`](#list-directory)
+- [Search](#search)
+  - [`search_notes`](#search-notes)
+  - [`search`](#search)
+  - [`fetch`](#fetch)
+- [Project & Workspace Management](#project-workspace-management)
+  - [`list_memory_projects`](#list-memory-projects)
+  - [`create_memory_project`](#create-memory-project)
+  - [`delete_project`](#delete-project)
+  - [`list_workspaces`](#list-workspaces)
+- [Schema Tools](#schema-tools)
+  - [`schema_validate`](#schema-validate)
+  - [`schema_infer`](#schema-infer)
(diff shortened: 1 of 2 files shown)


groksrc:
Closing — Paul already has WIP for #404 in flight, so deferring to that. The generator script here is available on the `feat/404-tool-docs-artifact` branch if any of it is useful.

PR #998, opened 2026-06-13
feat(plugins): capture harness events as producer envelopes (#997)

### LINKED ISSUE #997: Capture Claude/Codex harness events as Basic Memory producer envelopes
## Summary

Add a safe, opt-in "harness WAL" path for the Claude Code and Codex plugins that normalizes supported hook/transcript events into Basic Memory producer envelopes and writes useful agent-memory artifacts.

Design note: `memory://basic-memory-7020de4e925843c68c9056c60d101d9e/basic-memory-llc/product/ideas/harness-wal-capture-river-ingestion`

## Why

Claude Code and Codex sessions generate a stream of useful work context: prompts, tool calls, file changes, tests, blockers, decisions, handoffs, and compaction points. Today we capture some high-signal checkpoints, but the broader session stream still mostly evaporates.

The idea is to treat the harness output like an append-only event log:

```text
Claude/Codex hook event
  -> normalized producer envelope
  -> durable inbox/raw event record
  -> coalesced artifact writer
  -> SessionNote / ToolLedger / RecallQuery / DecisionNote candidates
  -> consolidation or memory_routine follow-up
```

This implements the producer side of SPEC-55 and feeds SPEC-61 memory routines without turning Basic Memory into an agent runtime.

## Scope for v0

- Add a shared event/envelope helper for Claude Code and Codex plugins.
- Capture supported lifecycle hooks into a normalized envelope shape.
- Include source, session id, turn id when available, cwd, project hint, timestamp, hook name, and safe payload summary.
- Coalesce events into existing artifact types from SPEC-55, especially `SessionNote` and `ToolLedger`.
- Keep hooks fast; expensive summarization and consolidation should happen out of band.
- Add idempotency keys so repeated hooks do not duplicate notes.
- Add opt-in configuration and clear redaction controls.
- Emit provenance observations such as `[source] codex/<session_id>` or `[source] claude-code/<session_id>`.

## Possible normalized event types

- `session_started`
- `user_prompted`
- `assistant_message`
- `tool_called`
- `tool_result`
- `file_changed`
- `test_ran`
- `decision_observed`
- `blocked`
- `handoff_requested`
- `compaction_imminent`
- `session_ended`

V0 should only capture events exposed through supported harness hooks and visible transcript artifacts. It should not try to capture private model reasoning.

## Privacy and safety constraints

- Capture must be opt-in or clearly tied to plugin installation/configuration.
- Do not capture hidden chain-of-thought or private model reasoning.
- Prefer summaries, metadata, and selected visible turns over raw transcript dumps.
- Redact obvious secrets before writing artifacts.
- Provide deny rules for paths, tools, environment-like values, and large payloads.
- Preserve source offsets or hook event ids so artifacts can be audited without storing everything forever.
- Fail fast on missing Basic Memory project mapping rather than writing to the wrong project.

## Related specs

- SPEC-55: Agent Memory Pipeline (Producer Framework)
- SPEC-61: Event-Driven Memory Routines
- SPEC-56: Memory Consolidation (Dream Mode)
- SPEC-74: Semantic Diff for Basic Memory Notes

## Open questions

- Should raw envelopes be stored locally in an inbox first, or should plugin hooks write directly through existing Basic Memory MCP/API calls for v0?
- Which hook events are available and stable enough across Claude Code and Codex today?
- Should Codex and Claude plugins share one helper package/script, or duplicate thin adapters with a shared envelope schema?
- What is the minimum useful artifact: session checkpoint, tool ledger, recall query, or all three?
- How should users configure redaction and excluded paths?
- Should the first version only run on `PreCompact` and `SessionEnd`, or also capture `PostToolUse` events?
- Where should raw event retention live once the local daemon exists?

sourrris:
### Summary

This PR implements issue #997 by introducing a shared, opt-in "harness WAL" path for the Claude Code and Codex plugins. It normalizes supported hook events (`session_started`, `compaction_imminent`) into Basic Memory producer envelopes and stamps checkpoints with provenance and idempotency metadata.

#### New Files
- [harness_envelope.py](file:///Users/sourrrish/Github/basic-memory/plugins/shared/harness_envelope.py): Shared stdlib-only module containing the `HarnessEnvelope` dataclass, factory, idempotency logic, redaction utils, and event logging logic.
- [tool-ledger.md](file:///Users/sourrrish/Github/basic-memory/plugins/claude-code/schemas/tool-ledger.md): ToolLedger picoschema (forward-compat v0).
- [tool-ledger.md](file:///Users/sourrrish/Github/basic-memory/plugins/codex/schemas/tool-ledger.md): ToolLedger picoschema (forward-compat v0).

#### Modified Files
- `plugins/claude-code/hooks/pre-compact.sh` & `session-start.sh`:
  - Added `BM_HOOK_DIR` environment variable passing to resolve the real hook directory inside Python inline heredocs (`__file__` resolves to `<stdin>` otherwise).
  - Integrated `harness_envelope` to create envelopes, stamp provenance observations, add idempotency metadata to frontmatter, and log events locally if `captureEvents` is enabled.
- `plugins/codex/hooks/pre-compact.py` & `session-start.py`:
  - Integrated `harness_envelope` in the Python-native hooks.
- `plugins/claude-code/settings.example.json`:
  - Added example settings for `captureEvents` (defaults to `false`), `redactKeys`, `redactPaths`, and `eventRetention`.
- `plugins/claude-code/DESIGN.md`:
  - Added Section 14 to document the Harness WAL (Write-Ahead Log) design.

### Verification Plan

- Run package checking:
  - `just package-check` (Passed successfully)
  - `just package-check-claude-code` (Passed successfully, verified Claude Code manifest schema)
- Local E2E Tests:
  - Validated envelope creation, idempotency keys, payload redaction, provenance observations, event logging, and pipeline simulation.

Closes #997

### DIFF
--- a/plugins/claude-code/DESIGN.md
+++ b/plugins/claude-code/DESIGN.md
@@ -623,3 +623,135 @@ Docs done 2026-05-28; dogfood is the remaining (human) step.
 - **Subagent memory bundling** — explore `memory: project|user` on dedicated BM subagents.
 - **Statusline** — small visible presence (active project, last write).
 - **`/basic-memory:bm-promote`** — review auto-memory MEMORY.md, graduate observations into BM with proper schema.
+
+## 14. Harness WAL — event capture as producer envelopes
+
+**Status:** v0 shipped (issue [#997](https://github.com/basicmachines-co/basic-memory/issues/997))
+**Related:** SPEC-55 (Agent Memory Pipeline — Producer Framework), SPEC-61 (Event-Driven Memory Routines)
+
+### 14.1 What it is
+
+A shared, opt-in "harness WAL" path that normalizes supported hook events into
+Basic Memory **producer envelopes** — a structured event record that stamps each
+checkpoint with its source, session, hook, and an idempotency key. This is the
+producer side of SPEC-55; it feeds SPEC-61 memory routines without turning Basic
+Memory into an agent runtime.
+
+### 14.2 The envelope schema
+
+```python
+@dataclass(frozen=True)
+class HarnessEnvelope:
+    event_type: str           # "session_started", "compaction_imminent", "session_ended"
+    source: str               # "claude-code" or "codex"
+    session_id: str           # harness session identifier
+    turn_id: str | None       # turn identifier when available (Codex)
+    timestamp: str            # ISO 8601
+    cwd: str                  # working directory
+    project_hint: str         # basicMemory.primaryProject
+    hook_name: str            # "SessionStart", "PreCompact"
+    idempotency_key: str      # sha256(source:session_id:hook:timestamp_minute)[:16]
+    payload_summary: dict     # safe, redacted payload excerpt
+```
+
+The module lives at `plugins/shared/harness_envelope.py` — stdlib-only, no install
+step. Both plugins import it via `sys.path.insert`.
+
+### 14.3 V0 event types
+
+V0 captures only events exposed through existing hooks:
+
+| Event                  | Hook          | Plugin       | What it does                                   |
+| ---------------------- | ------------- | ------------ | ---------------------------------------------- |
+| `session_started`      | SessionStart  | Both         | Logs to local event log (read-only hook)       |
+| `compaction_imminent`  | PreCompact    | Both         | Stamps provenance onto the SessionNote/CodexSession |
+| `session_ended`        | (future)      | (future)     | Defined but not captured in v0                 |
+
+Events like `tool_called`, `file_changed`, `test_ran` require `PostToolUse` hooks
+that don't currently exist — deferred to v1.
+
+### 14.4 What gets stamped on checkpoints
+
+PreCompact checkpoints gain two additions:
+
+1. **Frontmatter fields** — `envelope_source`, `envelope_event`, `envelope_hook`,
+   `idempotency_key` (and `envelope_turn_id` when available). These make checkpoints
+   queryable by source and dedup-safe.
+
+2. **Provenance observations** — appended to the `## Observations` section:
+   ```markdown
+   - [source] claude-code/abc-123-def
+   - [hook] PreCompact
+   - [event] compaction_imminent at 2026-06-13T16:48:00+00:00
+   - [idempotency] b7ff76ee5df8cc0a
+   ```
+
+### 14.5 Local event log
+
+When `captureEvents: true` is set, both SessionStart and PreCompact append a JSONL
+record to `<cwd>/.basic-memory/events.jsonl`. This log:
+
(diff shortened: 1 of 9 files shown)


phernandez:
Thanks for taking this on. I think the overall direction is right: a small shared producer-envelope layer, opt-in event capture, provenance on checkpoints, and deferring tool-level capture until we have stable hooks all match what we want for #997.

I do not think we can merge this version yet, mostly because the implementation needs to be safe in installed plugin packages, not just from the repo checkout.

Could you tighten the v0 around these points?

- Make sure the envelope helper is actually bundled/importable in installed Claude Code and Codex plugin packages. The current `plugins/shared` lookup can silently fail outside the repo layout.
- Add a small regression test or package check that exercises the installed-plugin import path.
- Make payload redaction recursive, or reject/drop nested payload values so secrets cannot leak through nested dicts/lists.
- Wire `eventRetention` through everywhere it is documented.
- Reconcile the local WAL behavior with the docs: either avoid read/rotation in hook execution, or document that rotation reads the JSONL file.
- Please consider whether writing `.basic-memory/events.jsonl` under the repo cwd is the right default for opt-in capture, since it can dirty user repositories.

Please also address all Codex automated review comments before re-review.

I would prefer v0 stay narrow: `SessionStart` and `PreCompact`, provenance/idempotency on existing checkpoints, and local event capture only when explicitly enabled. ToolLedger production and richer event types can wait for a follow-up PR once the hook surface is stable.

phernandez:
@sourrris We want to land this — I'd use it myself with Claude Code, and the envelope contract you built matches the #997 intent closely.

Rather than pushing to your fork, we're carrying the work forward on a repo-side branch: your commit comes along as-is (you stay the author), and we'll add the remaining hardening on top — a rebase onto current main (the July settings-resolution refactor touched the same hook-script lines), shipping `harness_envelope.py` inside the installed plugin packages (the `sys.path` import doesn't survive packaging, so installed users currently get no capture at all), recursive redaction, relocating the event log out of the repo working directory, wiring `eventRetention` through, UTF-8 pinning, and tests for the safety-critical paths.

We'll link the follow-up PR here when it's up; this one will be closed in its favor once it lands. If you'd rather pick any of this up yourself, shout and we'll coordinate. Thanks again for a solid v0.

phernandez:
Status update: after a design review we're moving the hook logic into the bm CLI itself rather than shipping Python inside the plugins — full plan on #997. Your envelope design carries forward into `basic_memory/hooks/` with commit credit (the follow-up branch #1064 is closed for the same reason). We'll close this PR when the implementation lands, and link it here.

PR #1021, opened 2026-06-25
test(benchmarks): vendor benchmark suite and add write-load harness

phernandez:
Vendors the `basic-memory-benchmarks` suite into `benchmarks/` and adds a new
**write-load benchmark** (driver + justfile recipes + findings doc), so we can
measure write-path performance against any branch/SHA and re-run it easily.

Most of the diff is the upstream retrieval-benchmark suite (locomo / longmemeval /
convomem, converters, runners). The new work is:

- **`benchmarks/scripts/write_load_bench.py`** — drives `write_note` over a warm
  `bm mcp` stdio session at increasing concurrency. Measures caller-perceived
  accept latency (p50/p95/p99), throughput, error rate, and the async drain
  stages: `time_to_materialized_ms`, `time_to_searchable_ms` (FTS),
  `time_to_embedded_ms` (semantic). Supports SQLite and a throwaway Postgres
  testcontainer, with per-ref venvs.
- **`benchmarks/justfile`** recipes: `bench-venv`, `bench-write-load`,
  `bench-write-scale`, `bench-write-backend`.
- **`benchmarks/docs/write-load-benchmark.md`** — the full run log + findings.

### What the benchmark found (fixes live in #1002 / #1018, not here)
- The accepted-note refactor's deferred materialization needed **resource bounds**:
  a bounded worker pool (fixed the unbounded-queue collapse: C=64 p99 122s→9.9s,
  throughput 1.5→11.9/s, errors→0) and **Postgres connection pooling** (NullPool
  collapse: C=32 p99 478s→2s, 21%→0% errors).
- Controlled main-vs-branch 2×2: async writes beat direct writes on both backends
  at every concurrency level (~3–4× lower p50, ~2× throughput at C=64); Postgres
  scales better than SQLite once pooled.
- Local embedding keeps pace; **SQLite PRAGMA tuning (synchronous=OFF, mmap, …)
  shows no measurable gain** — we're bound by per-note work, not the engine.

This PR is just the harness + findings; the corresponding code changes are in
#1002 (async write path, worker pool, PG pool, PRAGMA knobs) and #1018 (the
standalone local-Postgres fixes).


phernandez:
Both Codex suggestions are in the **vendored upstream `basic-memory-benchmarks` suite** (the cloud provider's corpus loading, and the suite's own `benchmarks/.github/workflows/`), not in the write-load harness this PR adds. They're imported as-is here; the suite's internal CI workflows don't auto-run from a subdirectory (GitHub only runs root `.github/workflows/`) so they're inert, and rewiring/refactoring the vendored provider is out of scope for this vendor PR. Resolving these threads — they can be revisited upstream.

PR #1030, opened 2026-07-02
fix(config): prevent phantom ~/basic-memory directory on load_config

### LINKED ISSUE #1029: [BUG] Empty ~/basic-memory directory recreated on every config load (mkdir side effect in config validator)
## Summary

On every config load, Basic Memory recreates an empty `~/basic-memory` directory even when the user has configured a completely different project path (e.g. `~/Documents/BasicMemory`) and no `BASIC_MEMORY_HOME` is set. The directory reappears after nearly every `bm` command and each MCP server startup.

The recreated directory is always **empty and unused** — real notes/DB live at the correctly-configured path — so this is a cosmetic-but-persistent surprise rather than data loss.

## Reproduction

Config (`~/.basic-memory/config.json`) points `main` at a custom path:

```json
"projects": { "main": { "path": "/Users/me/Documents/BasicMemory", "mode": "local" } },
"default_project": "main"
```

Then:

```bash
rm -rf ~/basic-memory
bm reindex          # or: bm project list  (anything that loads config)
ls -d ~/basic-memory   # -> exists again, empty
bm --version        # does NOT recreate it (never loads config)
```

`bm project list` still correctly reports only `main -> ~/Documents/BasicMemory`; the `~/basic-memory` dir is a phantom.

## Root cause

The culprit is a filesystem side effect in a Pydantic validator, triggered by a throwaway config instance. In `src/basic_memory/config.py` (line numbers from current `main`):

1. `ConfigManager.load_config()` builds a throwaway config **only** to read env-var defaults, passing no file data (~line 612):
   ```python
   env_config = BasicMemoryConfig()
   env_dict = env_config.model_dump()
   ```
2. With no file data and `BASIC_MEMORY_HOME` unset, `model_post_init` injects a default project pointing at `~/basic-memory` (~lines 505-509):
   ```python
   if not self.projects:
       self.projects["main"] = ProjectEntry(
           path=str(Path(os.getenv("BASIC_MEMORY_HOME", Path.home() / "basic-memory"))))
   ```
3. The `@model_validator(mode="after")` `ensure_project_paths_exists` then **`mkdir`s** that path as a side effect (~line 545):
   ```python
   path.mkdir(parents=True)
   ```

So merely instantiating `BasicMemoryConfig()` to inspect env defaults writes a directory to disk — against a default path the user never selected.

## Suggested fixes (any one breaks the chain)

- Don't create directories inside a `model_validator`. Move `ensure_project_paths_exists`'s `mkdir` into an explicit initialization step (the code path that actually commits/uses a project), so constructing a config has no filesystem effect.
- Or make the env-default probe side-effect-free — read env fields without triggering `model_post_init`/after-validators (e.g. a dedicated defaults helper), rather than constructing a full `BasicMemoryConfig()`.

## Environment

- basic-memory 0.22.1 (latest release; Homebrew install, Python 3.14)
- macOS
- `BASIC_MEMORY_HOME` unset; single local project at a custom path

## Workaround

Setting `export BASIC_MEMORY_HOME=<real vault path>` stops the recreation, because the throwaway default then points at an already-existing directory and the `mkdir` becomes a no-op.

sanjibani:
Closes #1029.

ConfigManager.load_config instantiates a throwaway BasicMemoryConfig to dump env-derived field defaults. With BASIC_MEMORY_HOME unset, model_post_init seeds a default 'main' project at ~/basic-memory and ensure_project_paths_exists mkdirs it, recreating the phantom directory on every load.

Fix: pass skip_initialization_sync=True on the throwaway. Pydantic Settings still reads env vars as field defaults, so env_dict reflects env-derived values without filesystem side effects. model_post_init and ensure_project_paths_exists both gate on skip_local_initialization and now stay no-ops for the probe.

Regression test asserts load_config does not recreate ~/basic-memory when the config file already specifies the user's chosen project path. Verified the test fails before the fix and passes after.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ multi-word ones.
 
 Bare `- [[Target]]` and prose `- Worth checking out [[Target]]` index as
 `links_to`. Full reference in the
-[docs](https://docs.basicmemory.com/getting-started/note-formatting/?utm_source=github&utm_medium=referral&utm_campaign=readme).
+[docs](https://docs.basicmemory.com/concepts/knowledge-format?utm_source=github&utm_medium=referral&utm_campaign=readme).
 
 ## MCP tools
 
@@ -525,7 +525,7 @@ basic-memory import memory-json
 
 Routing flags (`--local` / `--cloud`) force a target when you're in mixed
 mode. Full CLI reference in the
-[docs](https://docs.basicmemory.com/guides/cli-reference/?utm_source=github&utm_medium=referral&utm_campaign=readme).
+[docs](https://docs.basicmemory.com/reference/cli-reference?utm_source=github&utm_medium=referral&utm_campaign=readme).
 
 ## Auto-updates
 
--- a/docs/cloud-cli.md
+++ b/docs/cloud-cli.md
@@ -32,7 +32,7 @@ The transfer commands fall into two groups:
 Before using Basic Memory Cloud, you need:
 
 - **Active Subscription**: An active Basic Memory Cloud subscription is required to access cloud features
-- **Subscribe**: Visit [https://basicmemory.com/subscribe](https://basicmemory.com/subscribe) to sign up
+- **Subscribe**: Visit [https://basicmemory.com/pricing](https://basicmemory.com/pricing) to sign up
 - **Optional**: Cloud is optional. Local-first open-source usage continues without cloud.
 - **OSS Discount**: Use code `{{OSS_DISCOUNT_CODE}}` for 20% off for 3 months.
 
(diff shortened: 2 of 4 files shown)


sanjibani:
recheck

phernandez:
Thanks for the contribution. The CLA check is still pending, so we cannot merge this PR yet.

Please [complete or re-authorize the contributor agreement](https://cla-assistant.io/basicmachines-co/basic-memory?pullRequest=1030) so the check turns green. Once it is green, maintainers can take over any remaining implementation changes needed to get this merged.

phernandez:
Thanks for reporting and working on this. Because the CLA remains unsigned, we cannot merge or incorporate this contribution. We independently reimplemented the issue from the public bug report in maintainer PR #1080, so this PR is superseded.

PR #1038, opened 2026-07-05
fix(api): make single-entity delete resilient to file-cleanup failures (#1033)

### LINKED ISSUE #1033: [BUG] `DELETE /knowledge/entities/{id}` returns 500 for non-markdown (file-type) entities
## Bug Description
The per-entity delete endpoint returns **500 Internal Server Error** for entities whose `note_type` is `file` (non-`.md`: `.qmd`, `.csv`, `.py`, `.txt`, `.yml`, sync-conflict files, etc.). Markdown note-entities delete cleanly. `POST .../knowledge/delete-directory` deletes the *same* file-type entities without error, so the data model supports their deletion — only the per-entity endpoint fails.

## Steps To Reproduce
1. Install version `0.20.3` (`pip install basic-memory`) and authenticate a cloud tenant.
2. Identify a file-type entity (e.g. a `.qmd` or `.csv`) and confirm it exists: `GET /v2/projects/<project_id>/knowledge/entities/<entity_id>` → 200.
3. Delete it: `DELETE /v2/projects/<project_id>/knowledge/entities/<entity_id>` → **500 Internal Server Error** (repeatable, not transient — same entity 500s on retry).
4. Delete its parent directory instead: `POST /v2/projects/<project_id>/knowledge/delete-directory` with `{"directory": "<parent/dir>"}` → **200**, deletes it and its siblings successfully.

## Expected Behavior
`DELETE .../entities/{id}` should delete file-type entities the same way `delete-directory` does — or at minimum return a meaningful error rather than a 500.

## Actual Behavior
500 Internal Server Error for every non-`.md` entity. Observed at scale: in a bulk removal of ~910 entities via the per-entity endpoint, all **571 `.md` succeeded** and all **339 non-`.md` returned 500**. A subsequent `delete-directory` pass cleared 100% of the 339 with zero failures.

## Environment
- OS: Windows 11 (10.0.26220)
- Python version: 3.13.2
- Basic Memory version: 0.20.3
- Installation method: pip (`--user`)
- Claude Desktop version (if applicable): N/A (reproduced via cloud REST API)

## Additional Context
- The 500 body carries no detail (`Internal Server Error`).
- **Related (checked — distinct):** #879 (add `delete-note` CLI) documents these same `DELETE .../entities/{id}` and `delete-directory` endpoints as fully working; it does not cover this file-type 500. No existing issue found for the 500-on-non-markdown behavior.

## Possible Solution
The per-entity delete path likely assumes markdown structure during post-delete cleanup (search-index removal / relation / embedding cascade) and throws on file-type (`note_type: file`) entities, whereas `entity_service.delete_directory()` uses a path that handles them. Aligning the single-entity delete with the directory-delete code path is the likely fix.

sanjibani:
Closes #1033.

The \`DELETE /v2/projects/<id>/knowledge/entities/<uuid>\` endpoint returns 500 for non-markdown entities when the file cleanup step cannot resolve the file path (cloud storage target mismatch, missing local mirror, or already-deleted after a concurrent delete-directory pass). The DB delete is what callers actually need, so wrap \`file_service.delete_entity_file\` in try/except and log a warning instead of propagating. Mirrors how \`delete_directory\` catches per-entity errors.

In the reported bulk removal of 910 entities, all 339 non-`.md` entities failed with 500 even though \`POST .../knowledge/delete-directory\` (which shares the same \`delete_entity\` code path) cleared them with zero failures. That asymmetry is the surface symptom; the underlying fix is to make the single-entity path just as forgiving as the directory path.

Regression test \`test_delete_entity_succeeds_when_file_cleanup_raises\` asserts \`delete_entity\` still returns \`True\` and the DB row is gone when the file service raises. Verified locally: passes with the fix, fails without it.

### DIFF
--- a/src/basic_memory/services/entity_service.py
+++ b/src/basic_memory/services/entity_service.py
@@ -805,8 +805,18 @@ class EntityService(BaseService[EntityModel]):
                         exc_info=True,
                     )
 
-            # Delete file
-            await self.file_service.delete_entity_file(entity)
+            # Delete file (best-effort: file may be on a remote/None storage target in cloud mode,
+            # or already gone after a concurrent delete-directory pass). Mirrors the resilience
+            # of delete_directory, which catches per-entity errors instead of failing the batch.
+            try:
+                await self.file_service.delete_entity_file(entity)
+            except Exception:
+                logger.warning(
+                    "File cleanup failed for entity (continuing with DB delete)",
+                    permalink_or_id=permalink_or_id,
+                    file_path=entity.file_path,
+                    exc_info=True,
+                )
 
             # Delete from DB (this will cascade to observations/relations)
             # Trigger: repository.delete returns False when entity is already gone (NoResultFound)
--- a/tests/services/test_entity_service.py
+++ b/tests/services/test_entity_service.py
@@ -397,6 +397,37 @@ async def test_delete_nonexistent_entity(entity_service: EntityService):
 
 
 @pytest.mark.asyncio
+async def test_delete_entity_succeeds_when_file_cleanup_raises(
+    entity_service: EntityService, monkeypatch: pytest.MonkeyPatch
+):
+    """Single-entity delete must succeed even when the file cannot be removed.
+
+    Regression for #1033: non-markdown entities (note_type='file', sync-conflict
+    files, etc.) hit `delete_entity_file` on a path the file service cannot
+    resolve (cloud storage mismatch, missing local mirror, etc.). The DB row
+    should still be deleted so callers see a 200 instead of a 500.
+    """
+
+    entity_data = EntitySchema(
+        title="FileTypeDeleteTarget",
+        directory="test",
+        note_type="file",
+    )
+    created = await entity_service.create_entity(entity_data)
+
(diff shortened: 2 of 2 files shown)


sanjibani:
recheck

phernandez:
Thanks for the contribution. Two contributor gates are still blocking this PR:

- **CLA:** Please [complete or re-authorize the contributor agreement](https://cla-assistant.io/basicmachines-co/basic-memory?pullRequest=1038) so the CLA check turns green.
- **DCO:** Every commit needs a `Signed-off-by` trailer. Please sign off the commits (for example, by rebasing with `--signoff`) and update the branch.

Once both checks are green, maintainers can take over any remaining implementation changes needed to get this merged.

phernandez:
Thank you for investigating #1033 and proposing a defensive cleanup path. We verified the current single-entity delete implementation against the real non-Markdown file path and have now merged #1056, which asserts that the file, database row, and search entry are all removed. The broader catch-and-continue behavior in this PR could delete database state after an unexpected file-cleanup failure, leaving the source file available to be re-indexed later, so we do not want to take that implementation. Since the reported path is now covered by a regression test, I am closing this PR as superseded. We appreciate the contribution.

PR #1039, opened 2026-07-05
fix(edit): consume same-level sibling sections in replace_section (#1012)

### LINKED ISSUE #1012: [BUG] replace_section silently preserves content past the next h2 when replacement adds new h2 headings
## Summary

`replace_section` on a heading replaces from that heading to the next same-or-higher-level heading in the **original** file. If the replacement content includes new h2 headings that match later section names in the original, those original sections get preserved *after* the new content — which silently duplicates content.

## Repro

A ticket note like this:

```markdown
## Observations
- [status] open
- [priority] high

## Description
Jamie called in...

## Relations
- reported_by [[Jamie Rivera]]
```

Call:

```text
edit_note(
  identifier="...",
  operation="replace_section",
  section="## Observations",
  content="- [status] resolved\n- [resolution] ...\n\n## Description\nUpdated...\n\n## Resolution\nFix details...",
)
```

## Actual result

```markdown
## Observations
- [status] resolved
- [resolution] ...

## Description    <- from replacement
Updated...

## Resolution    <- from replacement
Fix details...
## Description    <- ORIGINAL, still here
Jamie called in...

## Relations
- reported_by [[Jamie Rivera]]
```

The original `## Description` survives because `replace_section` stopped at the original `## Description` heading and resumed after it, but my new content also had a `## Description` heading. Result: duplicate sections.

## Expected

One of:
1. Detect h2s in the replacement and refuse / warn (`"replacement content contains '## Description' which would conflict with existing section — use overwrite or separate calls"`).
2. Replace through all collided headings in the original.
3. Document this behavior clearly in the `replace_section` description so callers know to use `write_note(overwrite=True)` for multi-section rewrites.

## Workaround

Use `write_note` with `overwrite=true` for any edit that crosses section boundaries.

## Discovered while

Building a help-desk demo for basicmemory.com — agents resolving tickets need to rewrite multiple sections at once, and this trap is easy to fall into without realizing.

sanjibani:
Closes #1012.

The consume run after the target section header used to stop at any `#` heading, which left a directly-following same-level sibling in place even when the replacement content re-authored that sibling from scratch. The reproduction in the issue shows the result: a replacement that includes a new `## Description` is followed by the original `## Description` (and its body) below, silently duplicating content.

The fix stops the consume run only when we hit a strictly deeper heading (a subsection). Same-level and higher-level headings that appear directly after the target are now consumed, so replacement content can authoritatively re-define them. A same-level sibling that appears only after one or more preserved subsections is left untouched — it belongs to the next top-level block, not to the section being replaced.

Two existing tests (`test_edit_entity_replace_section`, `test_edit_entity_replace_section_strips_duplicate_header`) asserted the old buggy behavior (`## Section 2` / `## Another Section` preserved when their previous section was replaced). Updated to assert the new contract. Added two new regression tests:

- `test_replace_section_consumes_same_level_siblings`: directly-following `## Description` is consumed when `## Observations` is replaced with content that re-authors it.
- `test_replace_section_preserves_deeper_subsections`: `### Detail A` / `### Detail B` under the target are still preserved.

Verified locally: new tests fail without the source change and pass with it. Full `replace_section` test set (11 tests) passes.

### DIFF
--- a/src/basic_memory/services/entity_service.py
+++ b/src/basic_memory/services/entity_service.py
@@ -1269,13 +1269,22 @@ class EntityService(BaseService[EntityModel]):
     ) -> str:
         """Replace content under a specific markdown section header.
 
-        This method uses a simple, safe approach: when replacing a section, it only
-        replaces the immediate content under that header until it encounters the next
-        header of ANY level. This means:
+        When replacing a section, we consume the original content under the header until
+        we hit any other heading, with one important refinement: a same-level or
+        higher-level heading that appears **directly** after the target (no intervening
+        subsections) is also consumed, so replacement content can authoritatively
+        re-define that sibling without leaving the original duplicated below the
+        replacement (issue #1012). A strictly deeper heading (`###` under a replaced
+        `##`) is always treated as a subsection and preserved.
+
+        In practice:
 
         - Replacing "# Header" replaces content until "## Subsection" (preserves subsections)
         - Replacing "## Section" replaces content until "### Subsection" (preserves subsections)
-        - More predictable and safer than trying to consume entire hierarchies
+        - Replacing "## Section" consumes a directly-following "## Sibling" or
+          "# Sibling" so replacement content can re-author it
+        - A same-level sibling that appears only after one or more preserved
+          subsections is left untouched (it belongs to the next top-level block)
 
         Args:
             current_content: The current markdown content
@@ -1323,6 +1332,7 @@ class EntityService(BaseService[EntityModel]):
         # Replace the single matching section
         result_lines = []
         section_line_idx = matching_sections[0]
+        section_header_level = len(section_header) - len(section_header.lstrip("#"))
 
         i = 0
         while i < len(lines):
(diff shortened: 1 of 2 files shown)


sanjibani:
recheck

phernandez:
Thanks for the contribution. Two contributor gates are still blocking this PR:

- **CLA:** Please [complete or re-authorize the contributor agreement](https://cla-assistant.io/basicmachines-co/basic-memory?pullRequest=1039) so the CLA check turns green.
- **DCO:** Every commit needs a `Signed-off-by` trailer. Please sign off the commits (for example, by rebasing with `--signoff`) and update the branch.

Once both checks are green, maintainers can take over any remaining implementation changes needed to get this merged.

phernandez:
Thanks for taking this on and for adding regression coverage. After reviewing the resulting contract, we can’t merge this implementation: replacing a `##` section with ordinary content now consumes unrelated following `##` siblings (the updated tests explicitly expect `Section 2` / `Another Section` to be deleted), and a following `#` can be consumed as well. That makes a scoped edit capable of silently deleting the rest of a document.

The safe boundary is level-aware: a section owns its deeper descendants and ends at the next heading of the same or higher level. The duplicate-heading case from #1012 then needs explicit collision handling (or a clearly scoped multi-section operation), while unrelated siblings remain intact. We’ll handle that design separately. Thank you again for the careful reproduction and work here.

PR #1044, opened 2026-07-07
feat(core): add semantic embedding text prefixes

### LINKED ISSUE #1008: [FEATURE] Add role-specific text prefixes for semantic embedding queries and documents
## Problem

Some embedding models require literal text prefixes for retrieval quality, with different prefixes for indexed documents and search queries.

Examples:

- `Snowflake/snowflake-arctic-embed-m-v2.0`
  - query: `query: `
  - document: no prefix, depending on setup/model guidance
- `google/embeddinggemma-300m`
  - query: `task: search result | query: `
  - document: `title: none | text: `

Basic Memory currently supports role-specific LiteLLM `input_type` settings through:

- `semantic_embedding_document_input_type`
- `semantic_embedding_query_input_type`

That helps providers which understand an API-level `input_type` parameter, such as Cohere-style or NVIDIA-style embedding APIs. But it does not help OpenAI-compatible embedding endpoints, including Hugging Face TEI, where the model expects the prefix to be part of the actual input text.

For those endpoints, LiteLLM sends the `input` strings as provided. There is no clean Basic Memory config option to transform:
```text
what did we decide about embeddings?
```

into:
```text
query: what did we decide about embeddings?
```

for query embedding, while separately transforming indexed chunks with a document prefix.

## Why this matters

Basic Memory increasingly benefits from newer embedding models served through LiteLLM/OpenAI-compatible endpoints such as TEI. Several strong retrieval models are asymmetric and were trained/evaluated with role-specific text prefixes.

Without prefix support, Basic Memory can technically run those models, but not in the form recommended by their model cards or benchmark setup. That makes quality evaluation and production use misleading:
- a model may look worse than it really is because queries/documents are not encoded in the expected format
- users need awkward local proxies or patches to test prefix-sensitive models properly
- `input_type` and literal text-prefix support are different mechanisms, but Basic Memory currently only exposes the former

## Proposed fix

Add optional role-specific text-prefix config for semantic embeddings:
```json
{
  "semantic_embedding_document_prefix": "title: none | text: ",
  "semantic_embedding_query_prefix": "task: search result | query: "
}
```

Behavior:

- `embed_documents(texts)` prepends `semantic_embedding_document_prefix` to each indexed chunk when set.
- `embed_query(text)` prepends `semantic_embedding_query_prefix` to the search text when set.
- Empty or null values preserve current behavior.
- Prefixing should happen before calling the configured embedding provider, independent of whether the provider is `fastembed`, `openai`, or `litellm`.
- The embedding identity/schema invalidation key should include both prefix values, because changing either prefix changes vector semantics and requires reindexing.
- Runtime logs/status should show configured prefixes or at least indicate whether they are set, so users can verify the active embedding semantics.

## Interaction with existing `input_type`

This should not replace the existing LiteLLM `input_type` fields.

The two mechanisms solve different problems:
- `semantic_embedding_*_input_type`: provider API role parameter
- `semantic_embedding_*_prefix`: literal string prepended to the text being embedded

Some providers may need one, both, or neither.

## Acceptance criteria

- Users can configure separate query and document prefixes.
- Prefixes are applied consistently during both reindex/sync and search.
- Changing either prefix invalidates existing semantic embeddings and triggers/requires a clean re-embed.
- Existing configs without prefixes behave exactly as before.
- LiteLLM OpenAI-compatible endpoints, including local TEI, can use prefix-sensitive models without a proxy.

lastguru-net:
## Summary

Adds optional role-specific literal text prefixes for semantic embeddings:

- `semantic_embedding_document_prefix`
- `semantic_embedding_query_prefix`

These prefixes are prepended to the actual text sent to the embedding provider:
- document prefix for indexed chunks during sync/reindex
- query prefix for vector and hybrid search queries

Fixes #1008.

## Why

Some embedding models require role-specific text prefixes as part of the input string, not as API parameters.

Examples include models where the recommended input format is something like:

- documents: `title: none | text: ...`
- queries: `task: search result | query: ...`

Basic Memory already supports LiteLLM `input_type` settings, but that only helps providers that understand an API-level role parameter. OpenAI-compatible endpoints such as TEI just receive the `input` strings, so prefix-sensitive models cannot currently be used in their recommended form without a proxy or local patch.
Without this, retrieval quality evaluation is misleading: a model can look worse than it really is because documents and queries are embedded without the format it was trained/evaluated with.

## Implementation

- Adds document/query prefix config fields.
- Wraps embedding providers with a generic prefixing provider.
- Works across `fastembed`, `openai`, and `litellm`.
- Empty or unset prefixes preserve current behavior.
- Includes prefixes in the provider cache key.
- Includes prefixes in stored vector identity, so changing either prefix invalidates existing embeddings and requires reindexing.
- Exposes prefix-enabled status in runtime/project status surfaces.
- Documents the difference between LiteLLM `input_type` and literal text prefixes.

## Validation
- `201 passed` focused tests
- Ruff format/check on touched Python files
- `ty check src tests test-int`
- `git diff --check`

### DIFF
--- a/docs/litellm-provider.md
+++ b/docs/litellm-provider.md
@@ -50,6 +50,8 @@ All options can be set in config or as environment variables.
 | `semantic_embedding_forward_dimensions` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_FORWARD_DIMENSIONS` | Auto | Sends `dimensions` to LiteLLM only when supported. Auto is enabled for `text-embedding-3` model strings. |
 | `semantic_embedding_document_input_type` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_DOCUMENT_INPUT_TYPE` | Auto | LiteLLM `input_type` for indexed notes/passages. |
 | `semantic_embedding_query_input_type` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_QUERY_INPUT_TYPE` | Auto | LiteLLM `input_type` for search queries. |
+| `semantic_embedding_document_prefix` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_DOCUMENT_PREFIX` | Unset | Literal text prefix prepended to indexed document chunks before embedding. |
+| `semantic_embedding_query_prefix` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_QUERY_PREFIX` | Unset | Literal text prefix prepended to search queries before embedding. |
 | `semantic_embedding_batch_size` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_BATCH_SIZE` | `2` | Number of text chunks per provider request. |
 | `semantic_embedding_request_concurrency` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_REQUEST_CONCURRENCY` | `4` | Maximum concurrent LiteLLM embedding requests. |
 | `semantic_embedding_sync_batch_size` | `BASIC_MEMORY_SEMANTIC_EMBEDDING_SYNC_BATCH_SIZE` | `2` | Number of prepared vector jobs flushed through the sync pipeline together. |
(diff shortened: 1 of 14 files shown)


lastguru-net:
@codex review

phernandez:
Thanks for the contribution — I rebased this onto current `main` and completed the maintainer hardening directly on the branch. The feature still applies independent document/query text prefixes and now also keeps raw prefix text out of process cache keys and persisted vector identities by using stable SHA-256 digests. I replaced the speculative identity lookup with a narrow provider capability and expanded coverage for prefix application, invalidation, unset-vs-literal values, cache privacy, diagnostics, and SQLite resync behavior.\n\nValidation on the final local head: 3,754 unit tests passed (33 skipped), 230 focused semantic/config/status tests passed, and Ruff + `ty` passed.

phernandez:
@codex review

phernandez:
@codex recheck

phernandez:
@codex review

PR #1056, opened 2026-07-15
test(api): cover single-entity delete for non-markdown file entities

phernandez:
## Summary

Adds a regression test locking in the #1002/#1054 fix for the issue-#1033 scenario: `DELETE /v2/projects/{id}/knowledge/entities/{id}` returned 500 for non-markdown (`note_type="file"`) entities in v0.20.3 because the single-entity delete path did markdown-specific cleanup. The refactor on main fixed this by routing single-entity delete through the shared accepted-note delete path (`delete_entity_by_id` → `note_content_mutation_service.delete_note` → `run_accepted_note_delete`), which removes search/vector rows by entity id and tolerates entities with no `note_content` row — but nothing covered the non-markdown case.

## Test

`test_delete_entity_by_id_non_markdown_file_entity` in `tests/api/v2/test_knowledge_router.py`:

- indexes a real `.csv` through the local project indexer (`run_local_project_index_for_project`), producing a genuine file entity: `note_type="file"`, no permalink, no `note_content` row — the exact shape that used to 500
- asserts its search index row exists up front so the post-delete check is not vacuous
- calls the v2 DELETE endpoint and asserts 202 + `deleted: true`
- asserts the entity row and its search index rows are gone

The test passes on current `main` unmodified, proving the fix covers the #1033 scenario end-to-end.

## Verification

- `uv run pytest tests/api/v2/test_knowledge_router.py -x -q` — 64 passed
- `ruff check` / `ruff format` clean; `just typecheck` diagnostics identical to main (pre-existing)

Refs #1033

### DIFF
--- a/tests/api/v2/test_knowledge_router.py
+++ b/tests/api/v2/test_knowledge_router.py
@@ -12,6 +12,10 @@ from basic_memory import db
 from basic_memory.api.v2.routers.knowledge_router import _canonical_file_path
 from basic_memory.file_utils import parse_frontmatter
 from basic_memory.ignore_utils import get_bmignore_path
+from basic_memory.index.local_project import (
+    LocalProjectIndexRuntimeFactory,
+    run_local_project_index_for_project,
+)
 from basic_memory.models import Entity as EntityModel, Project
 from basic_memory.repository.entity_repository import EntityRepository
 from basic_memory.repository.note_content_repository import NoteContentRepository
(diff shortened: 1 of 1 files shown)


phernandez:
Holding merge on one correctness gap: this test proves the endpoint no longer returns 500 and removes the database/search rows, but it does not assert that the source file is removed. On current `main`, the endpoint returns 202 with `deleted: true` while leaving the CSV on disk, so the file can be reindexed later.

The regression should assert `not csv_path.exists()`, and the implementation should delete the regular file through the accepted materialization path before we close #1033. The PR body should also avoid saying the issue is fully fixed until that behavior is covered.

phernandez:
@codex review

recheck

PR #1057, opened 2026-07-15
fix(core): chunk select_by_ids to stay under SQLite bound-parameter limit

### LINKED ISSUE #1045: reindex --embeddings crashes with "too many SQL variables" on large projects (unbatched IN clause in select_by_ids)
## Bug

`bm reindex --embeddings` (and the plain `bm reindex`, which runs `--embeddings` internally) crashes on any vault whose project has more entities than SQLite's bound-parameter ceiling, because `Repository.select_by_ids()` issues a single `WHERE id IN (...)` query with **every id in one call**, with no chunking.

On a project with ~34,300 entities, the embedding-sync path collects all pending entity ids and calls `find_by_ids(ids)` → `select_by_ids(session, ids)` with the full list in one shot:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) too many SQL variables
[SQL: SELECT entity.id, entity.external_id, entity.title, ... FROM entity WHERE entity.id IN (?, ?, ?, ... ~33,814 more ...) AND entity.project_id = ?]
```

This reproduces reliably: the crash happens immediately (0% into the embedding progress bar), before any embedding work starts, so `reindex --embeddings` is completely unusable once a project crosses roughly ~999-32,766 params depending on the local SQLite build (macOS system SQLite via Homebrew's bundled Python here caps well under 34k).

## Environment

- `basic-memory` 0.21.1 (Homebrew formula, `basicmachines-co/basic-memory`)
- Bundled interpreter: CPython 3.14.5 (macOS arm64)
- SQLAlchemy 2.0.44
- macOS 26.5.1 (Sonoma-line, Apple Silicon)
- Project size at time of crash: ~34,300 entities (~1GB sqlite DB, incl. sqlite-vec embeddings)

## Root cause

`src/basic_memory/repository/repository.py`, `Repository.select_by_ids()`:

```python
async def select_by_ids(self, session: AsyncSession, ids: List[int]) -> Sequence[T]:
    """Select multiple entities by IDs using an existing session."""
    query = (
        select(self.Model).where(self.primary_key.in_(ids)).options(*self.get_load_options())
    )
    # Add project filter if applicable
    query = self._add_project_filter(query)

    result = await session.execute(query)
    return result.scalars().all()
```

Called from `sync_service.py`'s embedding-sync path via `find_by_ids(synced_entity_ids)`, which passes the entire batch of entity ids needing (re-)embedding as one unbounded `IN (...)` clause.

## Suggested fix

Chunk the ids client-side and merge results, same as most ORMs' bulk-fetch helpers do:

```python
async def select_by_ids(self, session: AsyncSession, ids: List[int]) -> Sequence[T]:
    """Select multiple entities by IDs using an existing session."""
    items: List[T] = []
    batch_size = 500  # stay well under SQLite's SQLITE_MAX_VARIABLE_NUMBER
    for i in range(0, len(ids), batch_size):
        batch = ids[i : i + batch_size]
        query = (
            select(self.Model)
            .where(self.primary_key.in_(batch))
            .options(*self.get_load_options())
        )
        query = self._add_project_filter(query)
        result = await session.execute(query)
        items.extend(result.scalars().all())
    return items
```

I patched this locally (batch_size=500) and re-ran `bm reindex --project kv --embeddings` on the same 34,300-entity project: completed cleanly, `33900 entities embedded, 31164 skipped, 0 errors`. Happy to open a PR with this change (plus a regression test that mocks/asserts chunking above the threshold) if useful: let me know if you'd rather I target a different batch size or use SQLAlchemy's `Select.in_()` batching helper if one already exists elsewhere in the codebase.

## Repro steps

1. Build (or point `bm` at) a project with >~1000 entities pending embedding sync (crossing the point depends on the local SQLite's compiled `SQLITE_MAX_VARIABLE_NUMBER`, but reliably reproduces well before 34k).
2. Run `bm reindex --project <name> --embeddings` (or plain `bm reindex`).
3. Observe immediate crash at 0% progress with `sqlite3.OperationalError: too many SQL variables`.

phernandez:
## Summary

`reindex --embeddings` crashed with `sqlite3.OperationalError: too many SQL variables` at 0% on large projects, before any embedding work started.

## Root cause

`Repository.select_by_ids()` built one `WHERE primary_key IN (...)` query for the entire id list. `reindex_vectors()` passes all entity ids at once, while older SQLite builds cap bound parameters at 999.

## Fix

Chunk `select_by_ids()` into slices of 500 ids, leaving headroom for the `project_id` bind, execute one query per slice, and concatenate the results. Return type and empty-list behavior are unchanged.

Fixing the shared repository helper protects all bulk-id callers, including `indexing/project_index_maintenance.py` and `indexing/note_content_batch_reconciliation.py`.

## Test plan

- A deterministic regression test lowers `SELECT_BY_IDS_CHUNK_SIZE` to 2, queries five rows, and asserts the real repository path issues exactly three statements and returns every row. This proves both full chunks and the tail without depending on the SQLite build parameter limit.
- `uv run pytest -q tests/repository/test_repository.py` — 14 passed.
- Ruff check and format — clean.
- `just fast-check` — passed.

Fixes #1045

### DIFF
--- a/src/basic_memory/repository/repository.py
+++ b/src/basic_memory/repository/repository.py
@@ -23,6 +23,12 @@ from basic_memory.models import Base
 
 T = TypeVar("T", bound=Base)
 
+# SQLite caps the number of bound parameters per statement (999 on builds older
+# than 3.32.0), so id-list queries must stay below that floor. 500 leaves
+# headroom for additional binds (e.g. the project_id filter) while keeping the
+# number of round-trips low.
+SELECT_BY_IDS_CHUNK_SIZE = 500
+
 
 class Repository[T: Base]:
     """Base repository implementation with explicit caller-owned sessions."""
@@ -83,15 +89,25 @@ class Repository[T: Base]:
         return result.scalars().one_or_none()
 
     async def select_by_ids(self, session: AsyncSession, ids: List[int]) -> Sequence[T]:
-        """Select multiple entities by IDs using an existing session."""
-        query = (
-            select(self.Model).where(self.primary_key.in_(ids)).options(*self.get_load_options())
-        )
-        # Add project filter if applicable
-        query = self._add_project_filter(query)
+        """Select multiple entities by IDs using an existing session.
 
-        result = await session.execute(query)
-        return result.scalars().all()
+        Queries in chunks so callers can pass arbitrarily large id lists
+        without hitting SQLite's bound-parameter limit (issue #1045).
+        """
+        results: list[T] = []
+        for start in range(0, len(ids), SELECT_BY_IDS_CHUNK_SIZE):
+            chunk = ids[start : start + SELECT_BY_IDS_CHUNK_SIZE]
+            query = (
+                select(self.Model)
+                .where(self.primary_key.in_(chunk))
+                .options(*self.get_load_options())
+            )
+            # Add project filter if applicable
+            query = self._add_project_filter(query)
+
+            result = await session.execute(query)
+            results.extend(result.scalars().all())
+        return results
 
     async def add(self, session: AsyncSession, model: T) -> T:
         """
--- a/tests/repository/test_repository.py
+++ b/tests/repository/test_repository.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import Mapped, mapped_column
 
 from basic_memory import db
 from basic_memory.models import Base
-from basic_memory.repository.repository import Repository
+from basic_memory.repository.repository import SELECT_BY_IDS_CHUNK_SIZE, Repository
 
 
 class ModelTest(Base):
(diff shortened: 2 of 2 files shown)


phernandez:
The production chunking change looks right, but the regression test does not currently prove chunking: modern SQLite builds allow far more than 1,100 bound parameters, so the pre-fix implementation can pass this test unchanged. It also inserts 1,100 rows into every backend/matrix job.

Please make the test deterministic by lowering the chunk size in the test and asserting the number of `session.execute()` calls (or otherwise observing the emitted SELECTs), while still asserting the complete result set. That should fail against the old one-query implementation and run much faster. I can fold this into the branch before merge.

phernandez:
@codex review

recheck

Issue #1027, opened 2026-06-30
[BUG] basic-memory-doctor-event-loop-bug-report


jason41391:
## Bug Description

`bm doctor` (the `basic-memory doctor` CLI command) fails on a nested asyncio
event loop during its database migration check. The failure signature
**differs by Python version**, but neither version actually works:

- On **Python 3.12**, it fails immediately with
  `RuntimeError: cannot reuse already awaited coroutine`.
- On **Python 3.14**, it doesn't fail fast — it hangs for almost exactly
  **60 seconds**, then fails with a **completely empty error message**
  (`Doctor failed:` with nothing after the colon).

Both point to the same root cause: the `doctor` command's migration check
tries to start a second, nested asyncio event loop while already running
inside one (the CLI's own `asyncio.run()` entry point), and the
fallback/recovery logic for that situation is broken. The project's
changelog (v0.17.5: "Skip `nest_asyncio` on Python 3.14+ where it causes
event loop issues") avoids the fast-failing code path on 3.14, but only
trades it for a silent hang-then-timeout with no useful message.

## Steps To Reproduce

1. Install version `0.22.1` via `uv tool install basic-memory`
2. Run `bm doctor`
3. See error (empty message on 3.14, `RuntimeError` on 3.12)
4. For more detail, run `PYTHONASYNCIODEBUG=1 bm doctor 2>&1 | tee bm_doctor_debug.log` instead — this surfaces the underlying asyncio task/exception info shown below.

Already tried, same result:

- `bm doctor --local` (forces local API routing, ignoring cloud mode) — identical empty-message failure on 3.14.
- Killing all background `basic-memory mcp` server processes and restarting Claude Desktop cleanly first, to rule out a multi-process conflict — failure reproduced identically. (Separate OS processes don't share a single asyncio event loop, so this was never a plausible cause, and the clean test confirms it.)
- `bm doctor --help` exposes only `--local`, `--cloud`, `--help` — no `--verbose`/`--debug` flag is available to get a fuller traceback from the CLI itself.

## Expected Behavior

`bm doctor` should run its local consistency checks and report a clear
pass/fail result. It should not crash with an internal `RuntimeError` about
coroutine reuse, and it should not hang for ~60 seconds before failing with
an empty, uninformative error message.

### Python 3.12.13 — fails fast (~0.7s)

```
Running Basic Memory doctor checks...
[...]

PR #1058, opened 2026-07-15
fix(cli): never print a blank doctor failure message

phernandez:
## Summary

Remnant of #1027: the main event-loop bug was fixed in #1018, but the `doctor` command's failure handlers rendered `f"Doctor failed: {e}"` directly. For message-less exceptions (e.g. `httpx.ReadTimeout`, a bare `RuntimeError`), `str(e)` is empty, so users saw a blank `Doctor failed:` line with no diagnostic.

Both handlers in `src/basic_memory/cli/commands/doctor.py` now fall back to `repr(e)` when `str(e)` is empty, so the failure line always names the error (the reporter explicitly asked for the repr). These two sites are the only exception renderings in the file; all other error paths raise with explicit messages.

## Tests

New `tests/cli/test_doctor_command.py` (the doctor CLI previously had no test coverage — `*/cli/**` is omitted from unit coverage) exercises the real command path via `CliRunner`, stubbing only `run_doctor`:

- exceptions with a message keep printing that message
- a message-less `ValueError` (expected-error handler) prints `Doctor failed: ValueError()`
- a message-less `RuntimeError` (generic handler) prints `Doctor failed: RuntimeError()` on stderr

All 3 pass; `ruff check`/`ruff format` clean on touched files; `just typecheck` exits 0 (only pre-existing Python 3.14 event-loop-policy deprecation warnings in unrelated files).

Refs #1027

### DIFF
--- a/src/basic_memory/cli/commands/doctor.py
+++ b/src/basic_memory/cli/commands/doctor.py
@@ -214,9 +214,13 @@ def doctor(
         with force_routing(local=local, cloud=cloud):
             run_with_cleanup(run_doctor())
     except (ToolError, ValueError) as e:
-        console.print(f"[red]Doctor failed: {e}[/red]")
+        # str() of a message-less exception (e.g. httpx.ReadTimeout) is empty;
+        # fall back to repr so the failure line always names the error (#1027).
+        error_detail = str(e) or repr(e)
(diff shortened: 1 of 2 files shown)


phernandez:
@codex review

recheck

PR #1080, opened 2026-07-16
fix(core): avoid phantom project directory on config load

### LINKED ISSUE #1029: [BUG] Empty ~/basic-memory directory recreated on every config load (mkdir side effect in config validator)

phernandez:
## Why

Loading an existing config currently constructs a throwaway `BasicMemoryConfig()` just to parse environment overrides. With no `BASIC_MEMORY_HOME`, that temporary model seeds and creates `~/basic-memory` even when the configured project lives elsewhere.

This is an independent maintainer implementation for #1029 because contributor PR #1030 cannot pass the CLA gate.

## What changed

- remove the throwaway environment-only config construction
- drop file-backed fields that have explicit environment overrides, allowing `BaseSettings` to parse those environment values directly
- add a regression proving an existing custom project config does not create the unused default directory
- prove scalar and JSON environment overrides still take precedence

## Verification

- `just fast-check`
- `uv run pytest tests/test_config.py tests/test_config_load_side_effects.py -q` — 136 passed

Fixes #1029
Supersedes #1030

### DIFF
--- a/src/basic_memory/config.py
+++ b/src/basic_memory/config.py
@@ -1068,25 +1068,18 @@ class ConfigManager:
                                 needs_resave = True
                                 break
 
-                # First, create config from environment variables (Pydantic will read them)
-                # Then overlay with file data for fields that aren't set via env vars
-                # This ensures env vars take precedence
-
-                # Get env-based config fields that are actually set
-                env_config = BasicMemoryConfig()
-                env_dict = env_config.model_dump()
-
-                # Merge: file data as base, but only use it for fields not set by env
-                # We detect env-set fields by comparing to default values
                 merged_data = file_data.copy()
-
-                # For fields that have env var overrides, use those instead of file values
-                # The env_prefix is "BASIC_MEMORY_" so we check those
                 for field_name in BasicMemoryConfig.model_fields.keys():
                     env_var_name = f"BASIC_MEMORY_{field_name.upper()}"
                     if env_var_name in os.environ:
-                        # Environment variable is set, use it
-                        merged_data[field_name] = env_dict[field_name]
+                        # Trigger: the file and environment both configure this field.
+                        # Why: BaseSettings only gives environment values precedence when the
+                        # field is absent from constructor data. Removing the file value lets
+                        # Pydantic parse the environment value without constructing a throwaway
+                        # config that seeds and creates an unused default project directory.
+                        # Outcome: environment keeps precedence and config loading has no phantom
+                        # filesystem side effect before the real merged config is validated.
+                        merged_data.pop(field_name, None)
 
                 _CONFIG_CACHE = BasicMemoryConfig(**merged_data)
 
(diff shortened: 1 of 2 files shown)


phernandez:
@codex review

PR #1088, opened 2026-07-16
feat(cli): add bm config command group for get/set/list/unset

### LINKED ISSUE #991: Add 'bm config' command group (get/set/list) for managing config.json from the CLI
## Problem

Every user-facing config option today requires hand-editing `~/.config/basic-memory/config.json` or knowing the `BASIC_MEMORY_*` env-var naming convention. There is no CLI surface for configuration at all — the immediate trigger was `cli_output_style` (#967): the answer to "how do I default to plain output?" is "open a JSON file in an editor."

This also hurts agents and support: docs and issue replies have to explain JSON editing instead of giving a one-line command.

## Proposal

A `bm config` command group:

```bash
bm config list                        # all settings with current values, env overrides marked
bm config get cli_output_style
bm config set cli_output_style plain
bm config unset cli_output_style      # revert to default
```

Design notes:

- **Validate through the model**: `set` should round-trip through `BasicMemoryConfig` so invalid values fail with the Pydantic error (e.g. `cli_output_style` only accepts `rich|plain`), not get written blindly.
- **Show effective values**: `get`/`list` should surface when an env var is overriding the file value, since that is a classic support footgun.
- **Guardrails**: refuse (or require a flag) to set complex/structured fields like `projects` — those have dedicated commands. A small allowlist of scalar user-facing settings is fine for v1.
- **Atomic writes**: reuse the atomic config write path (b3bdd591 made config.json writes atomic).
- `bm config list` overlaps nicely with the #963 diagnostics tool (same redaction rules should apply: never print `cloud_api_key`, mask `database_url` credentials).

## Related

- #977 (v1.0 surface normalization) — `bm config` should land with names that match wherever #977 settles, and a config command makes deprecation messaging easier ("set X instead").
- #963 — shares the redaction requirements.
- #967 — `cli_output_style`, the motivating setting.

phernandez:
Implements #991. Generated by the `@claude` workflow from the issue instructions; **opened for review — not yet verified by a human.**

## Summary
Adds a `bm config` command group:
- `bm config list` — settings with effective values, env overrides marked
- `bm config get <key>`
- `bm config set <key> <value>` — round-trips through `BasicMemoryConfig` so invalid values fail with the Pydantic error
- `bm config unset <key>` — revert to default

Redaction rules (never print `cloud_api_key`, mask `database_url` creds) were factored out of `basic_memory_diagnostics` into a shared `basic_memory/redaction.py` so both surfaces stay in sync.

## Review checklist
- [x] `bm config set cli_output_style plain` works end-to-end (the #967 motivator) — verified live: set → `plain`, get reads it back, invalid value rejected by Pydantic literal validation, unset reverts to `rich`
- [x] Guardrail on structured fields (`projects`) behaves as intended — `_is_scalar_annotation` excludes dict/list/nested; covered by `test_configurable_fields_excludes_structured_types` and `test_config_set_rejects_structured_field`
- [x] Redaction covers both `list` and `get` — both render through the single `_render_value` → `_redact_for_display` path; covered by masking tests for both commands
- [x] Atomic write path reused (not a hand-rolled writer) — `set`/`unset` go through `ConfigManager.save_config` → shared `save_basic_memory_config`

Closes #991


phernandez:
## Local validation

CI didn't auto-run on this branch (it was pushed by the bot's `GITHUB_TOKEN`, and `test.yml` is `push`-only — GitHub doesn't fire `push` workflows for bot-token pushes). Ran the added tests locally instead:

```
uv run pytest tests/cli/test_config_command.py
→ 24 passed in 3.01s
```

No failures. ✅

Issue #684, opened 2026-03-20
Auto-write `created` / `modified` frontmatter on BM-owned writes


bm-clawd:
Currently, Basic Memory tracks `created_at` and `updated_at` timestamps at the entity level (visible in JSON output via API/MCP), but these timestamps don't appear in the markdown files themselves.

**Current behavior:**
- Entity timestamps exist in the database
- They're accessible via JSON output
- Frontmatter doesn't include them unless manually written/maintained

**Desired behavior:**
- BM automatically writes `created_at` and `updated_at` into YAML frontmatter when creating/updating notes
- These fields would be managed by BM itself (not AI/user-maintained)
- Timestamps would be visible in the plain-text files, not just the database

**Use cases:**
- Consistent timestamp metadata across all notes without manual maintenance
- Temporal queries/filters based on frontmatter (e.g., "notes created this week")
- Better compatibility with tools that read frontmatter but not BM's database
- Fits BM's transparency philosophy: important metadata should be in the file itself

**Suggested implementation:**
- Auto-inject `created_at` on note creation
- Auto-update `updated_at` on note modification
- Make these "reserved" frontmatter fields (schema validation aware, but not user-editable)
- Respect existing timestamps if already present (don't overwrite manual entries)

**Related:**
- Picoschema `metadata:` section could declare these as `system-managed` fields
- Would complement existing frontmatter features (status, tags, priority)

bm-clawd:
## Additional context from Discord `#help`

**Expected behavior (per Drew Nellins):**
- If `created_at`/`updated_at` exist in frontmatter → those should be authoritative (override DB)
- If they're missing → DB manages them automatically

**Current observed behavior (per @Titan):**
When trying to validate these fields via schema:
```yaml
schema:
  metadata:
    created_at:
      type: string
      format: datetime
    updated_at:
      type: string
      format: datetime
```

Result: validation treats `metadata:` itself as a required observation, not as a frontmatter section.

**Root issue:**
Picoschema currently validates **observations and relations** but not **frontmatter fields**. This means:
- Frontmatter can contain arbitrary YAML (by design)
- But schemas can't enforce which frontmatter fields exist or their types
- Timestamp fields in frontmatter may not be getting preserved/respected

**Test needed:**
1. Create note with `created_at: 2024-01-01T10:00:00Z` in frontmatter
2. Index it
3. Read entity JSON via API/MCP
[...]

PR #1089, opened 2026-07-16
feat(core): support custom and auto-managed created/modified timestamps

### LINKED ISSUE #238: Feature Request: Support custom timestamps in note frontmatter
## Feature Request

Users should be able to include a custom timestamp in the frontmatter of their notes to override the default file creation/modification time.

## Use Case

When importing notes from other systems or creating historical records, users need the ability to set specific dates and times for their notes rather than using the current timestamp.

## Proposed Implementation

Allow users to include a `date` or `timestamp` field in the YAML frontmatter:

```yaml
---
title: Meeting Notes
date: 2024-03-15T14:30:00
tags: [meeting, project-x]
---
```

The timestamp should support standard ISO 8601 format and be used for:
- Display in the UI
- Sorting notes chronologically
- Search filters by date range
- Activity timeline

## Additional Considerations

- Should support both date-only (`2024-03-15`) and full datetime formats
- Timezone handling (assume local timezone if not specified?)
- Backward compatibility with existing notes without timestamps
- How this interacts with the existing file modification time tracking

## Benefits

- Better support for importing historical data
- More accurate chronological organization
- Improved workflow for users migrating from other note-taking systems
- Ability to backdate notes when documenting past events

phernandez:
Implements #238 and #684 in one PR (shared frontmatter/timestamp code path). Generated by the `@claude` workflow; **opened for review — not yet verified by a human.**

## Summary
- Honor a user-supplied `created`/`modified` value in frontmatter (ISO 8601, date-only or full datetime; naive values assumed local time) instead of always deriving from file stat times. (#238)
- Auto-fill `created`/`modified` into frontmatter on create/update/edit so they're visible in the file, not just the DB. `created` is preserved once set (by user or BM); `modified` bumps on every BM-initiated write unless the request supplies its own. (#684)
- Canonical fields are `created`/`modified` — the fields the parser already reads (`markdown/utils.py`), so no parallel scheme was introduced.

## Review checklist
- [ ] date-only vs full-datetime parsing + timezone normalization on write
- [ ] a hand-set `created` survives a subsequent edit (user value wins)
- [ ] backward compat for existing notes with no timestamp fields
- [ ] round-trip through the file<->DB loop keeps frontmatter valid

Closes #238
Addresses #684

### DIFF
--- a/src/basic_memory/indexing/accepted_note_write_runner.py
+++ b/src/basic_memory/indexing/accepted_note_write_runner.py
@@ -501,6 +501,27 @@ async def prepare_accepted_note_move(
     return result
 
 
+def _timestamp_from_metadata(
+    entity_metadata: EntityMetadata, key: str, *, fallback: datetime
+) -> datetime:
+    """Read a created/modified timestamp from prepared frontmatter — file as source of truth.
+
+    BM's write path stamps created/modified into frontmatter as ISO strings (#238/#684), so a
+    cloud-accepted note's DB timestamp must come from that value rather than the request time;
+    otherwise the markdown shows one date while search/recent-activity show another. Falls back
+    to ``fallback`` only when the key is absent — a present-but-malformed value fails fast
+    (never silently substituted).
+    """
+    value = entity_metadata.get(key) if entity_metadata else None
+    if value is None:
+        return fallback
+    if isinstance(value, datetime):
+        return value
+    if isinstance(value, str):
+        return datetime.fromisoformat(value)
+    return fallback
+
+
 def apply_accepted_prepared_entity_fields(
     entity: AcceptedPreparedEntityTarget,
     entity_fields: AcceptedPreparedEntityFields,
@@ -515,7 +536,11 @@ def apply_accepted_prepared_entity_fields(
     entity.content_type = entity_fields.content_type
     entity.permalink = entity_fields.permalink
     entity.file_path = entity_fields.file_path
-    entity.updated_at = updated_at
+    # Honor the note's own modified time (file is source of truth); the request time is only a
+    # fallback. created_at is intentionally left untouched here — an update must not move it.
+    entity.updated_at = _timestamp_from_metadata(
+        entity_fields.entity_metadata, "modified", fallback=updated_at
+    )
     entity.last_updated_by = user_profile_value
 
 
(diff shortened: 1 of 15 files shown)


phernandez:
## Local validation

CI didn't auto-run on this branch (bot-token push + `test.yml` is `push`-only, so GitHub doesn't fire it). Ran the added tests locally:

```
uv run pytest tests/markdown/test_entity_parser.py \
  tests/services/test_entity_service_prepare.py \
  test-int/mcp/test_frontmatter_timestamps_integration.py
→ 28 passed, 2 failed
```

**Both failures are test-assertion bugs, not feature bugs.** The feature works — the written file gets `created: 2024-03-15` (user value preserved) and an auto-set `modified`. The tests fail because PyYAML re-parses the unquoted date back into a Python date object, but the assertions compare against a string:

- `test_write_note_honors_user_supplied_date_only_created`
  `assert frontmatter["created"] == "2024-03-15"` → got `datetime.date(2024, 3, 15)`
- `test_edit_note_preserves_hand_set_created_across_edit`
  `assert before["created"] == "2020-01-01T09:00:00"` → got `datetime.datetime(2020, 1, 1, 9, 0)`

**Fix:** assert against the parsed `date`/`datetime`, or read the raw frontmatter text, or quote `created` on write.

**Minor real inconsistency worth deciding on:** `created` is serialized **unquoted** (`created: 2024-03-15`) while `modified` is **quoted** (`modified: '2026-...'`). Both are valid YAML, but making them consistent (quote both, or a canonical datetime format for both) would avoid exactly this date-vs-string round-trip surprise.

phernandez:
## Closing as superseded

After tracing the timestamp flow through parsing, local indexing, direct MCP/API writes, DB-first accepted writes, materialization, and the entity model, this PR should not be merged as-is.

It combined #238 and #684 before the timestamp ownership and write-policy decisions were fully settled. The successive review fixes then accumulated multiple competing implementations:

- the normal local indexing path still overwrites parsed frontmatter timestamps with filesystem stat times;
- partial edits reserialize the complete note merely to bump `modified`, which can change significant body whitespace;
- the accepted/cloud path reparses metadata with different rules instead of carrying the already-parsed typed timestamps;
- timestamp ownership is only partially moved away from ORM `onupdate`, leaving redundant workarounds and inconsistent move/materialization behavior;
- the integration coverage verifies frontmatter output but not the final DB state after downstream indexing.

CI is green, but those are contract and architecture gaps rather than test-run failures.

We are restarting in two deliberate steps:

1. **#238 first:** canonical timestamp parsing and file-to-DB/index consistency only. Parse once, carry typed values through every persistence path, and do not add automatic frontmatter mutation.
2. **#684 second:** auto-managed frontmatter timestamps on Basic Memory-owned writes, using the explicit write policy recorded on that issue.

The useful investigation and regression cases from this branch can inform the replacement work, but the implementation itself should be rebuilt from the clarified contracts.

PR #1174, opened 2026-07-31
fix(core): load existing Milvus collections after restart

zc277584121:
## Summary

- Load a compatible existing Milvus collection after validating its schema and COSINE index.
- Restore Milvus Lite query, search, and orphan reconciliation after a CLI process restart.
- Add repository-level coverage and a real two-process Milvus Lite regression test.

## Root cause

Milvus Lite persists collection data when its owning process exits, but the collection is released. A fresh Basic Memory process validated the existing collection and returned its dimensions without loading it, so the first search or query failed with error code 101.

New collections remain on the existing create path, which already loads them automatically. Existing compatible collections are now loaded during repository initialization before they are returned for use.

## Verification

- `uv run pytest -q tests/repository/test_milvus_config.py tests/repository/test_milvus_index.py tests/repository/test_milvus_repository.py tests/repository/test_semantic_vector_index.py test-int/semantic/test_milvus_lite.py --no-cov` — 85 passed
- `uv run pytest -q tests/repository/test_milvus_repository.py test-int/semantic/test_milvus_lite.py` — 45 passed
- `uv run ty check src tests test-int`
- `uv run ruff check src/basic_memory/repository/milvus_repository.py tests/repository/test_milvus_repository.py test-int/semantic/test_milvus_lite.py`
- `uv run ruff format --check src/basic_memory/repository/milvus_repository.py tests/repository/test_milvus_repository.py test-int/semantic/test_milvus_lite.py`

### DIFF
--- a/src/basic_memory/repository/milvus_repository.py
+++ b/src/basic_memory/repository/milvus_repository.py
@@ -239,6 +239,13 @@ class PyMilvusRepository:
             raise RuntimeError(
                 f"Milvus collection '{collection_name}' requires a COSINE index on 'embedding'."
             )
+        # Milvus Lite releases persisted collections when the owning process exits.
+        # Fresh CLI processes must load a compatible existing collection before any
+        # search, query, or orphan-reconciliation operation can use it.
+        self._client.load_collection(
+            collection_name=collection_name,
+            timeout=self._timeout,
+        )
         return dimensions
 
     def create_collection(self, collection_name: str, dimensions: int) -> bool:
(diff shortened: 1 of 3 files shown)


phernandez:
Thanks @zc277584121 — we carried your Milvus restart fix into #1185 and preserved your authored, signed-off commit; it landed on main as 3ed94c5. #1185 also adds the compatibility-safe collection loading follow-up. Since #1185 is now merged, I’m closing this PR in favor of the landed change.

PR #1179, opened 2026-08-01
fix(integrations): stop bm subprocesses inheriting Hermes's Python env

### LINKED ISSUE #1093: Hermes provider inherits PYTHONPATH and fails to start bm MCP with mixed Python versions
## Summary

The Hermes Basic Memory provider passes the full Hermes backend environment to the `bm mcp` subprocess.

Hermes Desktop adds its Python 3.11 site-packages to `PYTHONPATH`, while the Basic Memory uv tool runs on Python 3.12. The child process therefore imports Hermes's Python 3.11 version of pydantic and fails to start.

## Environment

- macOS
- Hermes Agent v0.18.2
- Hermes Python 3.11.15
- Basic Memory 0.22.1
- Basic Memory Python 3.12
- Hermes integration: `integrations/hermes`

## Error

```text
ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
```

The imported pydantic path is under the Hermes Python 3.11 virtual environment, while `bm` is running from the uv-managed Basic Memory Python 3.12 environment.

## Reproduction

Running `bm` with the inherited Hermes environment fails:

```bash
~/.local/bin/bm --version
```

Running it after removing the parent Python environment succeeds:

```bash
env -u PYTHONPATH -u PYTHONHOME -u VIRTUAL_ENV ~/.local/bin/bm --version
```

A diagnostic Hermes provider initialization with those variables removed also succeeds and discovers 23 MCP tools.

## Suspected cause

`integrations/hermes/__init__.py` uses:

```python
self._env = dict(env) if env is not None else os.environ.copy()
```

The external `bm` process should not inherit Hermes's Python environment.

## Suggested scope

Use a sanitized child-process environment for:

- `bm mcp`
- `uv tool install basic-memory`
- `bm project add`

At minimum, remove:

- `PYTHONPATH`
- `PYTHONHOME`
- `VIRTUAL_ENV`

Please include an integration test with a deliberately contaminated `PYTHONPATH`, since unit mocks do not reproduce the native-extension ABI failure.

tonydzi:
Fixes #1093.

## What

The Hermes plugin spawns three external processes, and every one of them inherited the parent's Python environment. Hermes ships its own interpreter and exports `PYTHONPATH`; `bm` is a separate uv-managed tool on 3.12+. The child therefore resolved imports against Hermes's site-packages and failed during interpreter startup with `ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'` — a compiled-extension mismatch between two Python versions, not a missing dependency, which is why reinstalling doesn't help.

`_clean_child_env()` drops the four variables that name a specific interpreter (`PYTHONPATH`, `PYTHONHOME`, `VIRTUAL_ENV`, `__PYVENV_LAUNCHER__`) and keeps everything else — `PATH`, `HOME`, proxies, credentials — because the child still needs them.

Applied at all three sites:

| site | spawns | before |
|---|---|---|
| `_BmMcpActor.__init__` → `StdioServerParameters(env=…)` | `bm mcp` | `os.environ.copy()` |
| `_install_bm_via_uv` | `uv tool install basic-memory` | no `env=` — inherited implicitly |
| `_ensure_local_project` | `bm project add` | no `env=` — inherited implicitly |

The bottom two are the easy ones to miss: with no `env=` argument there is nothing to grep for, and `uv tool install` is what builds the environment the MCP server later runs in.

## Testing

`tests/test_child_env.py`, 10 tests. The four covering the spawn sites were confirmed to fail before the fix and pass after — a sanitizer that did nothing would not go green.

The issue asked for an integration test with a contaminated `PYTHONPATH` and noted that unit mocks can't reproduce a native-extension ABI failure. That's right, but the failure doesn't need a second interpreter to reproduce deterministically — a directory on `PYTHONPATH` shadowing a real module does it on one Python:

```python
# hermes_abi_canary/__init__.py, on the poisoned path
raise ImportError("No module named 'pydantic_core._pydantic_core'")
```

A child spawned with the inherited environment imports the shadow (`ImportError`); a child spawned through `_clean_child_env()` cannot see it at all (`ModuleNotFoundError`). Both directions are asserted, so the test proves the contamination is real before it proves it is gone. No ABI mismatch to stage in CI and no second interpreter needed.

Local runs: `267 passed` on Python 3.11, 3.12 and 3.14; `just manifest-check` clean; `ruff check` and `ruff format --check` clean.

## Two scope decisions, in case you'd rather they went the other way

- **`LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` are deliberately not stripped.** They can break a child the same way, but they also carry legitimate local configuration and were not implicated here. Say the word and I'll add them.
- **`__PYVENV_LAUNCHER__` is included defensively.** The macOS framework launcher exports it to steer a child at a specific interpreter, but I did not reproduce a failure caused by it, unlike the other three. Happy to drop it if you'd rather keep the list to what's proven.

I also swept `integrations/openclaw` for the same pattern — it has no subprocess spawns, so nothing to fix there.

Credit to @tellus1019: the diagnosis in the issue was already correct and complete, including all three scopes. This is that diagnosis implemented, plus a repro that runs in CI.

---

Disclosure: written by an AI agent acting on behalf of the account holder, as part of Palo Alto AI Research Lab. Everything reported above was run, not assumed; the one untested item is flagged as such.

### DIFF
--- a/integrations/hermes/__init__.py
+++ b/integrations/hermes/__init__.py
@@ -290,6 +290,47 @@ for _schema in TOOL_SCHEMAS:
 # ---------------------------------------------------------------------------
 
 
+# Variables that point a child process at the *parent's* Python installation.
+# Hermes ships its own interpreter (3.11 today) and exports these; `bm` is a
+# uv-managed 3.12+ tool with its own environment. A child that inherits them
+# resolves imports against Hermes's site-packages, which surfaces as a native
+# extension ABI failure rather than a missing package.
+#
+# `__PYVENV_LAUNCHER__` is included defensively: the macOS framework launcher
+# exports it to steer a child at a specific interpreter. We have not observed a
+# failure caused by it, unlike the other three.
+_PARENT_PYTHON_ENV_VARS = (
+    "PYTHONPATH",
+    "PYTHONHOME",
+    "VIRTUAL_ENV",
+    "__PYVENV_LAUNCHER__",
+)
+
+
+def _clean_child_env(env: dict[str, str] | None = None) -> dict[str, str]:
+    """
+    Copy of `env` (default: this process's environment) with the parent's
+    Python-interpreter variables removed.
+
+    Every external process this plugin spawns — `bm mcp`, `uv tool install`,
+    `bm project add` — runs under its own interpreter and must resolve its own
+    dependencies. Inheriting Hermes's `PYTHONPATH` makes it import Hermes's
+    site-packages instead, reported as:
+
+        ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
+
+    which is a compiled-extension mismatch between two Python versions, not a
+    missing dependency — so it is invisible to a `pip install` fix.
+
+    Everything else is preserved: the child still needs PATH, HOME, HTTP proxy
+    settings and credentials. An explicitly supplied `env` is sanitized too,
+    since the guarantee is about what the child receives, not about who
+    assembled it.
+    """
+    source = os.environ if env is None else env
+    return {k: v for k, v in source.items() if k not in _PARENT_PYTHON_ENV_VARS}
+
+
 def _bm_binary_path() -> str | None:
     """Find the bm CLI without making network calls. Used by is_available()."""
     candidates = [
@@ -336,6 +377,7 @@ def _install_bm_via_uv(timeout: float = 180.0) -> str | None:
             check=False,
             capture_output=True,
             timeout=timeout,
+            env=_clean_child_env(),
         )
     except Exception as e:
         logger.warning("basic-memory: `uv tool install basic-memory` failed: %s", e)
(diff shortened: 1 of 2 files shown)


phernandez:
thanks for the PR @Palo-Alto-AI-Research-Lab 
We'll take a look at this

tonydzi:
thanks — and one thing cleared on our side since you wrote: the CLA is signed, so `license/cla` is green and the PR is passing everything. nothing is blocked on us.

whenever you get to the review, the two scope calls in the description are one-word answers, and i'll push either the same day:

- strip `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` as well, or leave them alone?
- keep `__PYVENV_LAUNCHER__` (added defensively — unlike the other three i did not reproduce a failure from it), or drop it so the list is only what's proven?

no urgency on either; the fix stands as it is. the diagnosis in #1093 was @tellus1019's and already complete — this is that, implemented, with a repro that runs in CI.

phernandez (approved):
Reviewed at `33c2011e`; no findings.

I also applied the PR cleanly to current `main` (`eedce9bb`) and verified:

- `just package-check-hermes`: 267 passed, 12 skipped
- focused child-environment regression tests: 10 passed
- Ruff check/format, Hermes manifest validation, and `git diff --check`: clean

For the two scope questions:

- Leave `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` alone. They can carry legitimate native-library configuration and are not implicated in this failure.
- Keep `__PYVENV_LAUNCHER__` in the strip list. It is parent-interpreter launcher state and should not steer the independent `bm` child.

The boundary here looks right: isolate the child from Hermes's Python installation while preserving unrelated environment such as `PATH`, `HOME`, proxies, and credentials.

Thanks for the careful diagnosis, complete spawn-site sweep, and strong regression coverage.

phernandez:
@codex review

PR #1193, opened 2026-08-05
fix(core): enforce materialization lock order

### LINKED ISSUE #1187: Prevent Entity-NoteContent lock-order deadlocks during materialization
## Problem

Basic Memory's hosted note materialization and file-index reconciliation can acquire the `entity` and `note_content` rows in opposite orders. Under concurrent work this can deadlock a valid materialization transaction.

Production evidence from Basic Memory Cloud issue basicmachines-co/basic-memory-cloud#1661 shows a materialization job terminated by `asyncpg.exceptions.DeadlockDetectedError` while executing:

```sql
UPDATE entity SET mtime=$1::FLOAT, size=$2::INTEGER WHERE entity.id = $3::INTEGER
```

The trace overlapped a relation-resolution job for the same tenant/project. Logfire did not capture the opposing server query, so this issue does not claim that relation resolution was definitively the other deadlock participant.

The source-level ordering risk is concrete:

- `RepositoryNoteMaterializationPublisher.publish_written_file_state` plans and applies the `note_content` update before updating `Entity.mtime` / `Entity.size`.
- file indexing and derived-state reconciliation can update `Entity` before reconciling `NoteContent`.

Cloud already serializes materializers for one project/entity with a PostgreSQL advisory lock. That does not establish a common row-lock order with indexing or relation work, so duplicating the Core publisher in Cloud would be the wrong fix.

## Outcome

Materialization and indexing use one explicit transaction lock order for the shared `Entity` and `NoteContent` rows, without weakening note-content CAS checks, idempotent redelivery, or stale-write protection.

## Suggested fix

At the Core repository boundary, acquire/lock the target `Entity` row before applying the `NoteContent` publish update (or use an equivalent implementation that proves the same global order). Keep the existing accepted-version and checksum guards intact.

## Acceptance criteria

- [ ] Materialization takes the shared rows in the same `Entity` -> `NoteContent` order used by indexing/reconciliation.
- [ ] A concurrency regression test exercises materialization against an entity-first reconciliation transaction and completes without a PostgreSQL deadlock.
- [ ] The regression fails or detects the lock-order conflict against the pre-fix implementation.
- [ ] Stale database-version/checksum publication still skips or conflicts exactly as before; no last-writer-wins fallback is introduced.
- [ ] Materialization redelivery remains idempotent.
- [ ] The Core change stays tenant-neutral; Cloud continues supplying tenant routing and its materialization advisory lock.

## References

- Cloud observability umbrella: basicmachines-co/basic-memory-cloud#1661
- Cloud directory fanout dependency (separate concern): #1186

phernandez:
## Summary

Fixes the Entity/NoteContent lock-order inversion reported from hosted note
materialization. Version-guarded NoteContent state updates now acquire the
owning Entity row lock before updating NoteContent, matching the existing
Entity-first indexing and reconciliation order.

Fixes #1187.

## Problem

The materialization publisher updated NoteContent first and then flushed file
metadata to Entity in the same transaction. Entity-first reconciliation could
hold the Entity row while waiting to update NoteContent. Under concurrency,
those two transactions formed the cycle:

```text
materialization: NoteContent -> Entity
reconciliation:  Entity -> NoteContent
```

Postgres could select either transaction as a deadlock victim. Retrying at the
Cloud job boundary reduced impact but did not remove the Core lock inversion.

## Solution

- Allow `NoteContentRepository._load_entity_identity()` to request a
  `SELECT ... FOR UPDATE` lock.
- Have the publisher acquire the project-scoped Entity row lock before reading
  NoteContent or planning publication, so a same-version move cannot change the
  destination while publication waits.
- Acquire that Entity lock before every version-guarded NoteContent compare-and-set
  update.
- Keep the existing conditional `db_version` update, stale-job skip, checksum
  propagation, and Entity file-metadata flush unchanged.
- Leave the Cloud advisory materialization lock unchanged; this is the
  tenant-neutral Core row-lock invariant.

The ordering is now consistently:

```text
Entity -> NoteContent
```

## Regression coverage

The new real-Postgres integration test orchestrates two transactions:

1. an Entity-first reconciliation transaction locks Entity;
2. the materialization publisher enters its transaction;
3. the test proves materialization cannot update NoteContent while the opposing
   transaction owns Entity;
4. reconciliation either keeps the path stable or moves Entity and NoteContent,
   then commits;
5. materialization either publishes the stable path or marks the old-path write
   stale and orphaned, without deadlock or incorrect file-state advancement.

Before this fix, the publisher reaches the instrumented NoteContent update while
the Entity lock is held, exposing the inversion. The test also verifies the
stable-path file version, checksum, sync status, Entity mtime, and Entity size.
The moved-path case proves the old write is not recorded against the new path.

## Verification

- `BASIC_MEMORY_ENV=test LOGFIRE_IGNORE_NO_CONFIG=1 uv run pytest -q --no-cov tests/repository/test_note_content_repository.py tests/indexing/test_note_materialization_runner.py tests/cloud/test_note_content_materialization.py tests/runtime/test_note_content_store_materialization.py tests/runtime/test_note_materialization_request_matching.py`
  - 69 passed
- `BASIC_MEMORY_ENV=test BASIC_MEMORY_TEST_POSTGRES=1 LOGFIRE_IGNORE_NO_CONFIG=1 uv run pytest -p pytest_mock --no-cov -q tests/repository/test_note_content_repository.py test-int/test_note_materialization_lock_order.py`
  - 23 passed
- `just fast-check`
  - Ruff fix/check, Ruff format, and ty passed
- `git diff --cached --check`
  - passed before commit

## Risk

Version-guarded NoteContent updates now hold the Entity row lock until their
transaction completes. That adds intentional serialization for updates to the
same note, but it removes the conflicting lock order. Updates for different
entities remain independent.

### DIFF
--- a/src/basic_memory/indexing/note_materialization_runner.py
+++ b/src/basic_memory/indexing/note_materialization_runner.py
@@ -9,6 +9,7 @@ from typing import Protocol, Self
 
 from loguru import logger
 
+from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 
 from basic_memory import db
@@ -426,6 +427,17 @@ class RepositoryNoteMaterializationPublisher:
                 entity_id=request.entity_id,
             )
 
+            # Materialization publishes NoteContent and Entity in one transaction.
+            # Lock Entity before reading the publish state so an Entity-first move
+            # cannot change the destination path while this publisher waits.
+            entity = await session.scalar(
+                select(Entity)
+                .where(
+                    Entity.id == request.entity_id,
+                    Entity.project_id == request.project_id,
+                )
+                .with_for_update()
+            )
             note_content = await session.get(NoteContent, request.entity_id)
             publish_plan = plan_written_note_materialization_publish(
                 request=request,
@@ -436,7 +448,13 @@ class RepositoryNoteMaterializationPublisher:
                     if note_content is not None
                     else None
                 ),
-                current_file_path=note_content.file_path if note_content is not None else None,
+                current_file_path=(
+                    entity.file_path
+                    if entity is not None
+                    else note_content.file_path
+                    if note_content is not None
+                    else None
+                ),
             )
             if note_content is None:
                 return publish_plan.result
@@ -466,7 +484,6 @@ class RepositoryNoteMaterializationPublisher:
                     f"Unhandled note materialization publish action: {publish_plan.action}"
                 )
 
-            entity = await session.get(Entity, request.entity_id)
             if entity is None:
                 return RuntimeNoteMaterializationResult(
                     entity_id=request.entity_id,
--- a/src/basic_memory/repository/note_content_repository.py
+++ b/src/basic_memory/repository/note_content_repository.py
@@ -76,9 +76,18 @@ class NoteContentRepository(Repository[NoteContent]):
 
         return NoteContent(**model_data), set(model_data)
 
-    async def _load_entity_identity(self, session: AsyncSession, entity_id: int) -> Entity:
+    async def _load_entity_identity(
+        self,
+        session: AsyncSession,
+        entity_id: int,
+        *,
+        lock_for_update: bool = False,
+    ) -> Entity:
         """Load the owning entity so duplicated identity fields stay aligned."""
-        result = await session.execute(select(Entity).where(Entity.id == entity_id))
+        query = select(Entity).where(Entity.id == entity_id)
+        if lock_for_update:
+            query = query.with_for_update()
+        result = await session.execute(query)
         entity = result.scalar_one_or_none()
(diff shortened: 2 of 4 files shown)


phernandez:
Addressed the current-head Codex path-race finding in 6cb9f0ad. The publisher now acquires the project-scoped Entity row lock before reading NoteContent or planning publication, so a same-version move cannot change the destination while the publisher waits. The real Postgres regression now covers both a stable-path reconciliation (publish succeeds) and a moved-path reconciliation (old-path write is reported stale/orphaned and does not advance file state). Verification: 44 focused tests plus 2 SQLite skips, 23 real-Postgres repository/lock-order tests, 69 broader stale/CAS/checksum/redelivery tests, and just fast-check passed.

PR #1194, opened 2026-08-05
refactor(core): drop the discarded permalink resolution for regular files

### LINKED ISSUE #1182: Remove the dead resolve_permalink() call for regular files, and document the no-permalink invariant
> **Rescoped 2026-08-04** after review. `permalink IS NULL` for non-markdown files is the **intended design**, not a defect — see [discussion](https://github.com/basicmachines-co/basic-memory/issues/1182#issuecomment-5183996674). What remains is dead code plus an undocumented invariant. The original, wider framing is preserved in this issue's history.

## What remains

**1. A dead `resolve_permalink()` call.** `batch_indexer._upsert_regular_file` (`src/basic_memory/indexing/batch_indexer.py`, ~line 441) computes a permalink and discards it:

```python
if existing is None:
    await self.entity_service.resolve_permalink(file.path, skip_conflict_check=True)  # result discarded
    entity = Entity(
        note_type="file",
        file_path=file.path,
        ...                                    # no permalink=  (correct — see below)
    )
```

`resolve_permalink` is read-only (conflict detection + lookups, returns a string, no persistence side effect), and with `skip_conflict_check=True` it doesn't even emit the conflict warning. So this is a wasted DB round-trip per regular file during indexing, and it misleads readers into thinking the value was meant to be stored. Remove it, or comment why it's there. Check `sync_service` for the same pattern.

**2. The invariant is undocumented.** "Non-markdown entity ⇒ no permalink" holds in practice but is nowhere stated, so consumers keep rediscovering it the hard way — one that assumed a non-null permalink previously produced a hard failure when deleting a non-markdown file (fixed Cloud-side by making the delete/live-update contract permalink-optional). Document it, and keep consumer contracts explicitly permalink-optional.

## Why NULL is correct

The permalink's durable home is the **frontmatter**; the DB row is a projection. A non-markdown file has nowhere to record one, so a DB-only permalink would be derived state that cannot round-trip:

- **Unstable.** Collisions resolve by suffixing (`-1`, `-1-1`). With no file-side record, a rebuild-from-scratch regenerates permalinks in whatever order indexing happens — the same file could return with a different address, silently rotting any `memory://` link or wikilink to it.
- **Breaks rebuild-from-files.** Drop the DB, reindex, get the same graph back; a permalink with no file-side record is the one field that wouldn't survive.
- **Not needed.** File entities already have two stable, file-derived addresses: `file_path` and `external_id`. Permalink serves note semantics (`memory://` addressing, wikilinks, search keying), which a `.png` doesn't have.

## Supporting observation

In one sampled deployment the NULL-permalink set was *exactly* the files without frontmatter — 114 `note_type=file` plus 1 `note_type=canvas` (`.canvas` files are JSON; `file_service` assigns `application/json`) — while every markdown note_type had a permalink. The invariant is already consistent; only the documentation and the dead call are missing.

## Explicitly not wanted

- Storing permalinks for regular files (see "Why NULL is correct").
- Backfilling existing NULL rows.
- Any change to permalink-keyed routing, search, or wikilink resolution.

## Open question

Are non-markdown files excluded from full-text search *because* they lack a permalink, or by design independently of it? Observed but not determined — worth confirming so the search behavior is intentional rather than incidental.

---

Moved from the private Cloud tracker (`basicmachines-co/basic-memory-cloud#1536`), which holds the production evidence and the original failure report. Details omitted here because this repository is public.

jope-bm:
Closes #1182.

## Why
Indexing a regular (non-markdown) file called `resolve_permalink()` and discarded the result. I verified the call is **read-only** — conflict detection plus lookups, returning a string, no persistence side effect — and with `skip_conflict_check=True` it doesn't even emit the conflict warning. So it was a wasted DB round-trip per regular file during indexing, and it misled readers (me included) into thinking the value was meant to be stored.

Per @phernandez's point on the issue, **`permalink IS NULL` for non-markdown files is the intended design, not a defect**: a permalink's durable home is the markdown frontmatter, and a non-markdown file has nowhere to record one. A DB-only value couldn't survive a rebuild from files, and its collision suffix (`-1`, `-1-1`) would depend on indexing order, so the same file could return with a different address and links to it would silently rot. `file_path` and `external_id` are the stable addresses for these entities.

## What changed
- **Removed the discarded `resolve_permalink()` call** in `BatchIndexer._upsert_regular_file`.
- **Documented the invariant where the entity is built** — that's where the discarded call invited the wrong conclusion.
- **Extended the existing non-markdown indexing test** to assert `permalink is None`, locking the invariant on the batch-index path.

## Scope notes (two corrections to the issue as filed)
- **`sync_service` was clean.** The issue suggested auditing it for the same pattern. Every other `resolve_permalink` caller *uses* its return value — `batch_indexer`'s markdown path, `local_moves`, `entity_service.move` — so this was the only site. No sweep needed.
- **The invariant was already partly documented.** `models/knowledge.py` carries "required for markdown files only" on the column, and `test_local_watch_regular_file_parity.py` already asserts `permalink is None` twice for the watch path. The real gap was the batch-index path, which is what this covers. My "undocumented invariant" framing in the issue overstated it.

No behavior change beyond skipping the now-provably-unused lookup: no backfill, no data-shape change, nothing touching permalink-keyed routing, search, or wikilink resolution.

## Testing
- `tests/indexing` + `tests/index`: **602 passed**, 1 deselected (the `@pytest.mark.semantic` env test).
- `ty` and `ruff` clean on the changed files.
- `tests/cli` shows 145 pre-existing failures on this branch **and identically on clean `main`** (145 failed / 531 passed both ways), so they're unrelated to this change.

### DIFF
--- a/src/basic_memory/indexing/batch_indexer.py
+++ b/src/basic_memory/indexing/batch_indexer.py
@@ -438,7 +438,11 @@ class BatchIndexer:
         is_new_entity = existing is None
 
         if existing is None:
-            await self.entity_service.resolve_permalink(file.path, skip_conflict_check=True)
+            # Regular files intentionally carry no permalink. A permalink's durable home is the
+            # markdown frontmatter, and a non-markdown file has nowhere to record one: a DB-only
+            # value could not survive a rebuild from files, and its collision suffix would depend
+            # on indexing order, so links to it would silently rot. file_path and external_id are
+            # the stable addresses for these entities (#1182).
             entity = Entity(
                 note_type="file",
                 file_path=file.path,
--- a/tests/indexing/test_batch_indexer.py
+++ b/tests/indexing/test_batch_indexer.py
@@ -382,6 +382,10 @@ async def test_batch_indexer_indexes_non_markdown_files(
     assert pdf_entity.content_type == "application/pdf"
     assert image_entity is not None
     assert image_entity.content_type == "image/png"
+    # Non-markdown entities carry no permalink: there is no frontmatter to record one in, so a
+    # DB-only value could not survive a rebuild from files (#1182). file_path is their address.
+    assert pdf_entity.permalink is None
+    assert image_entity.permalink is None
 
 
 @pytest.mark.asyncio


phernandez:
Thanks for the careful investigation, and especially for spelling out why non-Markdown files should not receive database-only permalinks.

We independently landed the same fix in #1191 while this PR was open. Current main now:

- removes the discarded resolve_permalink() call from regular-file indexing;
- keeps an inline comment beside entity construction explaining that non-Markdown resources cannot persist a semantic address and use external_id plus file_path instead;
- states explicitly in DOMAIN_MODEL.md that non-Markdown entities always have permalink=None; and
- has the stronger regression: both file entities are asserted to have no permalink, and resolve_permalink() is asserted never to be awaited.

I compared the exact patches. This PR does not leave unique behavior or regression coverage to carry forward, and its useful rationale is preserved in both the implementation comment and the canonical domain documentation. Closing as superseded by #1191. The contribution was correct; the overlap was just timing. Thank you.

PR #1204, opened 2026-08-06
fix(core): batch relation target resolution

### LINKED ISSUE #1201: Batch relation resolution instead of serial per-target lookups
## Summary

Production relation-resolution jobs exceed PGQueuer's five-minute execution budget because Core resolves each unique unresolved target serially. A production-shaped project had 9,913 unique unresolved targets; the database had the expected indexes, so this is an algorithmic round-trip problem rather than a missing-index incident.

The root fix is bulk/batched resolution or durable bounded chunks with resumable progress. Extending the job timeout would only move the failure threshold.

## Production evidence

Observed release: Cloud release `837b8c258ddb9ca78d0d9d897cec9202f7650c28`, pinning Core revision `de4589d24155332002709f5529c9bdd4937861a4`.

Logfire fingerprints:

- #2350 — `cd690865-6a74-4d91-88b3-db2193a2f8c5`
- #2347 — `e328fa25-b51a-4a11-91eb-b1a61917fad5`
- #2341 — `0c586488-853b-4203-96ab-1eabc059208e`

Representative trace: [five-minute terminal relation-resolution job](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019fd458c146e8736cbe18a389a6c6bd%27&since=2026-08-05T23%3A53%3A34.278108%2B00%3A00&until=2026-08-06T23%3A53%3A34.286111%2B00%3A00).

The representative job terminated at five minutes. Read-only production database inspection found:

- 1,483 entities
- 46,988 relation rows
- 17,062 unresolved relation rows
- 9,913 unique unresolved target strings
- the expected B-tree/trigram entity indexes and unresolved-relation index

Tenant and project identifiers are intentionally omitted.

## Causal chain

1. `RepositoryRelationResolutionRuntime.resolve_relations` caches the set of unique target strings.
2. It still awaits `LinkResolver.resolve_link` once per target, serially.
3. Each resolution can call `_resolve_in_project` and `EntityRepository.get_by_permalink`, producing another database round trip.
4. Roughly 10,000 sequential resolutions cannot complete inside the worker budget.
5. The job is terminated and the graph retains unresolved work.

The existing indexes support the individual lookups; they do not compensate for the serial N-target control flow.

## Required outcome

Resolve target strings in bulk/batches while preserving the existing link contract, or process them as durable bounded chunks with explicit resumable progress. The implementation may combine both: bulk-fetch candidate entities per chunk, apply normal ambiguity/qualification rules in memory, persist resolved targets, and checkpoint progress.

Core owns these semantics because the same resolver serves local and hosted runtimes.

## Non-solutions

- Increasing the PGQueuer timeout
- Adding another retry around the same five-minute algorithm
- Dropping unresolved relations
- Replacing precise link semantics with a fuzzy first-match shortcut

## Acceptance criteria

- [ ] A production-shaped PostgreSQL fixture with about 10,000 unique unresolved targets completes within the normal worker budget.
- [ ] SQL round trips are bounded by batch/chunk count rather than unique-target count.
- [ ] Current-project, permalink/title, qualified cross-project, source-path, and ambiguity behavior remain unchanged.
- [ ] The operation is idempotent.
- [ ] If chunked work is interrupted, a later attempt resumes without repeating the full scan or losing unresolved rows.
- [ ] Result accounting reports accurate resolved and unresolved counts.
- [ ] The regression demonstrates that the pre-fix serial implementation exceeds a defined query/latency budget.
- [ ] No generic timeout increase is required for the fixture to pass.

phernandez:
## Why

Production relation-resolution jobs can contain roughly 10,000 unique unresolved target strings. Resolving each target through the single-link `LinkResolver` issued several indexed SQL lookups, so query count grew with target count and exhausted the five-minute worker budget.

Closes #1201.

## What Changed

- Kept `LinkResolver` focused on its valid single-link use cases.
- Added a separate `BulkLinkResolver` for relation repair. It loads one immutable identity snapshot for the current and referenced projects, then resolves the whole target set without per-target SQL.
- Preserved strict exact-match behavior for external IDs, permalinks, titles, normalized POSIX file paths, explicit `project::target` references, and `project/path` fallback.
- Added an explicit `RelationTargetBatchResolver` capability at the relation-runtime boundary so composition cannot accidentally substitute the regular resolver.
- Replaced arbitrary 250-row slicing with typed write-batch planning that never splits a `(from_id, relation_type)` uniqueness domain. A domain may exceed the preferred size when correctness requires it.
- Made relation writes compare-and-set operations over the original unresolved identity, so a deleted relation ID cannot redirect stale resolution work onto a replacement edge.
- Carried each resolved target's stable external ID into the write contract. Target rows are revalidated and protected through mutation, so a reused numeric entity ID cannot redirect an edge to an unrelated replacement.
- Kept each collision domain in one transaction while bounding duplicate deletion, temporary staging, and canonical assignment SQL statements; oversized domains no longer exceed SQLite's expression-depth limit.
- Kept each safe write batch independently committed so interrupted workers resume from the remaining unresolved rows.

## Design

`BulkLinkResolver` models the bulk operation as three explicit phases:

1. parse and normalize unique relation targets once;
2. load the project catalog and required lightweight entity identities once;
3. resolve every target from the immutable snapshot with no additional I/O.

The regular resolver retains source-relative and fuzzy single-link behavior. Bulk relation repair remains strict: duplicate exact titles stay unresolved, self-relations are skipped by the runtime, and an ambiguous current-project target cannot silently fall through to another project.

Relation writes are grouped by the collision domain shared by both database uniqueness constraints. This keeps alias/name swaps in one transaction; fixed slices could otherwise make an earlier batch treat a later row as an occupied target and delete a valid edge.

Each planned write carries the original source, unresolved target name, and relation type. Duplicate deletion and update staging require that complete identity to still match and use `RETURNING` to select the rows that may proceed. A source-note reindex can therefore make old work stale, but cannot make it mutate a different relation that reused the same database ID.

The target side uses the entity's stable external ID alongside its numeric database ID. PostgreSQL locks matching target rows in bounded reads for the transaction; SQLite rechecks the external ID in the first compare-and-set mutation before acquiring its write lock. A deleted target or reused numeric ID therefore leaves the original relation unresolved instead of connecting it to the replacement.

An oversized collision domain remains one transaction, but its SQL mutations execute in bounded statements. All accepted rows are first moved to unique temporary names across those statements; only after the entire domain is staged are canonical targets assigned. This preserves name swaps across statement boundaries without building unbounded `OR` or `CASE` expressions.

## Testing

- Focused final surface: 53 passed, including bulk/single-link parity, path normalization, cross-project routing, ambiguity, interruption/resumption, retryable search refresh, and canonical-name swaps.
- PostgreSQL production shape: 9 passed, including the 10,000-target bounded-query regression and canonical-name swap coverage.
- Final repository/runtime regression surface: 44 passed after formatting, including separate same-ID source-relation and target-entity replacement protection plus a 1,100-row single-source canonical-name rotation across SQL statement boundaries.
- Focused PostgreSQL source/target identity, canonical-swap, 1,100-row oversized-domain, and 10,000-target tests passed.
- Impacted testmon surface: 78 relevant tests passed. The only selected failure was the unrelated live OpenAI semantic benchmark, which returned `429 credit_balance_exhausted`.
- Broad local run: 5,110 passed, 36 skipped, with the same single live-provider quota failure.
- `just fast-check` passed Ruff formatting/lint and full `ty` checking.
- `just doctor` passed the end-to-end file/database/search loop.
- `git diff --check` passed.

## Risk

Bulk relation repair holds lightweight identity rows for each referenced project for the duration of one pass. The snapshot deliberately avoids eager-loading observations and relations; SQL round-trips scale with referenced projects and safe write batches, not unique target count.

This PR remains unmerged for current-head review.


src/basic_memory/services/link_resolver.py
```diff
@@ -274,6 +314,214 @@ async def resolve_link(
                 load_relations=load_relations,
             )
 
+    async def resolve_relation_targets(
```
phernandez: this function is too big

src/basic_memory/services/link_resolver.py
```diff
@@ -274,6 +314,214 @@ async def resolve_link(
...
+            )
+        )
+        projects_by_id = {project.id: project for project in projects}
+        current_project = projects_by_id.get(current_project_id)
+        if current_project is None:
+            raise RuntimeError(f"Current project {current_project_id} does not exist")
+
+        indexes_by_project_id: dict[int, _StrictLinkResolutionIndex] = {}
+
+        async def get_project_index(project: Project) -> _StrictLinkResolutionIndex:
```
phernandez: inner function smell

src/basic_memory/services/link_resolver.py
```diff
@@ -274,6 +314,214 @@ async def resolve_link(
...
+
+            resolved_targets[link_text] = await self._resolve_relation_target_in_project(
+                project=project,
+                link_text=remainder,
+                get_project_index=get_project_index,
+            )
+
+        return resolved_targets
+
+    async def _resolve_relation_target_in_project(
```
phernandez: too many _private functions

src/basic_memory/services/link_resolver.py
```diff
@@ -274,6 +314,214 @@ async def resolve_link(
...
+                return found_path_with_md
+
+        if ambiguous_title:
+            raise AmbiguousIdentifierError(
+                clean_text,
+                [(entity.permalink, entity.file_path) for entity in title_matches],
+            )
+
+        return None
+
```
phernandez: This needs to be resdigned to properly model processing in batches.

src/basic_memory/indexing/relation_resolution.py
```diff
@@ -227,60 +225,72 @@ async def resolve_relations(
...
-                        relation_type=relation.relation_type,
-                    )
+            )
+
+        # Each chunk commits independently. If a worker is interrupted, the next attempt sees
+        # only the still-unresolved rows instead of repeating already completed target writes.
+        write_chunks = [
+            writes[start : start + RELATION_RESOLUTION_WRITE_CHUNK_SIZE]
+            for start in range(0, len(writes), RELATION_RESOLUTION_WRITE_CHUNK_SIZE)
+        ] or [[]]
```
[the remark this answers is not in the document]
phernandez: Fixed in 7933ea42. Write planning now groups the full (from_id, relation_type) collision domain into one ResolvedRelationWriteBatch and never splits a domain merely to hit the 250-write target. Added both a planner boundary regression and a repository canonical-name-swap regression; the final SQLite and PostgreSQL surfaces pass.

src/basic_memory/services/link_resolver.py
phernandez: Removed in 7933ea42. There is no inner resolution closure or closure-owned cache now: snapshot loading is a top-level operation, and resolution runs through typed immutable snapshot values.

src/basic_memory/services/link_resolver.py
```diff
@@ -274,6 +314,214 @@ async def resolve_link(
...
+            if entity is None:
+                continue
+            if ambiguous_title and candidate_permalink != exact_identifier:
+                break
+            return entity
+
+        if len(title_matches) == 1:
+            return title_matches[0]
+
+        found_path = index.by_file_path.get(clean_text)
```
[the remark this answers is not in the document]
phernandez: Fixed in 7933ea42. The dedicated ProjectEntityIdentityIndex normalizes path identifiers with Path(identifier).as_posix() before lookup and preserves the .md fallback. Regression coverage includes ./assets//image.png plus a custom-permalink Markdown path, and passes on SQLite and PostgreSQL.

src/basic_memory/services/link_resolver.py
phernandez: Redesigned in 7933ea42 instead of shortening the oversized method in place. LinkResolver is restored to the focused single-link implementation from main. Bulk work now lives in BulkLinkResolver with explicit parse, snapshot-load, and in-memory resolve phases.

src/basic_memory/services/link_resolver.py
phernandez: Redesigned in 7933ea42. Bulk target lookup is now its own BulkLinkResolver flavor behind a RelationTargetBatchResolver capability, while write batching is modeled separately as collision-safe ResolvedRelationWriteBatch plans. The preferred batch size can be exceeded to keep a uniqueness domain atomic; interruption/resumption and 10,000-target PostgreSQL coverage pass.

src/basic_memory/services/link_resolver.py
phernandez: Addressed in 7933ea42. The bulk-only private-helper sprawl was removed from LinkResolver. The new module uses named domain values for target references, project identity lookup, project entity identity lookup, and the immutable resolution snapshot, each with one focused responsibility.

phernandez:
@codex review

phernandez (commented):
Addressed in 7933ea42 with separate single-link and bulk resolver designs, collision-safe batch planning, and regression coverage.

src/basic_memory/indexing/relation_resolution.py
```diff
@@ -227,60 +265,66 @@ async def resolve_relations(
...
-                        relation_type=relation.relation_type,
-                    )
+            )
+
+        duplicate_relation_ids: list[int] = []
+        for write_batch in plan_resolved_relation_write_batches(writes):
+            async with db.scoped_session(self.session_maker) as session:
+                write_result = await self.relation_repository.apply_resolved_targets(
+                    session,
+                    write_batch.writes,
```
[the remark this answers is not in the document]
phernandez: Fixed in f08d15f2. Each ResolvedRelationWrite now carries the original unresolved identity. Repository duplicate deletion and update staging use compare-and-set predicates over relation ID, source ID, unresolved state, original target name, and relation type; only rows returned by those mutations proceed. Added an explicit same-ID replacement regression. Final SQLite repository/runtime suite: 42 passed. Focused PostgreSQL stale-write, canonical-swap, and 10,000-target tests: 3 passed.

phernandez:
@codex review

src/basic_memory/indexing/relation_resolution.py
```diff
@@ -20,11 +20,51 @@
...
+        collision_domain = (write.from_id, write.relation_type)
+        writes_by_collision_domain.setdefault(collision_domain, []).append(write)
+
+    batches: list[ResolvedRelationWriteBatch] = []
+    pending_batch: list[ResolvedRelationWrite] = []
+    for collision_domain_writes in writes_by_collision_domain.values():
+        if pending_batch and len(pending_batch) + len(collision_domain_writes) > target_size:
+            batches.append(ResolvedRelationWriteBatch(tuple(pending_batch)))
+            pending_batch = []
+        pending_batch.extend(collision_domain_writes)
```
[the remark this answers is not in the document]
phernandez: Fixed in 0fb676df. The collision domain still executes in one transaction, but duplicate deletion, compare-and-set staging, and canonical assignment now use bounded 250-row SQL statements. Crucially, every accepted row is staged to a unique temporary name before any canonical assignment, so swaps remain safe across statement boundaries. Added a 1,100-row single-source regression that rotates canonical names across those boundaries; it passes on SQLite and PostgreSQL. The full focused repository/runtime suite is 43 passed, and just fast-check is green.

phernandez:
@codex review

src/basic_memory/indexing/relation_resolution.py
```diff
@@ -227,60 +265,80 @@ async def resolve_relations(
...
+                f"resolved_id={resolved_entity.id} "
+                f"resolved_title={resolved_entity.title}",
+            )
+            writes.append(
+                ResolvedRelationWrite(
+                    relation_id=relation.id,
+                    from_id=relation.from_id,
+                    original_target_name=relation.to_name,
+                    target_id=resolved_entity.id,
+                    target_name=resolved_entity.title,
```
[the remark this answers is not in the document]
phernandez: Fixed in 4e3da699. ResolvedRelationWrite now carries the target Entity.external_id alongside its numeric ID. apply_resolved_targets revalidates that stable identity; PostgreSQL locks matching targets for the transaction, while the compare-and-set mutation also checks target ID plus external ID for SQLite before its write lock. Added an explicit same-ID target replacement regression that leaves the original edge unresolved. Final focused suite: 44 passed; PostgreSQL identity/swap/1,100-row cases: 4 passed; the 10,000-target query-budget test still passes.

phernandez:
@codex review

src/basic_memory/indexing/relation_resolution.py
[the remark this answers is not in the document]
phernandez: Follow-up in ff5511cb: target identity locking reads are also bounded to 250 IDs inside the same transaction, so the stable-identity guard cannot reintroduce a backend parameter limit for an oversized collision domain. The 1,100-row SQLite/PostgreSQL regression and 10,000-target PostgreSQL query-budget test remain green.

phernandez:
@codex review

PR #1215, opened 2026-08-09
fix(core): survive a stale Windows log deleted mid-cleanup (#1211)

### LINKED ISSUE #1211: Windows: MCP server can die at startup on a log-cleanup race in _cleanup_windows_log_files
### Summary

On Windows, `_cleanup_windows_log_files` (`src/basic_memory/utils.py:490`) can raise `FileNotFoundError` and kill the process during logging setup, before anything else runs. When the process is the MCP server, the client sees the stdio connection close a couple of seconds after launch, and the session ends up with **zero** basic-memory tools registered — silently, because nothing in the client surfaces the traceback.

Still present on `main` @ `816accaa` (checked 2026-08-09), identical to the build I hit it on.

### The race

```python
# src/basic_memory/utils.py:492-504
stale_logs = [
    path
    for path in log_dir.glob("basic-memory-*.log*")
    if path.is_file() and path.name != current_log_name
]
...
stale_logs.sort(key=lambda path: path.stat().st_mtime, reverse=True)
for stale_log in stale_logs[WINDOWS_LOG_FILE_RETENTION - 1 :]:
    try:
        stale_log.unlink()
    except OSError:
        logger.debug(...)
```

The listing and the `stat()` in the sort key are separate steps. Windows log files are per-PID, so **every** basic-memory launch prunes the same shared directory. If a second process unlinks one of those files between the `glob` and the `stat`, the sort key raises and the process dies.

The `try/except OSError` at line 506 guards the `unlink` but not the `stat` — so the deletion is protected against exactly this and the read that precedes it is not.

Multiple concurrent launches are the normal case for the Claude Code plugin: several editor sessions, plus `basic-memory hook session-start` firing per session, all starting at once.

### Traceback

```
File "src/basic_memory/cli/app.py", line 85, in app_callback
    init_cli_logging()
File "src/basic_memory/config.py", line 280, in init_cli_logging
    _config_logging.initialize_file_logging(...)
File "src/basic_memory/config_logging.py", line 54, in initialize_file_logging
    setup_logging(log_level=log_level, log_to_file=True)
File "src/basic_memory/utils.py", line 451, in setup_logging
    _cleanup_windows_log_files(log_path.parent, log_path.name)
File "src/basic_memory/utils.py", line 504, in <lambda>
    stale_logs.sort(key=lambda path: path.stat().st_mtime, reverse=True)
File "pathlib/__init__.py", line 654, in stat
    return os.stat(self, follow_symlinks=follow_symlinks)
FileNotFoundError: [WinError 2] The system cannot find the file specified:
    '<data-dir>\\basic-memory-18592.log'
```

### What the client sees

From the MCP client's log:

```
Starting connection with timeout of 30000ms
Server stderr: <the traceback above>
Connection failed after 2599ms (-32000): MCP error -32000: Connection closed
```

Two things make this hard to notice:

1. **No error reaches the model or the user.** The session simply has no `mcp__basic-memory__*` tools, while the bundled skills, slash commands and output style all go on instructing against tools that do not exist.
2. **A health check cannot reproduce it.** `claude mcp list` launches a *fresh* server, which succeeds, so it reports the server as connected. Diagnosing it requires reading the client's per-session MCP log.

### Frequency

Rare but real: 1 failure in 111 connection attempts across my session logs, with 8 stale `basic-memory-<pid>.log` files in the data directory at the time.

### Suggested fix

Make the sort key tolerate a file that has already gone, e.g.

```python
def _mtime(path: Path) -> float:
    try:
        return path.stat().st_mtime
    except OSError:
        return -1.0  # vanished under us — sort last, unlink() below is already guarded

stale_logs.sort(key=_mtime, reverse=True)
```

Since the cleanup is best-effort housekeeping, wrapping the whole call in `setup_logging` so it can never take down the process would also be reasonable — a failure to prune old logs should not stop the server from starting.

### Environment

- basic-memory installed from git `main`; MCP server reports FastMCP 3.4.5
- Windows 11, Python 3.14.3
- Client: Claude Code, stdio transport

overgoy:
`_cleanup_windows_log_files` can kill the process during logging setup.

Windows log files are per-PID, so every launch prunes the same shared directory, and concurrent launches are the normal case with the Claude Code plugin: several editor sessions plus a session-start hook each start a process. The `unlink` is guarded against another launch removing a file first; the `stat` in the sort key that precedes it is not, so `FileNotFoundError` escapes through `setup_logging`. When that process is the MCP server, the client sees the stdio connection close a couple of seconds after launch and the session silently has no basic-memory tools, while the bundled skills and slash commands go on instructing against tools that are not there.

The sort key now treats a file that has already gone as oldest, which puts it in the delete slice where `unlink` already handles the same race. The regression test stands in for the racing launch by deleting the file during the `stat` it is about to be sorted by.

Fixes #1211.

**Tested.** `tests/utils/` passes (166 passed). The full `tests/` run is 2407 passed, 32 skipped, with one failure in `tests/cli/test_ci_commands.py::test_setup_writes_workflow_config_and_prompt`; that one fails identically on a clean checkout of the same commit in my environment (a missing external binary in a slim container) and is unrelated to this change.

Disclosure: prepared with an AI coding assistant, per the collaborative workflow described in CONTRIBUTING.

### DIFF
--- a/src/basic_memory/utils.py
+++ b/src/basic_memory/utils.py
@@ -501,7 +501,15 @@ def _cleanup_windows_log_files(log_dir: Path, current_log_name: str) -> None:
     # Trigger: per-process log filenames avoid Windows rename contention but fragment retention.
     # Why: loguru retention applies per sink, not across the whole basic-memory log directory.
     # Outcome: keep only the newest stale PID logs so repeated CLI/server launches stay bounded.
-    stale_logs.sort(key=lambda path: path.stat().st_mtime, reverse=True)
+    def _mtime(path: Path) -> float:
+        try:
+            return path.stat().st_mtime
+        except OSError:
+            # Another launch pruned it between the glob above and this read. Sorting it last means it lands in
+            # the delete slice, where unlink() is already guarded against the same race.
+            return -1.0
+
+    stale_logs.sort(key=_mtime, reverse=True)
     for stale_log in stale_logs[WINDOWS_LOG_FILE_RETENTION - 1 :]:
         try:
             stale_log.unlink()
--- a/tests/utils/test_setup_logging.py
+++ b/tests/utils/test_setup_logging.py
@@ -84,6 +84,50 @@ def test_setup_logging_trims_stale_windows_pid_logs(monkeypatch, tmp_path) -> No
     ]
 
 
+def test_setup_logging_survives_a_stale_log_deleted_mid_cleanup(monkeypatch, tmp_path) -> None:
+    """Regression guard for #1211: a stale log that disappears mid-cleanup must not kill the process.
+
+    Windows log files are per-PID, so every launch prunes the same shared directory and concurrent launches
+    are the normal case for the Claude Code plugin. The ``unlink`` below is guarded against exactly that race;
+    the ``stat`` in the sort key that precedes it was not, so ``FileNotFoundError`` escaped through
+    ``setup_logging`` and ended the process during logging setup. When that process was the MCP server, the
+    client saw the stdio connection close and the session silently had no basic-memory tools at all.
+    """
+    log_dir = tmp_path / ".basic-memory"
+    log_dir.mkdir()
+
+    for index in range(6):
+        log_path = log_dir / f"basic-memory-{1000 + index}.log"
+        log_path.write_text("old log", encoding="utf-8")
+        mtime = 1_000 + index
+        os.utime(log_path, (mtime, mtime))
+
+    vanishing = log_dir / "basic-memory-1003.log"
+    real_stat = Path.stat
+
+    def stat_with_a_racing_deletion(self: Path, *args: Any, **kwargs: Any) -> Any:
+        # Stand in for another launch pruning the same shared directory between the glob and the stat.
+        if self == vanishing and vanishing.exists():
+            vanishing.unlink()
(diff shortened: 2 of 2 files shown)


overgoy:
Thanks — the recursion report is correct. `Path.exists()` calls `stat()`, which is the method the test patches, so the existence check re-entered itself; it passed locally on 3.14 where that is not the case. Replaced with a one-shot flag that touches the filesystem not at all during the patched `stat`, and verified on 3.12 and 3.13 (166 passed in `tests/utils/`).

Also retitled from `fix(utils)` to `fix(core)` for the scope check.

phernandez:
Thank you, @overgoy — your production fix is now merged through #1218 as the distinct contributor-authored commit b712db8de11db8bf255d574cebec1617f009405f. Your authorship, DCO sign-off, and the original a00200e commit provenance were preserved.

I also saw the follow-up test commit 6d8a45c you pushed while the replacement PR was already in CI. We had independently made the same one-shot-flag correction for the Path.exists() recursion; the merged fixture additionally keeps Path.is_file() on the real filesystem so the simulated deletion is guaranteed to occur during the sort-key stat and directly exercises the production guard.

With the fix and regression coverage merged, I am closing this PR as superseded. Thanks again for finding and fixing the Windows logging startup race — and for responding quickly to the review feedback.

PR #1227, opened 2026-08-11
fix(core): remove entity read locks from materialization publish

### LINKED ISSUE #1224: Remove SELECT FOR UPDATE from note materialization publish; rely on the db_version CAS with guarded writes
## Problem

`publish_written_file_state` in `src/basic_memory/indexing/note_materialization_runner.py` acquires `FOR UPDATE` row locks on NoteContent and Entity and holds them across the whole plan-compute-write span. In Basic Memory Cloud this anchors real Postgres deadlocks: a three-transaction wait cycle on the entity row lock, recurring on the pgq worker (basicmachines-co/basic-memory-cloud#1727, Logfire #2384, 4 occurrences 2026-08-07 → 2026-08-10).

## Why the locks are removable

The path already has the mechanisms that make pessimistic locking redundant:

- Every NoteContent write goes through `apply_note_content_update_plan(..., expected_db_version=...)` — a compare-and-swap with an explicit stale-skip path. The code comments state this exists precisely because concurrent publishers are expected.
- `with_for_update` is a documented no-op on SQLite (`note_content_repository.py`), so core semantics must already be correct without row locks; the CAS is load-bearing by design.
- Hosted runtimes can serialize same-entity materialization externally via `NoteMaterializationSessionLock` (Cloud passes a `pg_advisory_xact_lock(project_id, entity_id)` taken first in the transaction, which is deadlock-free by construction). The default Noop lock is exactly the concurrent case the CAS guards.

## Fix

In `publish_written_file_state`:

- Read NoteContent and Entity plainly — drop both `with_for_update()` calls.
- NoteContent writes: unchanged; the CAS row lock is taken only at UPDATE time.
- Entity write: replace the blind ORM attribute writes (`entity.mtime`, `entity.size`) with one guarded single-statement UPDATE (`WHERE id AND project_id`; rowcount 0 → the existing "entity disappeared" result). It already runs only after a successful CAS in the same transaction, so the canonical NoteContent-before-Entity order is preserved with a millisecond lock window.
- Same treatment for `NoteFileVacateRepository.clear_vacate_path`, which takes `with_for_update` on Entity inside this transaction.

Out of scope: the canonical lock-order protocol in `relation_repository.py` (`current_relation_generation_statement` and friends) stays as is — this change only stops the materialization publisher from holding SELECT-time locks that other protocols (including FK KEY SHARE from relation writes) can deadlock against.

## Acceptance criteria

- [ ] `publish_written_file_state` and the vacate-marker path hold no SELECT-time row locks; all writes are guarded single-statement updates with rowcount checks.
- [ ] Synthetic regression test: concurrent materialization publish vs newer accepted write — the newer `db_version`/`file_version` is never reverted, and the stale publisher reports `stale` rather than erroring.
- [ ] Existing SQLite behavior unchanged (locks were no-ops there already).
- [ ] Cloud symptom tracked in basicmachines-co/basic-memory-cloud#1727 resolves (no Logfire #2384 recurrence).

phernandez:
Closes #1224. Cloud symptom: basicmachines-co/basic-memory-cloud#1727 (Logfire #2384).

## What changed

`publish_written_file_state` no longer takes any SELECT-time row locks. The existing
`expected_db_version` CAS is the only write guard, and the transaction's first row lock is
the CAS UPDATE itself — preserving the canonical NoteContent-first order, so the publisher
can no longer anchor a lock cycle.

- Dropped both `with_for_update()` calls on the NoteContent and Entity planning reads.
- **Also removed the Entity lock inside the CAS path** (`_load_entity_identity(lock_for_update=True)`
  in `update_state_fields`). This lock isn't in the issue text, but dropping only the
  publisher's read locks would have left the publisher acquiring Entity → NoteContent while
  accepted mutations acquire NoteContent → Entity — recreating the deadlock with inverted
  order. The dead `lock_for_update` parameter is deleted.
- Entity `mtime`/`size` writes are now one guarded single-statement UPDATE via the existing
  `Repository.update_fields` (`WHERE id AND project_id`, rowcount-checked). rowcount 0 maps
  to the existing "entity disappeared" missing result and skips the vacate-marker clear. No
  staleness guard is needed: a successful CAS holds the note_content row lock until commit,
  and every producer of newer entity file metadata crosses that row first.
- **CAS-loss orphan re-check** (slightly beyond the issue text — flagging for review): with
  the old locked reads, a concurrent move was visible at plan time and the written file at
  the vacated path got orphan cleanup. With plain reads that move surfaces as a CAS loss, so
  the loss branch now re-reads Entity (`populate_existing=True` to bypass the identity map);
  if the entity is gone or its path moved, the stale result carries
  `written_file_orphaned=True`, restoring exact cleanup parity.

## Issue-text corrections

- `clear_vacate_path` is untouched: it is already a plain project-scoped DELETE with no
  Entity lock. The Entity `with_for_update` in that file is `lock_recoverable_vacate`, which
  belongs to move-recovery and never runs in the publish transaction.

## Verification

- `tests/indexing/test_note_materialization_runner.py` + repo/reconciler suites: 75 passed
  (4 new unit tests: entity missing pre-CAS, entity missing post-CAS, CAS loss same-path not
  orphaned, CAS loss orphans vacated path).
- `test-int/test_note_materialization_lock_order.py` on real Postgres (testcontainers):
  3 passed — the FOR UPDATE deadlock probe is kept as a regression tripwire, the moved-path
  case now models a real move (db_version advances) and asserts the newer path is never
  reverted, and the new `test_publish_cas_loss_never_reverts_newer_accepted_write` pins the
  acceptance criterion (stale publisher reports `stale`, newer write survives untouched).
- `just typecheck`, `just lint`: clean.

Note for release: verify basic-memory-cloud's pgq worker doesn't rely on the publisher
holding Entity row locks (it injects the advisory `session_lock`, which is unchanged).

### DIFF
--- a/src/basic_memory/indexing/note_materialization_runner.py
+++ b/src/basic_memory/indexing/note_materialization_runner.py
@@ -57,6 +57,7 @@ from basic_memory.runtime.storage import (
     RuntimeFilePath,
 )
 from basic_memory.models import Entity, NoteContent
+from basic_memory.repository.entity_repository import EntityRepository
 from basic_memory.repository.note_file_vacate_repository import NoteFileVacateRepository
 
 type NoteMaterializationPreflightOutcome = (
@@ -427,25 +428,21 @@ class RepositoryNoteMaterializationPublisher:
                 entity_id=request.entity_id,
             )
 
-            # Materialization publishes NoteContent and Entity in one transaction.
-            # The canonical order lives in current_relation_generation_statement:
-            # claim NoteContent before any Entity lock so publication, accepted
-            # mutation, deletion, and materialization cannot form a lock cycle.
+            # These are plain snapshots for planning. This transaction's first row
+            # lock is the NoteContent CAS UPDATE, preserving the canonical order in
+            # current_relation_generation_statement. With no read locks held, the
+            # publisher cannot anchor the lock cycle reported in #1224.
             note_content = await session.scalar(
-                select(NoteContent)
-                .where(
+                select(NoteContent).where(
                     NoteContent.entity_id == request.entity_id,
                     NoteContent.project_id == request.project_id,
                 )
-                .with_for_update()
             )
             entity = await session.scalar(
-                select(Entity)
-                .where(
+                select(Entity).where(
                     Entity.id == request.entity_id,
                     Entity.project_id == request.project_id,
                 )
-                .with_for_update()
             )
             publish_plan = plan_written_note_materialization_publish(
                 request=request,
@@ -509,9 +506,20 @@ class RepositoryNoteMaterializationPublisher:
                 expected_db_version=expected_db_version,
             )
             if not applied:
-                # A newer accepted write (and its own materialization) superseded
-                # this one between our read and write; skip the stale file_version
-                # publish and the entity metadata update rather than reverting them.
+                # Trigger: a move can commit after the plain planning reads but
+                # before the CAS observes the newer NoteContent row.
+                # Why: that observation guarantees this later plain read sees the
+                # move's committed Entity path.
+                # Outcome: clean the just-written vacated path while keeping a
+                # same-path superseded write in place for its newer materialization.
+                current_entity = await session.scalar(
+                    select(Entity)
+                    .where(
+                        Entity.id == request.entity_id,
+                        Entity.project_id == request.project_id,
+                    )
+                    .execution_options(populate_existing=True)
+                )
                 return RuntimeNoteMaterializationResult(
                     entity_id=request.entity_id,
                     status=RuntimeNoteMaterializationStatus.stale,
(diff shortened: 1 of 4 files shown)


phernandez:
Reviewed both findings. Declining both — rationale for the record:

**P1 (stale identity copy under a concurrent project-index move):** the premise is correct —
`project_index_maintenance.py` repoints `Entity.file_path` and `NoteContent.file_path` in
bulk without advancing `db_version`, so an overlapping move can make the unlocked identity
read stale while the CAS still passes, transiently writing the old path back into
`note_content.file_path`. But the pre-PR code did not prevent this race — it deadlocked on
it: the mover claims NoteContent first (`lock_note_content_before_entity_mutation`) while
the old publisher held the Entity row lock, which is exactly the #1224 cycle this PR
removes. Post-PR the overlap resolves without deadlock, and the divergence is a transient
denormalized-copy drift: materialization planning already prefers `entity.file_path`, and
the next accepted write or index pass refreshes the copy. We explicitly prefer that
eventual consistency over reintroducing SELECT-time locks. If telemetry ever shows this
mattering, the lock-free fix is copying the identity fields via a correlated scalar
subquery inside the single CAS UPDATE — noted for the future, not needed now.

**P2 (re-check CAS loss in the already-stale branch):** that branch's written file is
already superseded at plan time; if its guarded stale-write additionally loses the race,
the vacated-path marker and the next index pass reconcile the leftover file. Derived files
are eventually consistent by design here — a second re-read on an already-rare branch is
complexity without a failure mode that doesn't self-heal.

phernandez:
Declining this finding as well — same doctrine, with the reasoning on record:

The race is real but the window is two adjacent SELECTs wide: a deletion must commit in the
microseconds between the NoteContent read and the Entity read while a materialization for
that same entity is mid-flight. And the outcome is not corruption — the written file is the
materialization of a real accepted generation. Under this system's files-are-source-of-truth
model, a file on disk *is* a note; the next scan indexing it is the system converging on
"the accepted write happened," which is a coherent last-writer-wins resolution of a
genuinely ambiguous race (the file was deleted while a legitimate write for the same note
was in flight). Marking the result orphaned would resolve the same ambiguity in the other
direction — deleting an accepted write's output on the strength of a row-visibility race.

We are deliberately not adding per-race special cases to this path. Derived state here is
eventually consistent; every projection converges via the next write or index pass. If
telemetry ever shows deleted notes resurrecting through this window in practice, the fix is
one line (`written_file_orphaned=True` on the missing-entity results, cleanup stays
checksum-guarded) — recorded here for that day.

Issue #1226, opened 2026-08-11
reindex --search cannot repair DB-only search index divergence


phernandez:
## Problem

`basic-memory reindex --search` cannot repair database-only divergence: it selects files to
index purely from file-content checksum changes, so when the `observation`/`relation`/
`search_index` tables disagree with an unchanged markdown file, the incremental pass selects
zero files and reports success while search keeps serving stale rows.

Field report on #1214 (SQLite, concurrent MCP clients): after trimming duplicate
`observation` rows, `search_index` still held 3886 rows of `type='observation'` against 2146
observation rows. `reindex --search` reported `Indexing files... 0/0 files` and changed
nothing; until `reindex --full --search` ran, search answered with content no longer in the
store.

## Root cause

`select_project_index_target_files` (`src/basic_memory/indexing/project_index_coordinator.py:229-242`)
builds its target set from `change_report.new_files | modified_files`, which
`change_detector.py` derives purely from file checksums (`load_indexed_file_checksums`).
DB-only divergence leaves every file in `unchanged_files`, so nothing reaches
`search_service.index_entity_data` — the only code path that deletes and rewrites an
entity's observation search rows. `_purge_stale_search_rows` only handles rows for deleted
entities, and only runs in the `--embeddings` path.

## Proposed direction

Add a DB↔FTS reconciliation sweep to the search reindex path (extend
`_purge_stale_search_rows` or a sibling): purge `search_index` rows whose `id` no longer
exists in the owning table (`type='observation'` → `observation.id`, `type='relation'` →
`relation.id`), and optionally detect entities whose row counts disagree and re-enqueue just
those entities, so `--search` without `--full` converges without a full rebuild.

The one-time damage from the observation duplication bug is handled by the #1214 migration's
FTS orphan purge; this issue is about making incremental `reindex --search` able to repair
such divergence in general.

## Acceptance criteria

- [ ] `reindex --search` (no `--full`) removes `search_index` rows of `type='observation'`/`type='relation'` whose backing row no longer exists.
- [ ] A seeded DB↔FTS divergence on an unchanged file converges via `reindex --search` alone (regression test).
- [ ] No full-file re-parse is required for the purge path; `--full` behavior unchanged.

Refs: #1214 (field report), #1213.

Issue #1213, opened 2026-08-09
Generation-versioned relation persistence: eliminate indexing deadlocks on the relation table


phernandez:
## Problem

Concurrent note indexing deadlocks on the tenant `relation` table in production, and the deadlocked jobs exhaust all 5 PGQueuer attempts and terminally fail — leaving notes unindexed until a later unrelated touch.

16 distinct Logfire fingerprints in the basic-memory-cloud project over 2026-08-01 → 2026-08-09 (#2256 #2310 #2323 #2328 #2339 #2348 #2367 #2377 #2378 #2380 #2382 #2384 #2385 #2387 #2388 #2395), across both `index_file` and `index_file_batch` entrypoints.

Trace-proven shapes (representative traces, auth-gated):

- [#2395](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019fe3e469b6f5f96215dd3f324c81aa%27&since=2026-08-09T00%3A20%3A25.142259%2B00%3A00&until=2026-08-09T01%3A20%3A33.351444%2B00%3A00): `index_file` attempt 5/5, deadlock in `relation_repository.delete_outgoing_relations_from_entity` (`DELETE FROM relation WHERE from_id = $1 AND project_id = $2`), blocking PIDs correlated to concurrent `index_file` jobs for the same tenant, including a three-process cycle.
- [#2388](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019fdf29ed25efc77d4727ab4199fb25%27&since=2026-08-08T02%3A18%3A14.693116%2B00%3A00&until=2026-08-08T03%3A18%3A28.390151%2B00%3A00): `index_file_batch` attempt 5/5, deadlock in `relation_repository.add_all_ignore_duplicates` (bulk `INSERT INTO relation ... ON CONFLICT DO NOTHING RETURNING id`) — with **two simultaneous `index.markdown_file.persist` children inside the same batch job**, one succeeding and one deadlocking (self-inflicted intra-job concurrency).
[...]

PR #1228, opened 2026-08-11
fix(core): fence observation persistence behind note content generation

### LINKED ISSUE #1214: Generation-versioned observation persistence (companion to relation deadlock fix)
## Problem

Observation persistence has the same delete-and-replace transaction shape as relation persistence: markdown persist deletes an entity's observation rows and re-inserts the new set inside the same multi-phase transaction as entity upserts and relation writes. This is the same lock structure that produces the production `relation`-table deadlock cluster (https://github.com/basicmachines-co/basic-memory/issues/1213), with lower observed collision frequency so far — concurrent persists of notes touching overlapping entities can interleave lock acquisition across phases.

No dedicated Logfire fingerprint has been attributed to observation rows yet; some of the 16 relation-cluster fingerprints' retried attempts recorded adjacent integrity/DBAPI failures consistent with contention beyond the relation table. This issue is the proactive completion of the same fix rather than a response to a distinct incident.

## Design

Adopt the identical generation-versioned model specified in https://github.com/basicmachines-co/basic-memory/issues/1213:

1. `observation` rows carry an indexed `generation` — the note's existing generation token (`db_version`/content checksum).
2. Persist upserts the desired observation set with `ON CONFLICT DO UPDATE SET generation = G` guarded by `WHERE observation.generation < G`; stale writers become no-ops.
3. Cleanup of superseded rows (`DELETE ... WHERE entity_id = ? AND generation < G`) is a separate short statement after upserts — brief union window, never a gap; observations are derived, eventually consistent state.
4. Observation writes commit in their own short transaction, decoupled from entity/relation phases.

Note: observations may lack a natural conflict key today (position/content-based identity) — if so, part of this issue is defining a stable upsert identity (e.g. content-hash within entity) so `ON CONFLICT` has an arbiter, or an equivalent generation-swap strategy that avoids delete-then-insert in one transaction.

## Scope

Core repo: observation repository + markdown persist flow, tenant-DB migration (`observation.generation` + index + backfill, plus upsert identity if needed). Sequence after (or with) the relation issue so both use one migration window and one persistence-flow refactor.

## Acceptance criteria

- [ ] Observation persistence performs no delete-then-insert inside a shared multi-phase transaction; writes are generation-guarded with separate bounded cleanup.
- [ ] A stale generation's observation writes are no-ops and cleanup cannot delete newer-generation rows (both race directions tested).
- [ ] Concurrent persists of notes with overlapping entities complete without deadlock in the same real-Postgres concurrency test added for relations.
- [ ] Migration backfills `generation` (and upsert identity if introduced) for existing rows.

phernandez:
Closes #1214. Companion to the relation generation fix (#1213 / #1220), plus the field
report from @apleith on the issue.

## Design: fenced replace, deliberately simpler than the relation mirror

Observations don't need the full generation machinery (generation column, identity key,
ON CONFLICT upserts): they're single-entity scoped, have no cross-entity FKs, and a note's
whole set fits one statement batch. Instead, the delete-then-insert moves out of the shared
multi-phase transactions into **its own short transaction guarded by the existing
note_content generation fence** (`current_relation_generation_statement`, `db_version == G`,
`FOR UPDATE` on Postgres; SQLite's whole-transaction serialization covers it):

- Fence miss → stale writer no-ops, rows untouched.
- Fence hit → replace the entity's observation set atomically. An empty set is a legitimate
  wipe (note edited to zero observations).
- Deadlock-free by the canonical NoteContent-first order; same-entity publishers serialize
  on the fence row. This also ends the SQLite silent duplication from the field report — a
  writer whose delete lost the race now no-ops instead of re-inserting.

This revises the issue's acceptance criteria: there is **no schema change and no generation
backfill** — the migration is data-only. In-note exact duplicate lines remain preserved
(fidelity with the markdown); per-persist row-id churn is unchanged (status quo).

## What changed

- `ObservationRepository.replace_observations_for_generation` replaces
  `replace_accepted_observations`: fence check, then delete + insert, returning
  `generation_is_current`.
- Observations ride the existing post-commit publication: `RelationGenerationPublication`
  carries them, and `RelationGenerationPublisher.publish` runs begin-marker → fenced
  observation replace (own transaction, early return on stale) → relation chunks → cleanup.
  One fence lifecycle, one durable retry marker.
- Accepted path: `_replace_accepted_note_observations` deleted from
  `persist_accepted_note_snapshot` — the graph no longer commits inside the accepted write
  transaction. **Moves carry the note's parsed observations** (`PreparedEntityMove` /
  `AcceptedPreparedNoteMove`), since a move advances `db_version` and an empty publication
  would otherwise wipe the note's observations (regression-tested).
- File-index path: the delete+insert is stripped from the renamed
  `update_markdown_entity_fields` (was `update_entity_and_observations`);
  `_publish_markdown_graph` (was `_publish_markdown_relations`) publishes observations and
  relations together. Batch/single-file indexers thread observations through
  `SyncedMarkdownFile`/`IndexedEntity`.
- Data-only migration `2d26b287813b`: dedupes existing duplicate rows keeping `MIN(id)` per
  `(entity_id, category, content, context, tags)` (the field report's mitigation;
  `tags::text` on Postgres where `json` has no equality), then purges
  `search_index` rows of `type='observation'` whose backing row is gone, guarded by a
  table-exists check. Downgrade is a documented no-op.

Follow-up filed: #1226 — incremental `reindex --search` can't repair DB-only search
divergence in general (root cause: checksum-based file selection).

## Verification

- Focused suites (repository, publisher, migration, API lock-order): 34 passed.
- Broad `tests/indexing/` + `tests/services/`: 904 passed, 3 skipped.
- Real-Postgres concurrency (`test-int/test_relation_generation_concurrency.py`): 4 passed —
  the mutual-persistence test now seeds observation lines and asserts the final sets match
  the markdown exactly with zero duplicate `(entity_id, category, content, context, tags)`
  tuples after concurrent batch + single-file passes.
- `just typecheck`, `just lint`, `just doctor`: clean.


phernandez:
Verified the premise before answering: correct, `_persist_markdown_file` stamps the final
checksum in the same transaction as the persist, so a transient search-refresh failure does
leave the old observation rows in place with the file skipped by later scans-by-checksum.

Declining the finding anyway — this is accepted eventual consistency, not a gap to plumb:

- It is not introduced by this PR. Relations have had the identical structure since #1220:
  a failed search refresh means the publisher never ran for that file, for relations and
  observations alike. This PR just makes observations consistent with that model.
- The window requires a transient FTS/database error at exactly the search-refresh step,
  moments after a successful persist on the same connection pool — rare in practice.
- The state converges through existing mechanisms: the next edit of that note, a
  `reindex --full`, or `doctor`. The drift is stale derived rows lagging canonical
  markdown, which is the documented design (see the module docstring in
  `relation_persistence.py`), never a wrong canonical state.
- Recording retry work before the fallible phase adds a marker lifecycle to guard a rare
  transient window whose repair already exists. Deadlocks and shared-transaction coupling
  are what this line of work is removing; we are not re-adding machinery at its edges.

Review guidance for this codebase now covers this class of finding: AGENTS.md
"Consistency Model — Review Expectations" (#1230). Derived-state races are findings only
when they converge to a wrong state that no existing mechanism repairs.

PR #1230, opened 2026-08-11
docs(core): set review expectations for eventually consistent state

phernandez:
Adds a "Consistency Model — Review Expectations" section to AGENTS.md so reviewers — the
Codex PR bot in particular — stop proposing lock/transaction machinery for derived-state
paths.

Context: every Codex review on #1227 and #1228 has flagged eventually-consistent projection
behavior as a bug and proposed SELECT-time locks, wider transactions, or compensating retry
markers. That class of "fix" is what produced the production deadlock clusters (#1213,
#1224) and the silent observation duplication (#1214). The new section draws the canonical
vs derived state line, lists the existing convergence mechanisms (next write, index pass,
reindex, doctor, cloud orphan sweeper), and defines the only shape of derived-state race
that warrants a finding: one that converges to a wrong state no existing mechanism repairs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -153,25 +153,33 @@ agents must apply this skill before changing or evaluating Python code.
 Basic Memory separates canonical state from derived state, and reviews (human or automated)
 must hold them to different standards:
 
-- **Canonical state** — the markdown files on disk and the accepted `note_content` row.
-  Writes are guarded: CAS on `db_version`, generation fences, checksum-guarded file
-  operations. Correctness findings here are welcome.
+- **Canonical state** — the markdown bytes of a note (on disk and as accepted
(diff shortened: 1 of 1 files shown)


phernandez:
All three findings accepted and applied — these are exactly the finding shape the section asks for (precise wrong-state claims, no serialization proposals): (1) the lock bullet now bans *adding* serialization while explicitly keeping violations of the canonical NoteContent-first ordering reviewable; (2) canonical state is now the note's markdown *bytes* wherever they live — materialization writes canonical content into portable form, so content corruption/wrongful deletion is always a real finding — and only materialization status/lineage timing is derived, per docs/DOMAIN_MODEL.md; (3) doctor removed from the repair mechanisms (it diagnoses in a temporary project, it does not repair).

PR #1258, opened 2026-08-14
fix(sync): guard project-index deletes against materialization

### LINKED ISSUE #1256: Project indexing can delete accepted-but-unmaterialized notes (production data loss)
## Summary

A project-wide index can delete a valid DB-first note while its Markdown file is still waiting for asynchronous materialization.

This is confirmed production data loss. The caller receives `202` / Created, but the initial project index can later classify the accepted database row as an externally deleted file, remove `Entity` (cascading `NoteContent`), and leave no durable note or file.

Production Logfire issue [#2489](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/issues/2489) is one observable interleaving. Other interleavings return `status=missing` normally and do not create an exception issue.

## Production evidence

Observed on 2026-08-13 with Cloud release `bc2e7e889e16552052094e21a99a7b03cb71005c` and Core `0.22.2.dev219+38a2ed67` (`38a2ed675f7f33f23be129f3948ff52ec5f023b3`).

Representative fresh-project sequence, with tenant and content identity redacted:

1. An MCP `create_memory_project` call created an empty project and immediately queued project-index job `9497726`.
2. A following MCP `write_note` accepted entity `10`, `NoteContent` version `1`, returned Created, and queued materialization job `9497727`.
3. PGQueuer picked both jobs at approximately `19:24:46Z`.
4. The index listed the project prefix from S3 and found zero objects.
5. `ChangeDetector.load_all_indexed_paths()` nevertheless returned the new DB-only note, so delete planning reported one deleted file.
6. Project-index maintenance locked and deleted the entity; `ON DELETE CASCADE` removed `note_content`.
7. Materialization had already loaded both ORM rows. Its `session.flush()` then raised:
   `sqlalchemy.orm.exc.StaleDataError: UPDATE statement on table 'note_content' expected to update 1 row(s); 0 were matched.`
8. Current tenant state contains neither entity `10` nor its `note_content`, and no later materialization job exists. This job failed before writing S3.

Traces:

- [Accepted MCP write](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%2734c5455c30eae6e4a61140e5a8536b95%27&since=2026-08-13T19%3A24%3A02.838247%2B00%3A00&until=2026-08-14T19%3A24%3A05.367077%2B00%3A00)
- [Concurrent project index](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffc9589daa6c2a87df627b4deea18%27&since=2026-08-13T19%3A24%3A46.426347%2B00%3A00&until=2026-08-14T19%3A24%3A53.867520%2B00%3A00)
- [Failed materialization / Logfire #2489](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffc9589dfa83be793bceabd6d712b%27&since=2026-08-13T19%3A24%3A46.431396%2B00%3A00&until=2026-08-14T19%3A24%3A52.649292%2B00%3A00)

The exception is not the only production signature. Over the latest 14-day query window, materialization emitted nine normal `status=missing` results across five tenants. At least two on 2026-08-13 align within one second with a same-tenant, same-project index deleting exactly one row:

- [15:16 index deletion](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffbb2409fda315516856d680d1b88%27&since=2026-08-13T15%3A16%3A31.007378%2B00%3A00&until=2026-08-13T16%3A16%3A37.600334%2B00%3A00) followed by [materialization reporting that note state disappeared after file write](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffbb240ad4b5f105fe99c8ed49ab7%27&since=2026-08-13T15%3A16%3A31.021555%2B00%3A00&until=2026-08-13T16%3A16%3A38.483854%2B00%3A00).
- [18:20 index deletion](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffc5a41828e12151e1443aa47e4c7%27&since=2026-08-13T18%3A20%3A01.282755%2B00%3A00&until=2026-08-13T19%3A20%3A07.332678%2B00%3A00) followed by [same-project materialization reporting missing state](https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%27019ffc5a56d1fb01eef44ce62d66bcad%27&since=2026-08-13T18%3A20%3A06.737636%2B00%3A00&until=2026-08-13T19%3A20%3A07.531851%2B00%3A00).

This gives one exception-confirmed and two independently correlated production occurrences today. The other six missing outcomes may be legitimate delete-after-enqueue races and are not claimed as data loss without further correlation.

## Root cause

The consistency model is DB-first for accepted Markdown, but project delete detection currently treats every indexed entity path as if its storage projection must already exist:

- `ChangeDetector.load_all_indexed_paths()` delegates to `EntityRepository.get_all_file_paths()`.
- `plan_file_changes()` computes deletes as `all_db_paths - storage_paths - moved_old_paths`.
- No `NoteContent.file_write_status`, `db_version`, or `file_version` state is carried into delete planning.
- `RepositoryProjectIndexMaintenanceStore` defaults to `TrustPlannedProjectIndexDeleteVerifier`. The S3 runtime therefore trusts the scan verdict at apply time.
- Project-index deletion correctly takes the shared `NoteContent` lock before deleting `Entity`, but locking cannot make the deletion semantically valid; it only decides which transaction wins.

A newly accepted note is a valid state even when its file is absent. Treating storage as authoritative during that window destroys the durable source of truth.

The `StaleDataError` is a secondary ORM check-then-flush race: preflight reads live models, project-index deletion removes them, and the generated UPDATE asserts one matched row. Converting only that exception to a missing/no-op result would hide the error but would not prevent data loss.

## Required design

Keep asynchronous materialization. Do not make the request wait for S3 and do not use timing delays or queue ordering as the correctness boundary.

The fix should establish both invariants:

1. Project-index delete planning/application cannot delete accepted content whose current file projection is not proven synchronized.
2. Materialization preflight handles a legitimate concurrent explicit delete without leaking `StaleDataError`.

A likely shape is:

- load typed delete candidates rather than bare paths, carrying the minimum accepted/materialized lineage needed to distinguish storage-owned indexed files from DB-owned pending content;
- exclude `pending`, `writing`, `failed`, and otherwise unsynchronized current DB versions from external-delete planning;
- revalidate the same lineage with a guarded SQL predicate at delete application so an accepted write or materialization transition after the scan cannot be deleted from a stale plan;
- use a guarded lower-level `UPDATE ... WHERE entity/project/version/checksum still match RETURNING ...` (or equivalent row-locking CAS) for materialization preflight and return a detached value object; zero rows must be classified from current state rather than becoming an ORM row-count exception.

The exact predicate needs to preserve non-Markdown/storage-origin entities and true external deletion of previously synchronized Markdown.

## Acceptance criteria

- [ ] A Postgres concurrency regression runs the real fresh-project sequence: initial project index is delayed, `write_note` is accepted, and index plus materialization are released concurrently.
- [ ] After all work settles, the entity, `note_content`, search state, and S3 object all exist and agree on current path/checksum/version.
- [ ] The regression fails against the pre-fix implementation by deleting the accepted note.
- [ ] Both race orders are covered: index attempts deletion before materialization claims the row, and after preflight has begun.
- [ ] Project delete planning excludes accepted-but-unmaterialized current content without relying on age, sleep, FIFO job order, or process-local locks.
- [ ] Delete application revalidates its snapshot/CAS state so a concurrent accepted write cannot be destroyed.
- [ ] A genuinely externally deleted, previously synchronized file still removes its entity and derived search/vector rows.
- [ ] An explicit user delete racing a queued materialization remains idempotent: the materializer reports a typed missing/stale outcome and guarded cleanup cannot remove a newer file.
- [ ] Materialization preflight no longer leaks `StaleDataError` when its target disappears.
- [ ] Cloud integration coverage exercises the S3/PGQueuer adapter composition, not only an in-memory planner.
- [ ] Production acceptance queries show no same-tenant/project index deletion followed by materialization missing/error for the deleted generation.
- [ ] Identify affected production writes and define recovery before closing; do not assume exception issue count equals impact.

## Related

- basicmachines-co/basic-memory#1159 covers relation resolution reading pending materialization, not deletion of the accepted source of truth.
- basicmachines-co/basic-memory#1187 fixed a different Entity/NoteContent lock-order deadlock. This issue is about an invalid external-delete decision; row-lock ordering alone cannot solve it.
- basicmachines-co/basic-memory-cloud#1661 tracks truthful production operation outcomes. The normal `status=missing` interleavings need causal diagnostics there, but the deletion invariant belongs in Core.

phernandez:
## Why

Project indexing compares a storage listing with database state observed later. A note accepted
after that listing but before delete maintenance could therefore be classified as externally
deleted, even when its canonical `NoteContent` had not yet reached storage. Applying that stale
plan deleted the accepted entity and note content.

This closes #1256 by making destructive project-index maintenance prove both sides of the delete
decision at apply time: the database row must be in a positive delete-eligible state and the path
must still be absent from current storage.

## What Changed

- Model project-index deletion eligibility as typed positive candidates:
  - storage-owned entities without `NoteContent`
  - accepted notes whose file status, version, and checksum are fully synchronized
- Require scan runtimes to inject a live path verifier; project scans now fail closed without one.
- Re-probe candidate paths outside the database transaction, then lock and revalidate the exact
  candidate lineage before deletion.
- Replace materialization's ORM load/mutate/flush preflight with a guarded
  `UPDATE ... RETURNING` claim.
- Add real PostgreSQL regression coverage for both materialization/delete winner orderings.
- Document the repository's PostgreSQL Read Committed assumptions and canonical lock order in
  `docs/TRANSACTION_ISOLATION.md`.

## Implementation Details

Deletion now runs in three phases:

1. Snapshot only valid DB candidates. Pending, writing, failed, externally changed, and partially
   synchronized notes are protected before any storage call.
2. Positively verify current path absence outside a database transaction.
3. Lock `NoteContent` before `Entity`, reload the candidates under Read Committed, and delete only
   rows whose full lineage still matches the snapshot.

The materialization-attempt timestamp participates in candidate equality. That closes the
same-version ABA window where materialization can move from `synced` to `writing`, create the file,
and return to `synced` while the absence probe is in flight.

Materialization claims the queued `(project_id, entity_id, db_version, db_checksum)` in one
conditional update. PostgreSQL waits on a conflicting delete and re-evaluates the predicate against
the current row version, so a delete winner becomes a typed missing result instead of an expected
`StaleDataError`.

Database locks remain narrow and are never held across storage I/O. The existing
`NoteContent`-before-`Entity` order is preserved.

### Automated

- `just fast-check`: passed
- `uv run pytest tests/indexing/test_note_materialization_runner.py tests/indexing/test_project_index_maintenance.py tests/indexing/test_project_index_runtime.py -q --no-cov`: 65 passed
- `BASIC_MEMORY_TEST_POSTGRES=1 uv run pytest test-int/test_project_index_delete_materialization_race.py -q --no-cov`: 2 passed
- `just doctor`: passed
- `git diff --check origin/main...HEAD`: passed

### Manual

- Reviewed the concurrency model against the PostgreSQL transaction-isolation and explicit-locking
  documentation linked from `docs/TRANSACTION_ISOLATION.md`.

## Risks / Follow-ups

- Eligible scan-planned deletes now pay for a current per-path storage probe. The probe happens only
  after DB eligibility filtering and outside the transaction.
- Cloud must inject its strongly consistent Tigris HEAD verifier after this Core API lands. That
  companion change is implemented and tested separately, but its dependency pin cannot be updated
  until this commit is available to Cloud.
- No production recovery or deployment is included in this PR.

## Reviewer Guide

1. Review the candidate model and three-phase delete flow in
   `project_index_maintenance.py`.
2. Review the guarded claim in `note_materialization_runner.py`.
3. Read `test_project_index_delete_materialization_race.py` for the two PostgreSQL interleavings.
4. Use `docs/TRANSACTION_ISOLATION.md` as the durable statement of the concurrency contract.

### DIFF
--- a/docs/ENGINEERING_STYLE.md
+++ b/docs/ENGINEERING_STYLE.md
@@ -100,6 +100,10 @@ For example, prefer separate `Completed(result)` and `Failed(reason)` values joi
 
 ## Database Writes And Concurrency
 
+See [TRANSACTION_ISOLATION.md](TRANSACTION_ISOLATION.md) for the PostgreSQL Read Committed
+semantics, guarded-write patterns, and NoteContent-before-Entity lock protocol that repository
+code may rely on.
+
 - Let the database arbitrate ordinary write contention. Express idempotency and uniqueness with
   constraints plus `INSERT ... ON CONFLICT DO NOTHING` or `DO UPDATE`, and return an explicit
   inserted, existing, or updated outcome when callers need to distinguish them.
--- a/docs/TRANSACTION_ISOLATION.md
+++ b/docs/TRANSACTION_ISOLATION.md
@@ -1,0 +1,107 @@
+# Transaction Isolation And Concurrency
+
+Basic Memory's PostgreSQL code assumes the database's default **Read Committed** isolation level.
+SQLite remains a supported local/test backend, but its single-writer behavior is not evidence that
+a PostgreSQL row-lock protocol is correct. Any design that depends on concurrent row behavior must
+have a PostgreSQL regression test.
+
+This document records the specific guarantees repository code may use. It is not permission to add
+broad locks: optimistic compare-and-swap remains the default, and canonical note safety is the main
+case where a narrow lock is justified.
+
+## Read Committed Statement Model
+
+Under PostgreSQL Read Committed:
+
+- Each command sees rows committed before that command began, plus changes already made by its own
+  transaction. Two commands in one transaction can therefore observe different committed state.
+- `UPDATE`, `DELETE`, and locking `SELECT` statements wait when a target row is concurrently
+  updated, deleted, or locked.
+- After the wait, PostgreSQL applies the operation to the current row version and re-evaluates the
+  command's `WHERE` predicate. If the row was deleted or no longer matches, the command affects no
+  row.
+- Row locks last until transaction end. `FOR UPDATE` conflicts with the weaker `FOR KEY SHARE`
+  lock used to protect referenced keys, which is relevant when a child row is inserted through a
+  foreign key.
+
+The authoritative references are PostgreSQL's
+[Transaction Isolation](https://www.postgresql.org/docs/current/transaction-iso.html) and
+[Explicit Locking](https://www.postgresql.org/docs/current/explicit-locking.html) chapters.
+
+Read Committed is deliberately not a transaction-wide snapshot. Never assume an earlier plain
+`SELECT` still describes the row when a later statement writes it.
+
+## Preferred Guarded-Write Pattern
+
+Put the expected lineage in the write itself and inspect whether it matched:
+
+```sql
+UPDATE note_content
+SET file_write_status = 'writing',
+    last_materialization_attempt_at = :attempted_at
+WHERE project_id = :project_id
+  AND entity_id = :entity_id
+  AND db_version = :db_version
+  AND db_checksum = :db_checksum
+RETURNING markdown_content, file_checksum;
+```
+
+This is a compare-and-swap claim, not a read followed by an ORM flush. If a concurrent delete wins,
+the statement returns no row. If the claim wins, its NoteContent row lock remains held until commit,
+and a later delete must observe the changed materialization lineage. Expected contention is returned
+as a typed missing or stale outcome; it is not an ORM `StaleDataError` control path.
+
+Use this pattern when one statement can name the invariant. Do not add a pre-read lock merely to
+make a later unconditional write appear safe.
+
+## Canonical Lock Order
+
+When a transaction must touch accepted `NoteContent` and then mutate or delete its `Entity` or
+relations, it acquires locks in this order:
+
+1. Lock every relevant `NoteContent` row, sorted by `entity_id`.
+2. Lock the corresponding `Entity` rows, also in stable order, only when the operation needs an
+   explicit entity fence.
+3. Mutate Entity, search, relation, or vector projection rows.
+
+The shared entry point is
+`lock_note_content_before_entity_mutation` in
+`src/basic_memory/repository/relation_repository.py`. Reversing this order can deadlock accepted
+writes, materialization, relation publication, and cascading deletes.
+
+The Entity lock matters for a legacy indexed row with no `NoteContent`. A concurrent accepted write
+can bootstrap that child row. Locking the referenced Entity makes PostgreSQL serialize the foreign
+key insertion; a subsequent Read Committed statement then either sees the committed `NoteContent`
+and preserves it, or the inserter waits until deletion completes and cannot accept content for a
+missing Entity.
+
+## External Storage Is Outside The Transaction
+
+Object storage and the local filesystem cannot participate in the database transaction. Project
+index deletion therefore uses three explicit phases:
+
+1. Snapshot only valid DB delete candidates: storage-owned entities with no `NoteContent`, or notes
+   where status is `synced`, `file_version == db_version`, and
+   `file_checksum == db_checksum`.
+2. Outside a DB transaction, positively verify that each candidate path is still absent from the
+   current storage backend. A scan listing alone is not sufficient.
+3. Lock NoteContent then Entity, reload the exact candidate lineage in a new Read Committed
+   statement, and delete only candidates equal to the snapshot.
+
+The materialization-attempt timestamp is part of a synchronized note candidate. It prevents an ABA
+window where a same-version materialization leaves `synced`, writes the file, and returns to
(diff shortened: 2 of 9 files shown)


phernandez:
@codex review

PR #1261, opened 2026-08-15
docs(core): fix broken star history chart in README

OctoBored:
The star history chart at the bottom of the README is currently broken and shows nothing, which looks bad for the project's main landing page. The chart was served by a service that no longer works, so this PR switches it to the replacement that is already widely used across open source projects.

No other content in the README is changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -689,11 +689,11 @@ Contributions welcome — see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Star History
 
-<a href="https://www.star-history.com/#basicmachines-co/basic-memory&Date">
+<a href="https://star-history.dera.page/#basicmachines-co/basic-memory&Date">
  <picture>
-   <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/svg?repos=basicmachines-co/basic-memory&type=Date&theme=dark" />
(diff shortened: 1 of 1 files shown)


phernandez (commented):
The replacement SVG endpoint is live, but `star-history.dera.page` is an independently operated fork rather than the official Star History domain. Before we move a README image (and its viewer requests) to that third party, could you add provenance for the operator/data source and why this is the preferred replacement? Please also amend the commit with a Signed-off-by trailer so DCO passes.

phernandez:
api.star-history.com is up and serving this repo's chart (verified: HTTP 200, valid SVG), so the premise of this PR does not hold. Redirecting the README to an unrelated third-party domain is not something we will accept. Closing.

phernandez:
Correction for the record: the chart is in fact broken — GitHub restricted the stargazers API on 2026-06-30, and the star-history SVG now renders an error card rather than a chart (the HTTP 200 in my earlier check was that error card). The close stands: redirecting the README to an unrelated third-party domain is not the fix. Tracked properly in #1287.

OctoBored:
It is the best fix out there. Requires no setup for you, open-source, operated by volunteers, and there is no commercial aspect (the original project is a cash cow)

PR #1265, opened 2026-08-16
fix: apply FastEmbed query/passage asymmetry via query_embed

### LINKED ISSUE #1264: Document query/passage prefixes for asymmetric FastEmbed models
Thanks for basic-memory — it is the retrieval backbone of my lab's bilingual (EN/JP) research knowledge base, queried daily by both humans and agents, and the local-first design is exactly why we can run it on real research notes at all.

While tuning multilingual retrieval I found a small but consequential gap in the FastEmbed path, with a fix that looks well-contained. Details and measurements below.

## Environment

- basic-memory v0.22.1, sqlite backend
- Embedding provider: FastEmbed (local)
- Corpus: bilingual (English/Japanese) research notes

## What happens

Asymmetric-prefix embedding models — notably the multilingual-e5 family, which FastEmbed itself ships (`intfloat/multilingual-e5-large`) — run off-label on the FastEmbed path: queries and documents are embedded identically, so the model's query/passage distinction is never applied and retrieval quality collapses.

## Why

Two code pointers:

1. `src/basic_memory/repository/fastembed_provider.py`: `embed_query()` simply calls `embed_documents([text])`. No query/passage asymmetry is ever applied. FastEmbed exposes `query_embed()` for exactly this purpose, but it is never called.
2. The config fields `semantic_embedding_query_input_type` / `semantic_embedding_document_input_type` exist, but are only honored by the litellm provider (`litellm_provider.py`) — never by the FastEmbed provider. So configuring them has no effect on the local path.

## Measurements

All measured today on real note chunks, through basic-memory's exact code path:

- **`multilingual-e5-large` without query prefix:** true-target vs unrelated-junk cosine margin collapses to **0.031** (Japanese query) and **0.016** (English paraphrase query) — effectively no discrimination.
- **Why e5 matters:** for a bilingual corpus the multilingual-e5 family is essentially the only strong local retriever in FastEmbed's registry. The default `bge-small-en-v1.5` is English-only — a Japanese query scores 0.52 against its true target vs **0.74 against unrelated Japanese text** (it encodes the language, not the content) — and the paraphrase-multilingual family truncates at 128 tokens, below typical chunk length.

## Suggested fix (deliberately minimal)

Either of these unlocks the e5 family with no schema or API changes:

1. Have `FastEmbedEmbeddingProvider.embed_query()` call the underlying model's `query_embed()` when available, falling back to `embed()`; and/or
2. Honor the existing `*_input_type` config fields as literal prefixes ("query: " / "passage: ") on the FastEmbed path, mirroring what the litellm provider already does.

## Offer to help

We maintain a bilingual golden set (100 queries, EN+JP) over this corpus. Happy to test a branch against it and report recall numbers before/after — just point me at it. And thanks again; this tool carries a lot of weight for us, which is the only reason a 0.031 margin ever got noticed.

Aryan-Pardeshi:
## Summary

`FastEmbedEmbeddingProvider.embed_query()` delegated to `embed_documents()`, so asymmetric models such as `bge-small-en-v1.5` embedded queries through the passage/document path and silently lost the model's query-side instruction — the retrieval-quality regression measured in #1264.

## Changes

- `embed_query` now routes through `TextEmbedding.query_embed(text)` in a worker thread, which is FastEmbed's query path for asymmetric models.
- The L2 normalization (required by `sqlite_search_repository`'s unit-vector distance contract) and the dimension mismatch check are shared with `embed_documents` via a `_normalize_vectors` helper — both paths keep identical post-processing.
- Documents are untouched: `embed_documents` behavior, identity keys, and stored vectors stay valid, so no re-indexing is triggered.

## Notes

- `semantic_embedding_document_input_type` / `semantic_embedding_query_input_type` remain LiteLLM-only: FastEmbed has no `input_type` request parameter — the role split *is* `query_embed` vs `embed`, so no config plumbing is needed for the native provider.
- `query_embed` accepts no `batch_size`/`parallel` kwargs (single query string), so the runtime knobs stay on the document path only.

## Testing

- New regression tests: query path uses `query_embed` (asserted distinct query vs passage vectors, and that `embed` is never called), plus unit-norm and dimension-mismatch coverage for query vectors.
- Full `tests/repository/` suite + semantic embedding watch tests: 654 passed, 30 skipped.
- `ruff check`, `ruff format --check`, and `ty check` clean.

Closes #1264

### DIFF
--- a/src/basic_memory/repository/fastembed_provider.py
+++ b/src/basic_memory/repository/fastembed_provider.py
@@ -276,6 +276,20 @@ class FastEmbedEmbeddingProvider:
             )
             return self._model
 
+    def _normalize_vectors(self, vectors) -> list[list[float]]:
+        # sqlite_search_repository.py uses a distance-to-similarity formula that assumes
+        # unit-normalized vectors (see the comment on line 65-67 of that file).
+        # Some models (e.g. multilingual ones) return vectors with norm > 1, so we
+        # L2-normalize here to satisfy that contract regardless of the chosen model.
+        normalized: list[list[float]] = []
+        for vector in vectors:
+            values = vector.tolist() if hasattr(vector, "tolist") else list(vector)
+            norm = math.sqrt(sum(x * x for x in values))
+            if norm > 0:
+                values = [x / norm for x in values]
+            normalized.append([float(v) for v in values])
+        return normalized
+
     async def embed_documents(self, texts: list[str]) -> list[list[float]]:
         if not texts:
             return []
@@ -297,19 +311,7 @@ class FastEmbedEmbeddingProvider:
             embed_kwargs: dict[str, int] = {"batch_size": self.batch_size}
             if effective_parallel is not None:
                 embed_kwargs["parallel"] = effective_parallel
-            vectors = list(model.embed(texts, **embed_kwargs))
-            # sqlite_search_repository.py uses a distance-to-similarity formula that assumes
-            # unit-normalized vectors (see the comment on line 65-67 of that file).
-            # Some models (e.g. multilingual ones) return vectors with norm > 1, so we
-            # L2-normalize here to satisfy that contract regardless of the chosen model.
-            normalized: list[list[float]] = []
-            for vector in vectors:
-                values = vector.tolist() if hasattr(vector, "tolist") else list(vector)
-                norm = math.sqrt(sum(x * x for x in values))
-                if norm > 0:
-                    values = [x / norm for x in values]
-                normalized.append([float(v) for v in values])
-            return normalized
+            return self._normalize_vectors(model.embed(texts, **embed_kwargs))
 
         vectors = await asyncio.to_thread(_embed_batch)
         if vectors and len(vectors[0]) != self.dimensions:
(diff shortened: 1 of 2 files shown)


phernandez (changes requested):
This does improve models whose FastEmbed implementation overrides `query_embed()`, but it does not fix the model reported in #1264.

With the repository's locked FastEmbed 0.8.0, `intfloat/multilingual-e5-large` is dispatched to `PooledEmbedding`. That class does not override either role method; its inherited `OnnxTextEmbedding.query_embed()` simply delegates to `embed()`. The new stub invents a distinct query vector, so it proves the call site changed but not that multilingual E5 receives the required `query: ` / `passage: ` inputs. The PR's “no re-index” claim is therefore also incompatible with the reported fix: applying the correct passage prefix changes stored document vectors.

Please wire the E5 defaults through the existing `PrefixingEmbeddingProvider`/provider-identity path (while retaining `query_embed()` for models that implement it), make the changed document semantics trigger reindexing, and add a regression where the underlying `query_embed()` delegates to `embed()` but the recorded inputs still receive the two role prefixes. The title also needs an allowed scope such as `fix(core): ...`, and DCO/CLA must clear before merge.

phernandez:
Thanks for digging into this — the direction (asymmetric query embedding) is right, and the normalization refactor was clean. The blocker is that in fastembed 0.8.0 `query_embed` is a passthrough to `embed` for the default BGE model and for the E5 family from #1264 (only Jina v3 overrides it), so this change produces identical vectors for the reported case — details in the review threads.

The fuller fix (per-model E5 prefixes through the existing prefixing/identity path, which also drives reindex automatically) is now tracked in #1284, with credit to you for this attempt. Closing this PR; you'd be welcome to pick up #1284 if you're interested — note it would need a signed CLA and DCO sign-offs.

PR #1277, opened 2026-08-19
docs(hermes): clarify version-specific monkeypatch

groksrc:
## Summary

- split the Hermes slash-command workaround into explicit modern and legacy paths
- document that Hermes v0.20.1 / v2026.8.13 and newer require only the active-provider startup-load change in `hermes_cli/plugins.py`
- warn that replacing the modern lifecycle-aware collector can leave stale commands and skills after provider unload
- provide version-specific verification commands

## Why

Follow-up to #1274. That PR updated the compatibility table but the detailed instructions still unconditionally listed both Hermes files and included the obsolete collector replacement. A modern Hermes user following the full document could therefore undo the lifecycle ownership fix that #1274 preserves.

## User impact

Users can now select the correct patch from their installed Hermes version without applying the legacy collector workaround to modern releases.

## Validation

- Hermes plugin manifest validation passed
- Hermes test suite: 269 passed, 12 skipped
- `git diff --check` passed

### DIFF
--- a/integrations/hermes/MONKEYPATCH.md
+++ b/integrations/hermes/MONKEYPATCH.md
@@ -4,7 +4,7 @@ This document describes the Hermes Agent-side patch needed when `/bm-*` slash co
 
 This issue is tracked upstream in https://github.com/NousResearch/hermes-agent/issues/23603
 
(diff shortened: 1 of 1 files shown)


groksrc:
Superseded by #1278, which carries the same documentation change with the repository-required DCO sign-off.

PR #1279, opened 2026-08-19
fix: allow subdirectory traversal during glob-filtered direc…

kdahal7:
…tory listing

In DirectoryService._collect_nodes_recursive, if file_name_glob was set and a directory name did not match the glob, the loop executed 'continue', skipping recursion into the directory. As a result, files in subdirectories were never collected when listing with a file glob and depth > 1. Updated the loop to match nodes against the glob for inclusion in results while continuing recursion into subdirectory nodes.

### DIFF
--- a/src/basic_memory/services/directory_service.py
+++ b/src/basic_memory/services/directory_service.py
@@ -391,12 +391,9 @@ class DirectoryService:
             return
 
         for child in node.children:
-            # Apply glob filtering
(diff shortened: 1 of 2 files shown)


phernandez:
Thanks @kdahal7 — the fix is correct and the regression test verified red/green. Fork PRs don't run this repo's test workflows, so we've rehomed your commit (authorship preserved) onto an in-repo branch with a small formatting/assertion follow-up: #1285. Closing this one in its favor; you'll be credited in the merged history.

PR #1286, opened 2026-08-20
test(mcp): add concurrent write_note MCP integration tests

phernandez:
## Summary

Adds integration coverage for **concurrent** `write_note` MCP operations — a gap the suite had despite the write path being guarded by the FileService semaphore and entity-service race handling. All 8 tests drive the full real stack (MCP Client → MCP Server → FastAPI → database), use `output_format="json"` for structured assertions, and are deterministic: no sleeps except a bounded convergence poll in the search test.

Coverage in `test-int/mcp/test_concurrent_write_integration.py`:

- **Sequential collision oracle** — same title + directory blocks the second write as a structured `conflict` (`NOTE_ALREADY_EXISTS`); first body survives untouched
- **Permalink suffix recovery** — two notes claiming the same explicit `permalink:` recover via a deterministic `-1` suffix, both independently readable
- **Concurrent same-title race** — 8 identical-title writes at once: every result is `created`/`conflict`, no permalink forks, and the surviving note holds exactly one complete writer body
- **Concurrent distinct writes** — parallel writes across directories all succeed and read back with their own content
- **Shared-directory contention** — 12 concurrent writes into one folder yield unique permalinks
- **Write-then-search convergence** — 8 concurrent writes, then a bounded poll until *every* note is findable via its unique FTS token (fails listing missing titles if the bound is exhausted)
- **Read/write interleaving** — anchor-note reads gathered *concurrently with* other writes stay coherent
- **High-volume stress** (`@pytest.mark.slow`) — 20 concurrent writes, all created and read back intact

## Supersedes #1183

This PR supersedes #1183 and carries **@FBISiri's three commits with authorship preserved** (cherry-picked onto current `main`). The final review round is applied on top in one commit:

- **Canonical-content validation** (Codex thread): the same-title race now asserts the read-back body matches one complete writer body, not just title/permalink — a torn or interleaved write would match no writer
- **Substring termination** (Codex thread): the high-volume assertion matches `"Volume body {index}."` so index 1 can no longer false-positive against `"Volume body 10."`–`"Volume body 19."`
- **Convergence oracle restored**: the search test's weakened `found > 0` assertion is replaced with a bounded poll (up to 20 attempts, 0.25s apart) that requires *all* titles to be findable. Local `write_note` indexes synchronously before returning, so the first pass normally passes — the poll is convergence safety, not a timing hack
- **Stale docstring cross-reference** fixed (`test_concurrent_permalink_conflict_recovery` → `test_concurrent_same_title_collision`)
- **Out-of-scope hermes hunk dropped**: `integrations/hermes/tests/test_integration.py` reverted to match `main` — the diff touches only the new test file

## Verification

- `uv run pytest test-int/mcp/test_concurrent_write_integration.py -q --no-cov` — **8 passed** on three consecutive runs (18.65s / 19.10s / 19.76s), confirming determinism
- `uv run pytest test-int/mcp/test_write_note_integration.py -q --no-cov` — 15 passed
- `ruff check` + `ruff format --check` — clean
- `uv run ty check` — clean on the touched file (only pre-existing optional-dependency `pymilvus` import diagnostics elsewhere, unrelated)
- `git diff origin/main --stat` — only `test-int/mcp/test_concurrent_write_integration.py` (+584)

Thanks @FBISiri for the original contribution and the review iterations.

https://claude.ai/code/session_01G4rbaeHJN3L7CREp5v38J9

### DIFF
--- a/integrations/hermes/tests/test_integration.py
+++ b/integrations/hermes/tests/test_integration.py
@@ -25,6 +25,7 @@ import subprocess
 import tempfile
 import time
 import uuid
+from typing import Any
 
 import pytest
 
@@ -108,7 +109,7 @@ def provider(bm, temp_bm_project, tmp_path):
     p.shutdown()
 
 
-def _parse_tool_result(raw):
+def _parse_tool_result(raw) -> dict[str, Any] | None:
     try:
         d = json.loads(raw)
     except Exception:
--- a/integrations/hermes/tests/test_integration.py
+++ b/integrations/hermes/tests/test_integration.py
@@ -25,7 +25,6 @@ import subprocess
 import tempfile
 import time
 import uuid
-from typing import Any
 
 import pytest
 
@@ -109,7 +108,7 @@ def provider(bm, temp_bm_project, tmp_path):
     p.shutdown()
 
 
-def _parse_tool_result(raw) -> dict[str, Any] | None:
+def _parse_tool_result(raw):
     try:
         d = json.loads(raw)
     except Exception:
--- a/test-int/mcp/test_concurrent_write_integration.py
+++ b/test-int/mcp/test_concurrent_write_integration.py
@@ -75,7 +75,7 @@ async def test_write_same_title_same_directory_collision(mcp_server, app, test_p
     (The lower-level ``EntityRepository._handle_permalink_conflict`` numeric-suffix
     recovery is NOT reachable on this sequential path — the API pre-check pre-empts
     it. It is exercised via the concurrent TOCTOU race in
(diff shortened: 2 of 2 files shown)


FBISiri:
Thanks for cherry-picking my commits and for the thorough review on #1183 — I learned a lot from your approach to the concurrent test structure. The authorship preservation was a nice touch. Looking forward to contributing more to basic-memory!

FBISiri:
Thanks for taking the time to cherry-pick and improve on the tests, @phernandez. The canonical-content validation and convergence oracle additions are solid — I learned a few things from your final round. Happy to see the suite land deterministically.

Looking forward to contributing more.

Issue #686, opened 2026-03-21
Progressive note loading — compact index with on-demand full content


bm-clawd:
## Idea
Inspired by OpenViking's L0/L1/L2 progressive loading and Hermes Agent's "hot memory vs cold recall" architecture.

## Problem
When agents work with BM notes, they currently load full note content. For large knowledge graphs, this means either:
- Stuffing too many tokens into context (expensive, slow)
- Loading too few notes and missing relevant context

## Proposal
Implement tiered note loading through the MCP tools:

**L0 — Index** (always cheap to inject)
- Title, type, tags, status
- Relation summary (connected entity names)
- ~50-100 tokens per note

**L1 — Observations** (medium cost)
- All observations with categories
- Key relations with types
- ~200-500 tokens per note

**L2 — Full content** (on demand)
- Complete markdown body
- All relations with context
- Attachments/references

The `search_notes` and `build_context` tools could return L0 by default, with a `depth` or `detail` parameter to request L1/L2. This keeps agent prompts compact while making the full graph accessible on demand.

## Benefits
- Dramatically fewer tokens for knowledge graph navigation
- Agents can scan more notes before deciding which to read fully
- Better prompt cache stability (compact index stays stable)
- Aligns with how humans browse — scan titles, then dive in

## References
- OpenViking L0/L1/L2: github.com/volcengine/OpenViking
- Hermes Agent memory layers (hot prompt memory + cold FTS recall)
- Relates to Momentum UI progressive rendering

digitalities:
I'm hitting MCP response size limits at 57 pages. `search_notes` with broad queries returns 300K-1.3M characters, exceeding the max allowed tokens. `build_context` across all pages for graph-wide analysis costs ~50K tokens and won't scale past ~200 pages. A compact response mode that returns only titles and relation targets (without full page content) would make graph-wide operations practical. Current workaround: sequential `build_context` per page + Obsidian CLI for structural queries.

TeaShaman-cyber:
In our experience, progressive loading is useful not only for reducing token usage, but also for separating evidence types.

A compact search or index response should be treated as discovery.
A full exact-address read should be treated as content verification.
Neither should silently imply current operational authority.

A useful implementation detail would be making the requested detail level explicit, for example:
- index or discovery;
- observations and relations;
- full note content.

It would also help if the response made clear whether the result came from the current indexed state or from a direct addressed read.

[...]

PR #1299, opened 2026-08-23
fix(mcp): bound context traversal requests

### LINKED ISSUE #1296: build_context has no page_size or max_related ceiling, while list_directory has had one since #1048
## The inconsistency

`list_directory` was bounded and paginated in #1048 (closed 2026-07-16), on the reasoning that an unbounded tool result *"can produce oversized MCP responses"*. It enforces a ceiling:

```python
if page_size > MAX_DIRECTORY_PAGE_SIZE:      # 200
    raise ValueError(f"page_size must be <= {MAX_DIRECTORY_PAGE_SIZE}, got {page_size}")
```

`build_context` was not included in that change. Its `page_size` and `max_related` are `Field(default=10)` with **no upper bound at all**, so a caller can raise them arbitrarily.

## What that looks like in practice

An agent asked to build a graph of a 149-note vault reached for `build_context` over a wildcard, and — finding the default of 10 gave it a slice rather than the graph — raised the parameters until it got everything. It settled on `page_size=500, max_related=1000`.

Measured at the reverse proxy, that returned **1,683,656 bytes**. The same call appears twice in the access log, byte-for-byte identical, three and a half minutes apart, because the turn it broke was retried and issued the same query again.

For scale, from the same vault in the same hour:

| tool | max response |
|---|---|
| `build_context` | **1.68 MB** |
| `list_directory` | 134 KB |
| `read_note` | 53 KB |
| `search_notes` | 4 KB |

Two things compound here, and neither is obvious from the tool's signature:

- **`max_related` multiplies.** It caps related rows *per result*, so raising it scales the response by the number of primaries. A caller reading "maximum number of related results" can reasonably read it as a total.
- **The defaults invite it.** `page_size=10` on a whole-vault question returns something that looks like a truncated answer with no clear signal about how much is missing, so raising the number is the natural next move. There is nothing to raise it *to*.

## Suggested fix, matching what #1048 already established

Ceilings on both, validated in the same shape, constants beside the schema as `MAX_DIRECTORY_PAGE_SIZE` is:

```python
MAX_CONTEXT_PAGE_SIZE = 200      # the number already proven on list_directory
MAX_MAX_RELATED = 500            # higher: an edge is hundreds of bytes, a note body thousands
```

**Refusing rather than truncating** seems right for this tool specifically: a graph silently cut to fit is a wrong answer that looks complete, and the caller has no way to tell which edges are missing. An explicit error lets them page.

It also helps if the message points somewhere. Ours reads:

> `page_size must be <= 200, got 500. For a whole-vault read, page through with include_content=False — the note bodies are most of the payload and the graph is not in them.`

*(`include_content` is a local addition on our fork — the equivalent upstream idea is #686's L0 tier. The general point stands without it: an error that only refuses leaves the caller guessing, and guessing is how they arrived at 500.)*

## Relationship to #686

#686 proposes L0/L1/L2 progressive loading across `search_notes` and `build_context`, and an April comment there reports the same symptom — *"`build_context` across all pages for graph-wide analysis costs ~50K tokens and won't scale past ~200 pages."*

This issue is narrower and independent. #686 is about **what** a response contains; this is about **how much** of it a caller can request. A ceiling is useful whether or not the tiering lands, and it is a much smaller change.

## Happy to send a patch

We have this implemented and tested on a fork — the constants, the validation, and four tests including one asserting that values *at* the cap still work, since a bound that refuses legitimate calls is worse than no bound. Glad to open a PR against `main` if the approach looks right; wanted to check the shape with you first rather than arrive with a design.

phernandez:
## Why

- `build_context` accepted arbitrary `page_size` and `max_related` values, allowing callers to turn a traversal tool into a multi-megabyte vault export.
- Large requests increase database work, response serialization, transport size, and model-context consumption.
- Fixes #1296 for the v0.23 release.

## What Changed

- Caps primary context pages at 50 results.
- Caps related context at 100 total results across the page.
- Rejects negative `max_related` values, which otherwise act like an unbounded SQL limit.
- Enforces the contract at both the MCP tool and direct API boundary.
- Clarifies that callers should paginate or follow returned `memory://` links for continued traversal.

## Implementation Details

- Shared limits live with the memory context schemas so the MCP tool, typed client, and API use the same defaults and ceilings.
- The MCP tool returns an actionable error rather than silently truncating a graph.
- The separate `recent_activity` contract remains unchanged.
- These are count guardrails, not a hard byte limit; progressive content loading remains separate follow-up work in #686.

### Automated

- `uv run pytest tests/api/v2/test_memory_router.py test-int/bughunt_fixes/test_navigation_pagination_integration.py -q`: 21 passed.
- `uv run ruff check src/basic_memory/schemas/memory.py src/basic_memory/mcp/tools/build_context.py src/basic_memory/api/v2/routers/memory_router.py src/basic_memory/mcp/clients/memory.py tests/api/v2/test_memory_router.py test-int/bughunt_fixes/test_navigation_pagination_integration.py`: passed.
- `just typecheck`: passed.

### Manual

- Reviewed the API and MCP contracts together to confirm oversized values are rejected and values exactly at each ceiling remain valid.

## Risks / Follow-ups

- Individual note bodies can still be large, so the limits do not promise a fixed response byte size.
- #686 remains the appropriate place for progressive content tiers.

### DIFF
--- a/src/basic_memory/api/v2/routers/memory_router.py
+++ b/src/basic_memory/api/v2/routers/memory_router.py
@@ -18,7 +18,11 @@ from basic_memory.deps import (
 )
 from basic_memory.schemas.base import TimeFrame, parse_timeframe
 from basic_memory.schemas.memory import (
+    DEFAULT_CONTEXT_PAGE_SIZE,
+    DEFAULT_CONTEXT_RELATED_RESULTS,
     GraphContext,
+    MAX_CONTEXT_PAGE_SIZE,
+    MAX_CONTEXT_RELATED_RESULTS,
     normalize_memory_url,
 )
 from basic_memory.schemas.search import SearchItemType
@@ -125,9 +129,13 @@ async def get_memory_context(
     project_id: str = Path(..., description="Project external UUID"),
     depth: int = 1,
     timeframe: Optional[TimeFrame] = None,
-    page: int = 1,
-    page_size: int = 10,
-    max_related: int = 10,
+    page: int = Query(1, ge=1),
+    page_size: int = Query(DEFAULT_CONTEXT_PAGE_SIZE, ge=1, le=MAX_CONTEXT_PAGE_SIZE),
+    max_related: int = Query(
+        DEFAULT_CONTEXT_RELATED_RESULTS,
+        ge=0,
+        le=MAX_CONTEXT_RELATED_RESULTS,
+    ),
 ) -> GraphContext:
     """Get rich context from memory:// URI.
 
@@ -143,8 +151,8 @@ async def get_memory_context(
         depth: How many levels of related entities to include
         timeframe: Optional time window for filtering related content
         page: Page number for pagination
-        page_size: Number of items per page
-        max_related: Maximum related entities to include
+        page_size: Number of primary items per page
+        max_related: Maximum total related entities to include
 
     Returns:
         GraphContext with the entity and its related context
--- a/src/basic_memory/mcp/clients/memory.py
+++ b/src/basic_memory/mcp/clients/memory.py
@@ -12,7 +12,11 @@ import logfire
 # call_* helpers live in basic_memory.mcp.tools.utils; importing that at module
 # level executes the whole tools package (fastmcp + mcp SDK) during CLI startup,
 # so each method defers the import to call time instead (#886).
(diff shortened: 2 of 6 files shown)


phernandez:
@codex review

PR #1306, opened 2026-08-23
feat: add #bm:links_to directive for ambiguous relations

### LINKED ISSUE #1301: Add a #bm:links_to directive for ambiguous relation-shaped bullets
## Problem

Basic Memory intentionally treats a single token before a wikilink as an explicit relation type:

```markdown
- Mother [[Target]]
```

That grammar is valid and cannot be distinguished reliably from a short prose label. Capitalization or vocabulary heuristics would be arbitrary and backward-incompatible.

PR #1291 fixed the actionable prose-tail bug from #1260: a line such as `- Added [[Target]] to the roster` now falls back to implicit `links_to` handling without dropping content or additional wikilinks. Authors still need an explicit escape hatch for the genuinely ambiguous no-tail form.

## Proposed syntax

Reserve an exact terminal directive:

```markdown
- Mother [[Target]] #bm:links_to
```

The directive means: preserve the line and its wikilinks, but treat every wikilink on this line as an implicit `links_to` edge rather than parsing the leading token as a relation type.

## Expected behavior

- Recognize only the exact `#bm:links_to` directive at the end of the line, allowing trailing whitespace.
- Emit `links_to` for every wikilink on the line.
- Do not expose the directive as a note tag, observation text, relation context, or indexed content.
- Preserve the directive in the canonical Markdown source.
- Leave explicit relations without the directive unchanged.
- Do not introduce capitalization, language, or relation-vocabulary heuristics.

## Examples

```markdown
- Mother [[Alice]] #bm:links_to
- Mentions [[Alice]] and [[Bob]] #bm:links_to
- spouse_of [[Alice]]
```

The first line creates `links_to [[Alice]]`; the second creates `links_to` edges to both targets; the third remains an explicit `spouse_of` relation.

## Acceptance tests

- Ambiguous single-token prefix plus the directive falls back to `links_to`.
- Multiple wikilinks on a directed line are all preserved as `links_to`.
- The directive does not become a hashtag/tag or indexed text.
- A normal explicit relation still retains its authored relation type.
- Similar text that is not the exact terminal directive has no special behavior.

Follow-up to #1260 and #1291.

mikemikimike:
## Summary

- Fixes #1301.
- Add the exact terminal `#bm:links_to` directive for ambiguous relation-shaped bullets.
- Preserve authored Markdown while excluding the directive from observation tags/content.

## Implementation

The markdown relation plugin strips only the exact terminal directive before inline wikilink parsing. Such lines emit `links_to` for every wikilink; explicit relation syntax and non-terminal lookalikes remain unchanged.

## Tests

- `python -m pytest -p no:cov -o addopts='' --confcutdir=tests/markdown -q tests/markdown/test_relation_edge_cases.py tests/markdown/test_markdown_plugins.py` (24 passed)
- `python -m ruff check src/basic_memory/markdown/plugins.py tests/markdown/test_relation_edge_cases.py` (passed)

The repository-wide test suite and full `just check` were not run because the local dependency bootstrap was incomplete and the GitHub transport was unstable; the focused parser tests and Ruff check ran successfully.


phernandez:
hi @mikemikimike thanks for the PR. Would you mind doing 

```
git rebase HEAD~3 --signoff
```
and force pushing to the branch again? We require commits to be signed off (-s) for our DCO.

phernandez:
Closing as landed: the feature is on main as 29a98acd / e05e62fe / 295124b5 with your commits and authorship intact — it was bundled into the v0.23.1 fast-follow branch before this PR could be merged directly (the only delta vs this branch is a ruff line-wrap in one test assert). `#bm:links_to` ships in v0.23.1, releasing today, credited to you. Thanks for the clean design — directive stripped from indexed content while preserved in source is exactly right.

PR #1318, opened 2026-08-24
fix(cli): preload rich's Unicode width table before in-place upgrade

### LINKED ISSUE #1316: bm update: crash after in-place upgrade — preload misses rich's Unicode width table
## Summary

`bm update` crashes with an unhandled `ModuleNotFoundError` when it prints its result message after an in-place Homebrew upgrade has already replaced the running installation on disk.

`_preload_lazy_console_modules()` (added in 497a4e0a, shipped in v0.22.1 / v0.23.0) was written for exactly this failure mode, but it preloads only `rich._emoji_codes` and `typer.rich_utils`. It misses rich's **lazily loaded Unicode cell-width table** (`rich._unicode_data.unicode<version>`), which is imported on the first `console.print()` that has to measure a non-ASCII character. That import happens *after* the upgrade subprocess has deleted the old prefix, so it fails.

**This is still reproducible against `main` (8c5410cd, v0.23.0)** — see the standalone repro below. The reported crash happened on 0.21.5, which predates the preload fix entirely, but the current fix does not close it.

## What happened

```
[drews-m5][~]$ bm update
Basic Memory is up to date (0.21.5).
[drews-m5][~]$ bm update
╭─────────────── Traceback (most recent call last) ────────────────╮
│ in update:25                                                     │
│ in print:1731                                                    │
│ in render:1339                                                   │
│ in __rich_console__:696                                          │
│ in wrap:1239                                                     │
│ in divide_line:45                                                │
│ in cell_len:109                                                  │
│ in cached_cell_len:95                                            │
│ in _cell_len:131                                                 │
│ in <genexpr>:132                                                 │
│ in get_character_cell_size:60                                    │
│ in load:90                                                       │
│ in import_module:88                                              │
│ in _gcd_import:1406                                              │
│ in _find_and_load:1371                                           │
│ in _find_and_load_unlocked:1335                                  │
╰──────────────────────────────────────────────────────────────────╯
ModuleNotFoundError: No module named 'rich._unicode_data.unicode17-0-0'
[1][drews-m5][~]$ bm status
╭────────────────────────╮
│ main: Project Index    │
│ └── 388 observed files │
╰────────────────────────╯
[drews-m5][~]$ bm --version
Basic Memory version: 0.23.0
```

The upgrade itself landed — `bm --version` reports 0.23.0 afterwards and `/opt/homebrew/Cellar/basic-memory/` now contains only `0.23.0`. Only the exit path crashed, with a traceback in place of the status message and exit code 1.

**Environment:** macOS (Apple Silicon), Homebrew install, upgrading 0.21.5 → 0.23.0. New env: Python 3.14.7, `unicodedata.unidata_version` 16.0.0.

## Root cause

Reading the frames bottom-up:

1. `run_auto_update()` runs `brew upgrade basic-memory` as a subprocess (`auto_update.py:380`). Homebrew removes `/opt/homebrew/Cellar/basic-memory/0.21.5` — the prefix the running interpreter is executing from. Already-imported modules stay resident in memory; **anything not yet imported is now unreachable.**
2. Control returns to `commands/update.py`, which calls `console.print(...)` on the result.
3. rich wraps the line → `cell_len` → `get_character_cell_size` → `rich._unicode_data.load()`, which lazily `import_module`s the width table on first use.
4. The file is gone → `ModuleNotFoundError`, unhandled, pretty-printed by typer's rich excepthook (which survived only because `typer.rich_utils` was already resident).

Two details worth recording, because both are counter-intuitive:

- **`unicode17-0-0` is not a Python/Unicode version mismatch.** `rich._unicode_data.load()` defaults to `"latest"` (`VERSIONS[-1]` = 17.0.0) rather than the interpreter's `unidata_version`, so the requested module name is 17-0-0 even on a py3.14 interpreter reporting Unicode 16.0.0. The file exists in the installed rich; it was simply no longer on disk at the moment of import.
- **The import only fires for non-ASCII text.** `_cell_len` has an ASCII fast path, so `"Basic Memory is up to date (0.23.0)."` never touches the table. The crash needs a message containing a non-ASCII character — which is exactly what the `FAILED` branch prints, since it echoes captured `brew` output (curly quotes, em dashes, `⚠`).

Frame `update:25` is the `AutoUpdateStatus.FAILED` branch in both 0.21.5 and current `main`:

```python
if result.status == AutoUpdateStatus.FAILED:
    detail = f" {result.error}" if result.error else ""
    console.print(f"[red]{result.message or 'Update failed.'}{detail}[/red]")   # <-- line 25
```

So `brew upgrade` exited non-zero even though the upgrade landed, and the long, non-ASCII stderr it captured is what forced the line wrap that triggered the table load. That is worth a look on its own (see "Secondary observations"), but the crash is not conditional on it — any wrapped non-ASCII status message after an in-place upgrade hits it, including the `UPDATED` message on line 29 if it ever wraps with non-ASCII content.

## Repro (no Homebrew needed, runs against current `main`)

Blocking every not-yet-imported module is a faithful stand-in for a deleted prefix:

```python
"""Repro: bm update crashes after an in-place upgrade removes the old install."""
import sys
from basic_memory.cli.auto_update import _preload_lazy_console_modules
from rich.console import Console


class UpgradedAwayFinder:
    """Every not-yet-imported module is gone from disk."""

    def find_spec(self, fullname, path=None, target=None):
        raise ModuleNotFoundError(f"No module named {fullname!r}")


_preload_lazy_console_modules()                  # what the current fix preloads
sys.meta_path.insert(0, UpgradedAwayFinder())    # <-- `brew upgrade` happens here

console = Console(width=80)
# brew's stderr, echoed by the FAILED path: long + non-ASCII
console.print("[red]Automatic update failed. Error: could not link ‘basic-memory’ "
              "— the formula's files were replaced while running. " + "detail " * 12 + "[/red]")
print("NO CRASH")
```

Result on v0.23.0 — the reported error, verbatim:

```
ModuleNotFoundError: No module named 'rich._unicode_data.unicode17-0-0'
```

Confirming the preload gap directly:

```python
>>> _preload_lazy_console_modules()
>>> [m for m in sys.modules if "_unicode_data.unicode" in m]
[]                                        # table not loaded
>>> console.print("… non-ASCII text that wraps …")
>>> [m for m in sys.modules if "_unicode_data.unicode" in m]
['rich._unicode_data.unicode17-0-0']      # loaded at print time, too late
```

## Suggested fix

Force the width table to load inside `_preload_lazy_console_modules()`, alongside the existing preloads. Importing the module by name is not enough to be future-proof — better to go through rich's own resolution so the correct version is chosen:

```python
def _preload_lazy_console_modules() -> None:
    import rich._emoji_codes  # noqa: F401
    import typer.rich_utils  # noqa: F401

    # Trigger: rich defers the Unicode cell-width table until the first print
    # that measures a non-ASCII character (`_cell_len` has an ASCII fast path).
    # Why: the FAILED/UPDATED messages echo brew output containing non-ASCII
    # characters, so the import lands after the upgrade removed our files.
    # Outcome: resolve and load the table now, while the files still exist.
    from rich.cells import cell_len

    cell_len("─‘✓")
```

Verified: with this in place, the repro above prints the wrapped message and reaches `NO CRASH`.

Worth considering as defense in depth, since this is the second lazy-import to be found this way and rich/typer can add more at any time:

- Wrap the post-upgrade print in `try/except Exception` and fall back to a plain `print()` — the update already succeeded at that point, so a status line must never be able to fail the command.
- Or hand off: `os.execv` into the freshly installed executable, or defer the message to a `subprocess` run of the new binary, so no post-upgrade code runs from a deleted prefix at all.

The first is cheap and closes the whole class rather than one instance.

## Secondary observations

Neither is the ask here — filing them as context in case they are worth their own issues:

1. **`brew upgrade` exited non-zero on a successful upgrade.** Frame `update:25` is the `FAILED` branch, yet 0.23.0 installed cleanly and the 0.21.5 Cellar directory was removed. Users get "Automatic update failed" plus exit code 1 for an upgrade that worked. A post-upgrade version re-check before classifying the result would make this self-correcting.
2. **First run reported "up to date (0.21.5)" with 0.23.0 available**, then the very next run found the update. 0.21.5 predates 13e81ef7 / c5cff99c, so this is plausibly already fixed — noting it only because it is visible in the transcript above.

groksrc:
Fixes #1316.

## The bug

`bm update` crashed with an unhandled traceback after Homebrew had already replaced the running installation:

```
ModuleNotFoundError: No module named 'rich._unicode_data.unicode17-0-0'
```

`_preload_lazy_console_modules()` (497a4e0a) was written for exactly this failure mode, but it only preloads `rich._emoji_codes` and `typer.rich_utils`. rich *also* defers its Unicode cell-width table until the first character it cannot measure with the ASCII fast path in `_cell_len` — so the import lands after `brew upgrade` removed the prefix we are executing from, and the message meant to report the result becomes a traceback plus exit code 1 for an upgrade that had already succeeded.

Reported on 0.21.5 (which predates the preload entirely), but **still reproducible on `main`** — the added test fails without this change with that exact error.

Two details worth keeping in mind while reviewing, both counter-intuitive:

- **`unicode17-0-0` is not a Python/Unicode mismatch.** `rich._unicode_data.load()` defaults to `"latest"` (`VERSIONS[-1]`), not the interpreter's `unidata_version`, so it asks for 17.0.0 even on a py3.14 interpreter reporting Unicode 16.0.0. The file ships in rich; it just was not on disk anymore.
- **It only fires for non-ASCII text.** `"Basic Memory is up to date (0.23.0)."` never touches the table. The `FAILED` branch echoing captured `brew` stderr — curly quotes, em dashes, `⚠` — does.

## Changes

**1. Preload the width table (the actual fix)**

```python
from rich.cells import cell_len
cell_len("─‘✓")
```

Going through `cell_len` rather than importing `rich._unicode_data.unicode<version>` by name means rich picks the table version itself and honors `UNICODE_VERSION` exactly as the print path does — no hard-coded version to drift, and it works across the `rich>=13.9.4` range in `pyproject.toml` (older rich has no `_unicode_data` package at all).

**2. Make the status line unable to fail the command (defense in depth)**

New `print_update_status()` falls back to a plain `print()` if the console raises, and all update status lines in `commands/update.py` and `maybe_run_periodic_auto_update` now go through it.

This is the second deferred import found this way, and rich/typer are free to add more. A preload list can only ever cover what we know about today; by the time these lines print, the upgrade has already succeeded, so a status line should never be what turns it into a traceback. Happy to drop this half if you would rather keep the change minimal.

## Tests

- `test_status_message_survives_upgraded_away_install` — simulates the deleted prefix with a `meta_path` finder that raises for any not-yet-imported module, then prints a long non-ASCII line (long and non-ASCII on purpose: that is what forces the wrap that reaches for the table). It calls `console.print` directly rather than `print_update_status`, so the fallback cannot mask a preload regression.
- `_cool_deferred_width_table()` — clears rich's caches and unloads the table module first. Without this the test passes whether or not the fix is present, since earlier tests in the session warm the cache. Verified by neutralizing only the `cell_len` line: the test then fails with the reported `ModuleNotFoundError`, and passes with it restored.
- `test_print_update_status_falls_back_to_plain_output` — covers the fallback path.

`tests/cli/test_auto_update.py` passes (33), `ruff check`/`format` clean, `ty check` shows no new diagnostics (the 4 `pymilvus` unresolved-import errors are present on `main` too).

## Not addressed here

Noted in #1316 as separate observations: `brew upgrade` appears to have exited non-zero on an upgrade that landed (frame `update:25` is the `FAILED` branch, yet 0.23.0 installed and the old Cellar directory was removed), which suggests a post-upgrade version re-check before classifying the result would be worthwhile.

### DIFF
--- a/src/basic_memory/cli/auto_update.py
+++ b/src/basic_memory/cli/auto_update.py
@@ -221,6 +221,38 @@ def _preload_lazy_console_modules() -> None:
     """
     import rich._emoji_codes  # noqa: F401
     import typer.rich_utils  # noqa: F401
+    from rich.cells import cell_len
+
+    # Trigger: rich defers its Unicode cell-width table (`rich._unicode_data.
+    # unicode<version>`) until the first character it cannot measure with the
+    # ASCII fast path in `_cell_len`.
+    # Why: status messages echo captured `brew`/`uv` output, which carries
+    # non-ASCII characters (curly quotes, em dashes, warning glyphs), so the
+    # deferred import lands after the upgrade removed our files. Importing the
+    # module by name would hard-code a table version; calling `cell_len` uses
+    # rich's own resolution and honors UNICODE_VERSION like the print path does.
+    # Outcome: the table rich will reach for is resolved and cached up front.
+    cell_len("\u2500\u2018\u2713")
+
+
+def print_update_status(console: Console, text: str, style: str) -> None:
+    """Print an update status line that cannot fail the command.
+
+    Trigger: the line is printed after an in-place upgrade may already have
+    replaced this installation on disk.
+    Why: `_preload_lazy_console_modules` can only preload the deferred imports
+    we know about today, and rich/typer are free to add more. By the time this
+    prints, the upgrade has already succeeded -- a status line must never be
+    what turns it into a traceback and a non-zero exit.
+    Outcome: fall back to a plain, unstyled write that needs no new imports.
+    """
+    try:
+        console.print(f"[{style}]{text}[/{style}]")
+    except Exception as exc:
+        logger.warning(
+            f"Rich console print failed after update, falling back to plain output: {exc}"
+        )
+        print(text)
 
 
 def _save_last_checked_timestamp(config_manager: ConfigManager, checked_at: datetime) -> None:
@@ -468,11 +500,11 @@ def maybe_run_periodic_auto_update(
     }:
         out = console or Console()
         if result.status == AutoUpdateStatus.UPDATED:
-            out.print(f"[green]{result.message}[/green]")
+            print_update_status(out, f"{result.message}", "green")
         elif result.status == AutoUpdateStatus.FAILED:
             error_detail = f" {result.error}" if result.error else ""
-            out.print(f"[yellow]{result.message}{error_detail}[/yellow]")
+            print_update_status(out, f"{result.message}{error_detail}", "yellow")
         elif result.message:
-            out.print(f"[cyan]{result.message}[/cyan]")
+            print_update_status(out, f"{result.message}", "cyan")
 
     return result
--- a/src/basic_memory/cli/commands/update.py
+++ b/src/basic_memory/cli/commands/update.py
@@ -4,7 +4,7 @@ import typer
 from rich.console import Console
 
 from basic_memory.cli.app import app
-from basic_memory.cli.auto_update import AutoUpdateStatus, run_auto_update
+from basic_memory.cli.auto_update import AutoUpdateStatus, print_update_status, run_auto_update
 
 console = Console()
 
@@ -22,19 +22,21 @@ def update(
 
     if result.status == AutoUpdateStatus.FAILED:
         detail = f" {result.error}" if result.error else ""
-        console.print(f"[red]{result.message or 'Update failed.'}{detail}[/red]")
+        print_update_status(console, f"{result.message or 'Update failed.'}{detail}", "red")
         raise typer.Exit(1)
 
     if result.status == AutoUpdateStatus.UPDATED:
-        console.print(f"[green]{result.message or 'Basic Memory updated successfully.'}[/green]")
+        print_update_status(
+            console, f"{result.message or 'Basic Memory updated successfully.'}", "green"
+        )
         return
 
     if result.status == AutoUpdateStatus.UP_TO_DATE:
-        console.print(f"[green]{result.message or 'Basic Memory is up to date.'}[/green]")
+        print_update_status(console, f"{result.message or 'Basic Memory is up to date.'}", "green")
         return
 
     if result.status == AutoUpdateStatus.UPDATE_AVAILABLE:
-        console.print(f"[cyan]{result.message or 'Update available.'}[/cyan]")
+        print_update_status(console, f"{result.message or 'Update available.'}", "cyan")
         return
 
-    console.print(f"[dim]{result.message or 'No update action was performed.'}[/dim]")
+    print_update_status(console, f"{result.message or 'No update action was performed.'}", "dim")
--- a/tests/cli/test_auto_update.py
+++ b/tests/cli/test_auto_update.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import re
 import subprocess
 import sys
 import urllib.error
@@ -21,6 +22,7 @@ from basic_memory.cli.auto_update import (
     _check_homebrew_update_available,
     _is_interactive_session,
     _preload_lazy_console_modules,
+    print_update_status,
     detect_install_source,
     maybe_run_periodic_auto_update,
     run_auto_update,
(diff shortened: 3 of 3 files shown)


phernandez:
Closing as landed: this fix is on main as 07511391 with your commit and authorship intact — it was bundled into the v0.23.1 fast-follow branch before this PR could be merged directly. It ships in v0.23.1 (releasing today) and is credited to you in the release. Thanks for the excellent diagnosis and fix — the meta_path-based regression test in particular. Sorry for the awkward duplicate-landing path.

Issue #1333, opened 2026-08-25
[BUG] `bm schema validate <type>` prints its result then never exits — interpreter shutdown blocks on a lingering thread


scarabone:
## Bug Description

`bm schema validate <note_type>` never returns on the CLI. The server answers the underlying request in about 0.2 s (visible in `basic-memory.log`), and the CLI prints the Rich table header — then stalls indefinitely with the table half-drawn. The equivalent tool-path command, `bm tool schema-validate --project main`, validates every schema-covered note (499 notes) and returns JSON in 4 s, so this is specific to the `bm schema validate` command's output path, not to validation itself.

First seen on 0.22.1 (left running >5 min, never finished). Unchanged on 0.23.2.

## Steps To Reproduce

1. Install version `0.23.2`; a local project with a `schemas/` folder holding a `type: schema` note (here `OpenItem`, 48 notes of `type: open_item`).
2. Run:

```sh
bm schema validate open_item --project main
```

3. Wait. Killed after 90 s in the run below; on 0.22.1 it was still running after 5 min.

## Expected Behavior

The table renders and the command exits, in roughly the time the server takes to answer.

## Actual Behavior

Output stops mid-table and the process never exits:

```
                          Schema Validation: open_item
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┓
┃ Note                                            ┃ Status ┃ Warnings ┃ Errors ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━┓
(nothing further; killed at 90 s)
```

Server log for the same run — the work itself took 190 ms:

```
19:20:08.202 | INFO | basic_memory.mcp.tools.schema:schema_validate:306 - MCP tool call tool=schema_validate project=main note_type=open_item identifier=None
19:20:08.392 | INFO | basic_memory.mcp.tools.schema:schema_validate:320 - MCP tool response: tool=schema_validate project=main total=48 valid=48 warnings=19 errors=0
```

Control, same install, same project:

```sh
$ time bm tool schema-validate --project main   # all types, 499 notes
{ "total_notes": 499, "total_entities": 499, "valid_count": 499, "warning_count": 655, "error_count": 0, ... }
4 s
```

[...]

PR #1346, opened 2026-08-28
fix: resolve app config on the event loop so the CLI can exit

### LINKED ISSUE #1345: `basic-memory schema validate` prints its result then hangs instead of exiting (Python 3.13)
### What happened

`basic-memory schema validate` does all of its work, prints its result, and then
never exits. The process has to be killed.

```
$ basic-memory schema validate --project <name> --json --local
{
  "error": "No schemas defined in this project"
}
                                     # ...and then it hangs, indefinitely
```

The output is complete and correct — only the exit is stuck.

### What I expected

The command prints its result and exits 0.

### Reproducing it

This reproduces reliably in a **clean, isolated environment**: a fresh
`uv`-installed 0.23.2 with its own `HOME`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`
and `BASIC_MEMORY_CONFIG_DIR`, on CPython 3.13.15 (macOS, Apple Silicon).

A/B in that environment, toggling only the `def`/`async def` keyword on
`get_app_config` and changing nothing else:

```
WITHOUT FIX (sync get_app_config):   HUNG after 90s
WITH FIX    (async get_app_config):  exit=0 in 8.24s
```

**A caveat that cost me time, and may cost yours:** I could *not* reproduce this
when running against my normal, already-warm user environment — there the same
command exits in ~2.5s either way. If it does not reproduce for you on the first
try, use a clean isolated config/home rather than concluding it is not there.

### Cause, as far as I can tell

`get_app_config` in `src/basic_memory/deps/config.py` is a **sync** FastAPI
dependency. FastAPI resolves sync dependencies in AnyIO's worker-thread pool,
and those workers are non-daemon and stay parked on `queue.Queue.get` after the
request so they can be reused.

The CLI drives the app under `asyncio.run` through `httpx.ASGITransport`, with
no blocking portal to join that pool on the way out. So `asyncio.run` returns
with a live non-daemon thread, and `threading._shutdown` waits on a thread
nothing will ever wake.

`run_with_cleanup` in `cli/commands/command_utils.py` drains pending
materializations, background tasks and the DB, but it does not join the AnyIO
worker pool, so it does not cover this.

Directly observable, independent of the CLI — a minimal FastAPI app with one
route depending on `AppConfigDep`, driven over `ASGITransport`:

```
WITH FIX:    workers before = 0 / status 200 / workers after = 0 []
WITHOUT FIX: workers before = 0 / status 200 / workers after = 1 ['AnyIO worker thread']
```

This is not limited to `schema validate` — any CLI or MCP command that drives
the API through the in-process ASGI transport and touches `AppConfigDep` is
exposed. `schema validate` is just where it is easiest to see, because that
command does little else.

### Environment

- Basic Memory v0.23.2 (`c0bd87c6d5a4a58034b1d6c8c5018e443b0bd048`)
- CPython 3.13.15, macOS (Apple Silicon)

### Happy to send a PR

I have a one-line fix and a regression test ready.

WebKingdom:
## Problem

`basic-memory schema validate` prints its result and then never exits. The
output is complete and correct; only the exit is stuck. Fixes #1345.

Reproduced on v0.23.2, CPython 3.13.15 (macOS, Apple Silicon), in a clean
isolated environment (fresh `HOME`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`,
`BASIC_MEMORY_CONFIG_DIR`):

```
$ basic-memory schema validate --project brain --json --local
WITHOUT FIX (sync get_app_config):   HUNG after 90s   (result printed, process never exits)
WITH FIX    (async get_app_config):  exit=0 in 8.24s
```

Same environment, same command, one keyword toggled and nothing else.

Note for reviewers: this does **not** reproduce against a warm, default user
environment — there the command exits in ~2.5s either way. A clean isolated
config/home is needed to see it.

## Root cause

`src/basic_memory/deps/config.py` declared `get_app_config` as a **sync**
FastAPI dependency:

```python
def get_app_config(request: Request) -> BasicMemoryConfig: ...
```

FastAPI runs sync dependencies in AnyIO's worker-thread pool. Those worker
threads are **non-daemon**, and AnyIO deliberately keeps them alive after the
task finishes — parked on `queue.Queue.get` — so they can be reused.

The CLI drives the app under `asyncio.run` via `httpx.ASGITransport`, with no
surrounding AnyIO blocking portal to join the pool on the way out. So
`asyncio.run` returns while a non-daemon worker is still parked, the interpreter
reaches `threading._shutdown`, and it blocks forever on a thread that nothing
will ever wake.

`run_with_cleanup` in `cli/commands/command_utils.py` drains pending
materializations, background tasks and the DB, but does not join the AnyIO
worker pool, so it does not cover this case.

The function body does no I/O — it reads `request.app.state.container` or calls
`resolve_container()` — so there is nothing that needs a worker thread.

## Fix

Declare the provider `async def`. Resolution stays on the event loop and no
worker thread is ever dispatched. One-line change; the body is untouched.

## Tests

Adds `tests/test_app_config_dependency_resolution.py`, which builds a minimal
FastAPI app with one route depending on `AppConfigDep`, drives it over
`ASGITransport`, and asserts no AnyIO worker thread survives the request:

```
WITH FIX:    workers before = 0 / status 200 / workers after = 0 []
WITHOUT FIX: workers before = 0 / status 200 / workers after = 1 ['AnyIO worker thread']
```

Verified RED before GREEN — with `deps/config.py` reverted to `def`, the
behavioural test fails with `before=0, after=1`, not merely on the
`iscoroutinefunction` assertion.

The conftest updates (`tests/api/v2/`, `tests/cli/`, `tests/mcp/`, `tests/test_deps.py`)
are the mechanical `await`/async-override follow-through for the now-async provider.

## Full suite

Run against upstream's canonical unit target
(`BASIC_MEMORY_ENV=test pytest -p pytest_mock --no-cov tests`) on CPython 3.13.15:

```
v0.23.2 baseline:  38 failed, 4984 passed, 38 skipped
with this branch:  38 failed, 4986 passed, 38 skipped
```

The failure sets are **identical** — zero regressions; the +2 are the new tests.
The 38 pre-existing failures are present on a clean `v0.23.2` checkout in this
environment (Rich output-width and cloud-credential assertions) and are
unrelated to this change.

### DIFF
--- a/src/basic_memory/deps/config.py
+++ b/src/basic_memory/deps/config.py
@@ -12,12 +12,27 @@ from fastapi import Depends, Request
 from basic_memory.config import BasicMemoryConfig
 
 
-def get_app_config(request: Request) -> BasicMemoryConfig:
+async def get_app_config(request: Request) -> BasicMemoryConfig:
     """Resolve the application configuration from the composition root.
 
     API requests read the container the lifespan stored on ``app.state``.
     Requests served without a lifespan (the CLI/MCP local ASGI flow) fall back
     to the API composition root, which creates a container on demand.
+
+    This provider must stay ``async`` even though it never awaits. FastAPI runs
+    *sync* dependencies in AnyIO's worker-thread pool, and those workers are
+    non-daemon: after the request finishes AnyIO parks the thread on
+    ``queue.Queue.get`` so it can be reused. Every CLI/MCP command drives the app
+    through an in-process ASGI transport under ``asyncio.run``, which returns
+    without joining that pool, so the interpreter reaches
+    ``threading._shutdown`` with a live non-daemon thread that nothing will ever
+    wake -- and the process hangs instead of exiting (observed as
+    ``basic-memory schema validate`` never returning on Python 3.13; 3.14 reaps
+    the stranded worker itself and masks the bug).
+
+    Declaring it ``async`` keeps resolution on the event loop, so no worker
+    thread is dispatched and nothing survives the request. See
+    ``tests/test_app_config_dependency_resolution.py``.
     """
     container = getattr(request.app.state, "container", None)
     if container is not None:
--- a/tests/api/v2/conftest.py
+++ b/tests/api/v2/conftest.py
@@ -22,7 +22,14 @@ async def app(test_config, engine_factory, app_config) -> AsyncGenerator[FastAPI
     from basic_memory.api.app import app
 
     previous_overrides = dict(app.dependency_overrides)
-    app.dependency_overrides[get_app_config] = lambda: app_config
+
+    # Async override, matching the production provider: a sync override would
+    # be dispatched to AnyIO's non-daemon worker pool and leave a parked thread
+    # behind, which is exactly the shutdown hang the async provider avoids.
+    async def _app_config():
+        return app_config
+
+    app.dependency_overrides[get_app_config] = _app_config
     app.dependency_overrides[get_engine_factory] = lambda: engine_factory
     try:
         yield app
--- a/tests/cli/conftest.py
+++ b/tests/cli/conftest.py
@@ -44,7 +44,14 @@ async def app(
     """Create test FastAPI application."""
     app = fastapi_app
     previous_overrides = dict(app.dependency_overrides)
-    app.dependency_overrides[get_app_config] = lambda: app_config
+
+    # Async override, matching the production provider: a sync override would
+    # be dispatched to AnyIO's non-daemon worker pool and leave a parked thread
+    # behind, which is exactly the shutdown hang the async provider avoids.
+    async def _app_config():
+        return app_config
+
+    app.dependency_overrides[get_app_config] = _app_config
     app.dependency_overrides[get_engine_factory] = lambda: engine_factory
     try:
         yield app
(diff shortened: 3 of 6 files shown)


phernandez:
Thanks for this — the diagnosis of the parked non-daemon `AnyIO worker thread` was right, and the sync `get_app_config` dependency is indeed what put a worker in play. I reproduced it in a clean env on 3.13 and got the same thread dump.

One correction to the story, which changes the fix: the worker isn't "stranded because `asyncio.run` doesn't join the pool" — anyio does register `worker.stop` on the root task. It never ran because `alembic/env.py` applied `nest_asyncio` on Python `< 3.14`, and nest_asyncio's `run_until_complete` returns the moment the task is done without running its completion callbacks. That's also why 3.12 hangs (#1333) and 3.14 doesn't — 3.14 skipped nest_asyncio, it doesn't "reap" anything.

Since the CLA isn't signed here, I've reimplemented this independently in #1349: nest_asyncio is removed (root cause), and `get_app_config` / `get_read_cache` are made async as you proposed, with regression tests for both. You're credited in the commit and PR. Closing this one in favour of #1349 — thanks again for the careful repro.

PR #1347, opened 2026-08-28
fix(core): honor batch metadata concurrency

phernandez:
## Summary

- thread the caller-provided metadata update concurrency through the portable batch runtime
- apply it to the internal search-index refresh phase instead of always using the independent application default
- preserve the existing application default when callers do not provide an override

## Why

Basic Memory Cloud derives an indexing fan-out from the worker process tenant database budget. The runtime already accepted that value for note-content reconciliation and relation publication, but the earlier entity search refresh still used `index_metadata_update_max_concurrent`. Two Cloud batch jobs could therefore exceed the process-wide tenant connection budget on the observed `search.index_entity_data` path.

## Verification

- `uv run pytest tests/indexing/test_batch_indexer.py tests/indexing/test_index_batch_runtime.py -q` (26 passed)
- focused ruff and ty checks passed
- regression test observes peak in-flight search writes across two files and requires the explicit limit of 1

Companion dependency for basicmachines-co/basic-memory-cloud#1845.

### DIFF
--- a/src/basic_memory/indexing/batch_indexer.py
+++ b/src/basic_memory/indexing/batch_indexer.py
@@ -159,11 +159,14 @@ class BatchIndexer:
         *,
         max_concurrent: int,
         parse_max_concurrent: int | None = None,
+        metadata_update_max_concurrent: int | None = None,
         existing_permalink_by_path: dict[str, str | None] | None = None,
     ) -> IndexingBatchResult:
         """Index one batch of loaded files with bounded concurrency."""
         if max_concurrent <= 0:
             raise ValueError("max_concurrent must be greater than zero")
+        if metadata_update_max_concurrent is not None and metadata_update_max_concurrent <= 0:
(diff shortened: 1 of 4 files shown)


phernandez:
@jope-bm please review this Core seam alongside basic-memory-cloud#1845.

The Cloud worker resolves an effective DB fan-out of 1 for the current two-job/two-permit production shape. This PR makes that caller-provided value load-bearing in Core's internal metadata/search refresh instead of silently falling back to app concurrency. The regression test measures real simultaneous search-writer activity across a two-file batch and observes a peak of exactly 1.

All Core CI checks are green at `7e6daa92d62b1964bdb66d8dcd823e3b5830f685`; Cloud pins that exact commit. Please focus on whether the optional override preserves non-Cloud defaults and whether any other `IndexBatchRuntime` implementation needs the new parameter.

PR #1348, opened 2026-08-28
fix(core): ignore wikilinks in inline code

### LINKED ISSUE #1332: [BUG] Wikilinks inside inline code spans are parsed as relations (fenced blocks are correctly ignored)
## Bug Description

A `[[wikilink]]` inside an **inline code span** (backticks) is parsed as a relation. CommonMark treats code spans as literal text, and Basic Memory already respects **fenced** code blocks (a wikilink inside ```` ``` ```` does not mint an edge), but inline spans are not honoured — so prose that *quotes* a relation line as an example creates a real edge to the quoted target.

Practical impact: retiring a hub note. Two real `part_of [[Daily log]]` edges were removed, the inbound count would not drop to zero, and the survivor turned out to be a sentence in another note that documented the template line inside backticks purely as an example. The quoted mention kept the retired hub "alive" in the graph, and any orphan/hub-integrity metric built on `relation` rows sees it as a live link.

## Steps To Reproduce

1. Install version `0.23.2` (also reproduced on `0.22.1`).
2. Write a note whose only wikilinks are inside code:

```sh
bm tool write-note --title "Codespan Probe" --folder inbox --project main --content '# Codespan Probe

The template line is documented as `- part_of [[Codespan Inline Target]]` and must not create an edge.

```markdown
- relates_to [[Codespan Fenced Target]]
```

- [fact] only prose and code here'
```

3. Inspect the relations minted for that note:

```sql
SELECT r.relation_type, r.to_name, r.to_id
FROM relation r JOIN entity e ON e.id = r.from_id
WHERE e.title = 'Codespan Probe';
```

## Expected Behavior

No relation rows. Both targets are inside code — one inline span, one fenced block — and neither is a link in the CommonMark sense.

## Actual Behavior

```
relation_type|to_name|to_id
links_to|Codespan Inline Target|
```

The fenced-block link is correctly ignored; the inline-span link becomes a `links_to` relation (unresolved here only because the target does not exist — with an existing target it resolves to a real `to_id`, which is the hub-retirement case above).

## Environment

- OS: macOS 27.0 (Apple Silicon)
- Python version: 3.12.13
- Basic Memory version: 0.23.2 (same on 0.22.1)
- Installation method: `uv tool install basic-memory==0.23.2 --prerelease=allow`
- SQLite backend, local mode

## Possible Solution

Mask inline code spans (`` `...` ``, including double-backtick spans) before the wikilink / relation regexes run, the same way fenced blocks are already skipped. The parser only needs to treat backtick spans as opaque text; nothing else about link extraction changes.

phernandez:
Supersedes #1337 by @yaodong-shen — their commit is cherry-picked here unchanged, with authorship and sign-off preserved, so the full CI matrix runs (fork PRs only get DCO/CLA/CodeQL). Will be rebase-merged so the contributor commit lands on `main` as-is.

Fixes #1332.

## Summary

A `[[wikilink]]` inside an inline code span was parsed as a relation, so prose that *quotes* a relation line as an example minted a real graph edge (the reporter's retired hub note kept an inbound link from a backticked template example). Fenced blocks were already ignored; inline spans were not.

MarkdownIt already tokenizes `code_inline` before Basic Memory's relation rule runs. The fix records the exact source ranges MarkdownIt classifies as inline code (by wrapping its own `backtick` rule, so CommonMark delimiter/escape semantics and its linear-time unmatched-run cache are reused rather than re-implemented), then masks only the `[`/`]` characters inside those spans before relation detection. Masking is position-preserving, so targets and context are still sliced from the original source text — escapes, formatting and trailing code in relation context are retained.

## Review history on #1337

Codex raised five P2s on the original PR (source preservation, escaped backticks, trailing code in explicit relations, quadratic rescans, code-span brackets inside outer wikilink targets); the author addressed each with a follow-up push and regression cases. The final push (`7caf720`) never got a Codex re-review, so this PR will pick that up.

## Verification (local, this branch)

- `uv run pytest tests/markdown tests/services tests/importers -q --no-cov` — 555 passed, 2 skipped
- `uv run ruff check` / `ruff format --check` on the touched files — clean
- `uv run --all-extras ty check src tests test-int` — clean

### DIFF
--- a/src/basic_memory/markdown/plugins.py
+++ b/src/basic_memory/markdown/plugins.py
@@ -5,6 +5,7 @@ from typing import List, Any, Dict
 
 from basic_memory.utils import normalize_project_reference
 from markdown_it import MarkdownIt
+from markdown_it.rules_inline.backticks import backtick
 from markdown_it.token import Token
 
 # Transcript timecodes like [00:00:11] or [1:02:03.500] share the bracket shape of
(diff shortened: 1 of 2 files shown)


phernandez:
@codex review

Issue #1351, opened 2026-08-28
[BUG] API/MCP note delete leaves linking notes' relation rows stale in search_index


phernandez:
## What happens

Deleting a single note through the API/MCP path (`DELETE /v2/.../entities/{id}` → `run_accepted_note_delete` → `delete_accepted_note`) removes the deleted note's own `search_index` rows and vectors, but never touches the **relation rows owned by notes that linked to it**. Those rows keep `to_id` pointing at the deleted entity and a title like `Alpha -> Beta`, and search still returns them with a permalink to a note that no longer exists.

Found while investigating #1344; this is a separate gap and stays a gap after #1344 is fixed (see below).

## Reproduction (current `main`, SQLite; same code path on Postgres)

```
write_note Alpha  (body: "- links_to [[Beta]]")
write_note Beta
delete_note Beta   # API / MCP single-note delete
```

`search_index` afterwards (`sqlite3 memory.db`):

```
id type     entity_id to_id title          permalink
1  entity   1         NULL  Alpha          demo/notes/alpha
1  relation 1         2     Alpha -> Beta  demo/notes/alpha/links-to/demo/notes/beta   <- entity 2 no longer exists
```

User-visible:

```
search_notes("Beta", entity_types=["relation"])
→ [{"type": "relation", "title": "Alpha -> Beta", "permalink": "demo/notes/alpha/links-to/demo/notes/beta", "file_path": "notes/Alpha.md"}]
```

Nothing schedules a refresh of `Alpha`, so the row stays until `Alpha` is edited or `bm reindex` runs. Under the consistency model that's a derived-state drift with no repair mechanism on the path that produced it.

## Why only this path

Every other delete path already handles it. They snapshot the surviving relation sources *before* the entity delete and refresh those sources' search rows afterwards:

- watcher / external file delete: `relation_cleanup_sources_for_deleted_entity` in `indexing/external_file_delete_runner.py`, refreshed by `index/local_runtime.py` (`refresh_moved_entities(relation_cleanup_entity_ids)`)
- directory delete: `indexing/directory_delete_runner.py` → `services/directory_deletes.py` (`refresh_relation_sources`)
- project scan deletes: `indexing/project_index_maintenance.py` → `project_index_runtime.py`

`delete_accepted_note` (`indexing/accepted_note_write_runner.py`) is the one that just does `delete_accepted_note_search_index` + `delete_accepted_note_vectors` + `session.delete(entity)`.

## Relationship to #1344

[...]

PR #1352, opened 2026-08-28
fix(core): unresolve inbound relations on entity delete

### LINKED ISSUE #1344: [BUG] Deleting an entity removes inbound relations instead of leaving them unresolved
## What happens

When an entity is deleted, every relation pointing **at** it is deleted too.

The source notes are never touched — their markdown still contains the `[[wikilink]]` — but the relation rows are gone from the database. So a query for unresolved relations returns nothing over a project that is full of them.

The rows come back only if something re-indexes the source note. Nothing schedules that, so for a note nobody edits again, it doesn't happen.

## Reproduction

Observed on 0.23.x, on both SQLite and Postgres.

1. Create note `Alpha` containing `[[Beta]]`, and note `Beta`. Both index; one relation row, correctly resolved.
2. Remove the link line from `Alpha` — the row disappears. Correct: `Alpha` no longer claims the link.
3. Restore it, let it re-index, then delete **`Beta`'s file**. `Beta`'s entity row goes, and **the relation row goes with it** — zero rows left.
4. `Alpha`'s markdown still contains `[[Beta]]`, but no relation row exists.
5. Touch `Alpha` to force a re-index — the row comes back, correctly unresolved with `to_id` null and `to_name` preserved.

Step 5 is the interesting one: the correct end state is reachable and already modeled. Deletion just skips it.

- **The unresolved state is already the intended resting place.** `to_id` null with `to_name` preserved is exactly what the indexer produces for a link to a note that hasn't been created yet, and every reader already branches on it. #1015 and #955 both treat it that way.
- **The docs already say the narrower thing.** `docs/DOMAIN_MODEL.md` notes that re-indexing a *source* note may remove its outgoing relations. Deleting a *target* isn't that case, so the removal doesn't look licensed by the design.
- **#764 is the mirror image** — rows that survive deletion and shouldn't — and explicitly scopes the schema question out, so I don't think this overlaps it.

I couldn't find this reported in open or closed issues, PRs, or discussions. Apologies if I missed it.

## Cause

`relation.to_id` carries `ON DELETE CASCADE`, and there's a matching `delete-orphan` cascade on `Entity.incoming_relations` doing the same thing a second way — so both the database and the ORM independently remove the row.

`from_id` keeping `CASCADE` seems right: an entity does own the relations it declares, and deleting it should take them.

## A side effect worth knowing about

Because forward-reference resolution already sweeps null-target rows, letting inbound relations survive as unresolved means **recreating a deleted note automatically re-links everything that pointed at it.** That falls out for free rather than needing new code.

## Offer

I have a patch — one foreign key, the matching ORM cascade, and a migration covering both backends, with tests watched failing before the fix and passing after, plus a test for the migration itself.

Happy to open a PR if this approach looks right to you, or to rework it if you'd rather solve it a different way. One question first, since I'd rather ask than assume: the CLA is a copyright assignment, so I want to be sure I'm reading the intended process correctly before submitting — is acceptance handled entirely by the `license/cla` status check on the PR, or is there a separate step? I couldn't find one linked from CONTRIBUTING.

sammywachtel:
Fixes #1344.

## What was wrong

Delete a note, and every link pointing *at* it quietly disappeared from the graph too.

The notes doing the pointing were never touched. Their markdown still said `[[That Note]]`, right there on the page. But as far as the database was concerned, no such link had ever been written. So if you asked for the links that lead nowhere — the report every knowledge base eventually needs, and the one that catches rot before it spreads — it came back clean over a vault full of them.

That is the part worth caring about. Not that the answer was missing, but that the system gave a confident wrong one. A tool that says "no broken links" is telling you to stop looking.

The link came back only if something happened to re-index the note containing it. Nothing schedules that. For a note nobody edits again: never.

## The cause

`relation.to_id` carried `ON DELETE CASCADE`, and `Entity.incoming_relations` carried a matching ORM `delete-orphan` cascade — two separate mechanisms doing the same damage on two different delete routes. The watcher and directory runners take a bulk SQL `DELETE`, so the database constraint ate the rows. The API single-entity delete takes `session.delete(entity)`, so the ORM ate them.

`to_id` has always been nullable, and unresolved-with-`to_name`-preserved is a state the system already produces and already handles everywhere — it is what the indexer writes for a link to a note that does not exist yet. The schema could hold the right answer the whole time. Nothing chose `CASCADE` over it; it was inherited from `from_id`, where it *is* correct.

`docs/DOMAIN_MODEL.md` already said as much: the relation is owned by its source, and resolution "enriches that statement; it does not replace it". Deleting the target should un-enrich, not delete.

## The fix

`to_id` becomes `ON DELETE SET NULL`, and `incoming_relations` drops `delete-orphan` for `passive_deletes=True` so the ORM hands the job to the database instead of doing it itself. `from_id` keeps `CASCADE` — an entity really does own the relations it declares.

Both halves are load-bearing, and the tests below demonstrate that separately rather than asserting it.

Because unresolved is an already-supported state, nothing downstream needed teaching. Forward-reference resolution already sweeps `to_id IS NULL` rows, so **recreating a deleted note re-links its inbound references for free** — no new pass was built, and the generation fence in `find_unresolved_relations` is untouched. The `relation_cleanup_entity_ids` machinery in the watcher and directory runners still works unchanged; it now repairs surviving sources' search rows to a null target rather than watching the rows disappear.

### Migration

`bcdbd5a942ca_relation_to_id_set_null_on_delete.py`, revising `7f6a2b8c9d10`.

SQLite goes through `op.batch_alter_table("relation")`, following `a1b2c3d4e5f6`. The initial-schema foreign key is unnamed, and SQLite cannot address an anonymous constraint, so the batch gets a `naming_convention` and refers to the reflected FK by the name that convention would have given it. One visible side effect: the table recreate writes those names into the schema, so `from_id`'s foreign key comes out named too. Verified against a real migrated database — the resulting DDL is identical except for the `ON DELETE` action and those two constraint names, with every index, unique constraint, the `generation` column's server default, and all rows intact.

Postgres drops and recreates `relation_to_id_fkey`. Verified against a real Postgres: after the migration, `pg_constraint.confdeltype` for `relation_to_id_fkey` is `n` (set null) while `relation_from_id_fkey` stays `c` (cascade).

**Rows already lost to the old behavior stay lost.** The migration changes the rule going forward; it cannot recover what `CASCADE` already ate. Re-index the affected notes to bring them back as unresolved.

## Tests, watched failing

Three new behavior tests, one per delete path, each doing delete → unresolved → recreate → re-linked:

- `tests/api/v2/test_delete_unresolves_inbound_relations.py` — the API single-entity delete (`DELETE /knowledge/entities/{id}`, the ORM `session.delete` route) and the directory delete (`POST /knowledge/delete-directory`, the guarded bulk-SQL route).
- `tests/index/test_watch_delete_unresolves_inbound_relation.py` — the watcher removing a note whose file vanished from disk (the `delete_by_fields` bulk-SQL route).

A fourth test pins the constraint question you raised. `uix_relation_from_id_to_id` covers `(from_id, to_id, relation_type)`, so the obvious worry about `SET NULL` is two inbound rows from one source colliding once both their targets are gone. They do not — NULL is distinct from NULL under both backends — but that is the kind of thing that should be a test rather than a paragraph, so it is one: one source linking to two notes in a directory, the directory deleted, both rows surviving unresolved and still told apart by `to_name`.

**How they were made to fail.** I reverted the model change three ways and re-ran, because "it passes now" says nothing on its own:

| Model state | Result |
|---|---|
| Both halves reverted — `ON DELETE CASCADE` + ORM `delete-orphan` (i.e. `main` today) | **3 failed** |
| Only the FK reverted — `CASCADE`, ORM keeps `passive_deletes` | **3 failed** |
| Only the ORM reverted — `SET NULL` kept, ORM back to `delete-orphan` | **1 failed** (the API path) |

```
FAILED tests/api/v2/...::test_delete_entity_unresolves_inbound_relation_and_recreate_relinks
  - AssertionError: the inbound relation must survive its target's deletion
FAILED tests/api/v2/...::test_delete_directory_unresolves_inbound_relations_from_outside
  - AssertionError: the keeper's relation must survive the directory delete
FAILED tests/index/...::test_watch_delete_unresolves_inbound_relation_and_recreate_relinks
  - AssertionError: the inbound relation must survive the watcher's bulk delete
```

That third row is the interesting one, and it is exactly the trap you flagged: with the database fixed but the ORM cascade left in place, the two bulk-SQL paths go green and only the ORM path stays red. Fix one and not the other and the bug does not go away — it just moves somewhere quieter and gets harder to find.

### Migration test

`tests/test_relation_to_id_set_null_migration.py`, five tests, following `tests/test_note_content_migration.py`.

This one earns its place: the SQLite suite builds its schema straight from SQLAlchemy metadata, so a corrected model with a missing or wrong migration would sail through the entire suite green while every database on disk kept the old `CASCADE`.

It runs the real Alembic chain to the previous revision, asserts `CASCADE`, upgrades, and asserts `SET NULL` — plus a data-preservation test that inserts a project, two entities and a relation before the upgrade, then deletes the target with `PRAGMA foreign_keys = ON` and watches the row survive with `to_name` intact. A downgrade test proves the step reverses. The Postgres branch is covered by recording the `op` calls, the way `tests/test_postgres_full_content_search_migration.py` does.

Made to fail by pointing `upgrade()` at `CASCADE`:

```
FAILED ...::test_sqlite_upgrade_sets_relation_to_id_null_on_delete
  - AssertionError: assert 'CASCADE' == 'SET NULL'
FAILED ...::test_sqlite_upgrade_preserves_relation_rows_and_indexes
  - AssertionError: assert None == (None, 'Note 2')
FAILED ...::test_sqlite_downgrade_restores_cascade
  - AssertionError: assert 'CASCADE' == 'SET NULL'
FAILED ...::test_postgres_upgrade_recreates_to_id_fk_with_set_null
  - AssertionError: ... 'CASCADE') != ... 'SET NULL')
```

The second line there is the bug itself reproduced at the schema level: the relation row was gone, so the query returned `None` instead of the unresolved row.

### Two existing assertions changed

`tests/index/test_local_watch_regular_file_parity.py` and `tests/index/test_local_project_index.py` each asserted `outgoing_relations == []` after a target delete. Both tests are *about* search-row repair; the empty-relations line rode along as scenery. Nobody argued for it, but leaving it would have turned an accident into a contract. Both now assert the surviving unresolved row.

## Out of scope

The stale `search_index` row on the API delete path — your #1351 — is untouched here, as you asked.

## Local run

`just test` — all four stages, SQLite and Postgres, unit and integration:

| Stage | Result |
|---|---|
| unit / SQLite | 5032 passed, 3 failed |
| integration / SQLite | 475 passed, 0 failed |
| unit / Postgres | 5014 passed, 3 failed |
| integration / Postgres | 0 failed, reached 73% before the recipe's own 600s `gtimeout` fired (exit 137, which `test-int-postgres` treats as success — the documented FastMCP/asyncpg cleanup hang) |

The three failures in both unit stages are the same three, all in `tests/cli/test_cloud_promo.py`, and they are not mine: I checked out unmodified `upstream/main` and ran them there, where they fail identically. The five files this PR touches or adds were also run explicitly against Postgres — 56 passed.

Two notes on running the suite locally, offered only in case they are useful to someone else:

- A shell exporting `FORCE_COLOR` fails ~41 CLI tests on any checkout, fix or not: Rich emits escape codes into `CliRunner`'s captured output and the assertions compare against plain strings. I ran with it unset.
- `just check` fails on a clean tree with four `unresolved-import` diagnostics for `pymilvus`, which `uv sync` does not install. Reproduced on unmodified `main`. `just lint` and `just format` are clean; `just format` runs `ruff format .`, which also reformats `benchmarks/`, so I formatted only `src tests test-int` to keep that out of the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- **#1344**: Deleting a note no longer erases the relations pointing at it. The
+  `relation.to_id` foreign key is now `ON DELETE SET NULL` rather than
+  `ON DELETE CASCADE`, and `Entity.incoming_relations` no longer cascades deletes
+  through the ORM. Deleting a target leaves its inbound relations in the unresolved
+  state the indexer already produces for a link to a note that does not exist yet --
+  `to_id` null, `to_name` holding the source's original link text -- so a
+  broken-link report stops coming back clean over a vault full of them, and
+  recreating the target re-links the references through the existing
+  forward-reference pass. `from_id` keeps `CASCADE`: an entity does own the
+  relations it declares. Rows already lost to the old behavior stay lost; re-index
+  the affected notes to bring them back as unresolved.
+
 ## v0.23.2 (2026-08-25)
 
 Patch release fixing case-duplicate folder creation.
--- a/docs/DOMAIN_MODEL.md
+++ b/docs/DOMAIN_MODEL.md
@@ -101,6 +101,13 @@ A relation is a directed semantic statement owned by its source entity.
 Incoming graph navigation does not transfer ownership to the target. Re-indexing the source note
 may recreate, resolve, or remove its outgoing relations.
 
+Deleting the target is not one of those events. When a target entity is deleted, its inbound
+relations survive with `to_id` cleared and `to_name` intact -- the same unresolved state the
+indexer produces for a link to a note that does not exist yet. The source note's markdown still
+carries the reference, so the graph must keep carrying it too; a graph that quietly drops the edge
+reports a clean bill of health over a broken link. Forward-reference resolution re-links the row
+if the target is ever recreated.
+
 ### NoteContent
 
 `NoteContent` is the operational record of accepted Markdown bytes and their file materialization
--- a/src/basic_memory/alembic/versions/bcdbd5a942ca_relation_to_id_set_null_on_delete.py
+++ b/src/basic_memory/alembic/versions/bcdbd5a942ca_relation_to_id_set_null_on_delete.py
@@ -1,0 +1,77 @@
+"""relation.to_id foreign key becomes ON DELETE SET NULL
+
+Deleting an entity used to take every inbound relation row with it. The wikilink
+stayed in the source note's markdown, but the graph forgot the edge existed, so a
+"find broken links" report came back clean over a vault full of them.
+
+to_id has always been nullable, and the indexer already produces exactly the state
+we want here -- to_id NULL with to_name holding the source's link text -- for a link
+pointing at a note that does not exist yet. A deleted target should land in that same
+unresolved state instead of vanishing, and forward-reference resolution then re-links
+the row for free if the target is ever recreated.
+
+from_id keeps CASCADE. An entity really does own the relations it declares.
+
+Revision ID: bcdbd5a942ca
+Revises: 7f6a2b8c9d10
+Create Date: 2026-08-28 00:00:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision: str = "bcdbd5a942ca"
+down_revision: Union[str, None] = "7f6a2b8c9d10"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+# The initial schema created relation's entity foreign keys without names, and SQLite
+# has no way to reference an anonymous constraint. Handing batch_alter_table a naming
+# convention lets it address the reflected FK by the name the convention would have
+# given it. Side effect worth knowing about: the table recreate writes those names into
+# the schema, so from_id's FK comes out named too. Cosmetic, and an improvement.
+RELATION_NAMING_CONVENTION = {"fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s"}
+SQLITE_TO_ID_FK = "fk_relation_to_id_entity"
+
+# Postgres named it for us when the initial schema left it anonymous.
+POSTGRES_TO_ID_FK = "relation_to_id_fkey"
+
+
+def _set_to_id_ondelete(ondelete: str) -> None:
+    """Point relation.to_id's foreign key at a different ON DELETE action."""
+    dialect = op.get_bind().dialect.name
+
+    if dialect == "sqlite":
+        # SQLite cannot alter a foreign key in place; batch mode does the copy-and-swap.
+        # Same shape as a1b2c3d4e5f6.
+        with op.batch_alter_table(
+            "relation", schema=None, naming_convention=RELATION_NAMING_CONVENTION
+        ) as batch_op:
+            batch_op.drop_constraint(SQLITE_TO_ID_FK, type_="foreignkey")
+            batch_op.create_foreign_key(
+                SQLITE_TO_ID_FK, "entity", ["to_id"], ["id"], ondelete=ondelete
+            )
+    else:
+        op.drop_constraint(POSTGRES_TO_ID_FK, "relation", type_="foreignkey")
+        op.create_foreign_key(
+            POSTGRES_TO_ID_FK, "relation", "entity", ["to_id"], ["id"], ondelete=ondelete
+        )
+
+
+def upgrade() -> None:
+    """Stop erasing an entity's inbound relations along with the entity."""
+    _set_to_id_ondelete("SET NULL")
+
+
+def downgrade() -> None:
+    """Go back to erasing inbound relations on delete.
+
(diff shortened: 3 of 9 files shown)


phernandez:
@codex review

phernandez:
This is excellent work — the two-mechanism diagnosis, the red/green matrix showing the DB and ORM halves are independently load-bearing, the migration test that runs the real Alembic chain, and the NULL-uniqueness test are exactly what this change needed. Verified locally on top of it: static checks clean, your 56 tests plus 2186 unit and 482 integration tests pass, and Codex found nothing.

Fork PRs don't get our test matrix, so I've cherry-picked your commit unchanged (authorship and sign-off preserved) onto an in-repo branch as #1358 to run the full matrix including Postgres; it will be rebase-merged so your commit lands on `main` as-is. Closing this one in favour of #1358 — nothing further needed from you. Thanks for taking it from report to fix so carefully.

Issue #1343, opened 2026-08-27
bm project add --visibility is silently ignored: private and shared both create team-visible projects


nellins:
## Summary

`bm project add --cloud --visibility <private|shared>` silently creates a project that **everyone in the team workspace can see**. The flag is validated but never applied.

This is worth treating as more than cosmetic: a user runs `--visibility private`, gets `Project 'x' added successfully`, and reasonably believes the project is private. It is not — the whole team can find and open it.

## Reproduce

```bash
bm project add visibility-check --cloud --workspace "Team Drew" --visibility private
# → Project 'visibility-check' added successfully
```

Then check it:

```bash
bm tool list-projects
# → { "name": "visibility-check", "is_private": false, ... }
```

And in the app, **Settings → Projects**:

> **Visibility Check**
> Project access: **Team**
> Everyone in Team Drew can find this project.

Same result with `--visibility shared` (verified separately as `vis-shared-check` — also `is_private: false`, also "Project access: Team" in the app).

## The flag *is* being parsed

It isn't ignored at the argument layer — a bogus value is rejected:

```bash
bm project add x --cloud --workspace "Team Drew" --visibility notavalue
# → Error: Invalid visibility. Expected one of: workspace, shared, private.
```

So the CLI accepts `private`, validates it against the allowed set, and then drops it.

## Related

#705 ("CLI: support project visibility when creating projects in a teams workspace") is closed, so this was expected to work. Either the implementation regressed or it never reached the create path.

## Two adjacent things worth a look

**1. Visibility can't be read back from the CLI at all.** Neither `bm project list`, `bm project info`, nor `bm tool list-projects` reports the effective access level (`is_private` is the closest, and it is false here regardless). There is no way to confirm from the terminal whether a project is private, which is how this went unnoticed.

**2. The CLI's vocabulary doesn't match the app's.** The CLI and docs describe three tiers — `workspace` / `shared` / `private`. The app's **Edit Project Access** dialog offers two: **Team** and **Invite-only**. There is no "private / just you" option in the UI. If `private` is not a real tier, the CLI probably shouldn't accept it; if it is, the app should expose it.

## Environment

basic-memory 0.23.2, macOS, cloud-routed projects in an organization workspace, Owner role.

phernandez:
Triage, verified in both repos:

[...]

PR #1357, opened 2026-08-28
fix(cli): stop accepting a private project visibility the cloud does not have

phernandez:
Core half of #1343 (the cloud half is basicmachines-co/basic-memory-cloud#1851).

## What's wrong

`bm project add --cloud --visibility private` validates `private` as a legal tier and prints "added successfully". The cloud's project-share schema only knows `workspace` and `shared` (`project_share_schemas.py`: `Literal["workspace", "shared"]`), and its `create_project` doesn't read `visibility` at all — so the project is created team-visible while the user believes it is private. The CLI side that made this *silently* wrong is accepting a tier that doesn't exist.

## Change

- `ProjectVisibility` narrows to `Literal["workspace", "shared"]`; `_normalize_project_visibility` rejects `private` with the corrected tier list; the `--visibility` help text describes the two real tiers.
- Test: `--visibility private` exits 1 before any create call.

Not in this PR: honoring `shared` at create time (cloud#1851), and exposing the effective visibility in `project list/info` — that needs the create/list responses to carry it, which is also on the cloud side.

### DIFF
--- a/src/basic_memory/cli/commands/project.py
+++ b/src/basic_memory/cli/commands/project.py
@@ -279,10 +279,12 @@ def _normalize_project_visibility(visibility: str | None) -> ProjectVisibility:
         return "workspace"
 
     normalized = visibility.strip().lower()
-    if normalized in {"workspace", "shared", "private"}:
+    if normalized == "workspace" or normalized == "shared":
         return normalized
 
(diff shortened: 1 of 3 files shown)


phernandez:
@codex review

PR #1358, opened 2026-08-28
fix(core): unresolve inbound relations on entity delete

### LINKED ISSUE #1344: [BUG] Deleting an entity removes inbound relations instead of leaving them unresolved

phernandez:
Supersedes #1352 by @sammywachtel — their commit is cherry-picked here unchanged, authorship and sign-off preserved, so the full CI matrix (Postgres included) runs; fork PRs only get DCO/CLA/CodeQL. Will be rebase-merged so the commit lands on `main` as-is.

Fixes #1344.

## Summary

Deleting an entity took every inbound relation row with it — `relation.to_id` carried `ON DELETE CASCADE` and `Entity.incoming_relations` a matching ORM `delete-orphan`, one per delete route (bulk-SQL runners vs. `session.delete`). The source notes still contained the `[[wikilink]]`, so a broken-link report came back clean over a vault full of them, and nothing re-created the rows short of re-indexing the source.

`to_id` becomes `ON DELETE SET NULL` and `incoming_relations` uses `passive_deletes=True`; `from_id` keeps `CASCADE`. Inbound rows land in the unresolved state the indexer already produces (`to_id` NULL, `to_name` preserved), and forward-reference resolution re-links them if the target is recreated — matching `docs/DOMAIN_MODEL.md`'s "resolution enriches the statement; it does not replace it". Migration `bcdbd5a942ca` (SQLite batch mode with a naming convention for the anonymous initial-schema FK; Postgres drop/recreate `relation_to_id_fkey`), with migration tests that run the real Alembic chain.

## Verification (local, this branch, SQLite)

- `ruff check` / `ruff format --check` / `ty check src tests test-int` (all extras) — clean
- PR's own tests (API delete, directory delete, watcher delete, NULL-uniqueness, 5 migration tests) — 56 passed
- `tests/api tests/index tests/indexing tests/repository tests/services tests/db` — 2186 passed
- `test-int` (not slow/benchmark) — 482 passed
- Codex on #1352: no major issues

The contributor also verified the migration against real Postgres (`pg_constraint.confdeltype` = `n` for `to_id`, `c` for `from_id`) and ran `just test` across both backends; see #1352 for the red/green matrix showing the DB and ORM halves are independently load-bearing.

Out of scope, tracked separately: the stale `search_index` row on the API delete path (#1351).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- **#1344**: Deleting a note no longer erases the relations pointing at it. The
+  `relation.to_id` foreign key is now `ON DELETE SET NULL` rather than
+  `ON DELETE CASCADE`, and `Entity.incoming_relations` no longer cascades deletes
+  through the ORM. Deleting a target leaves its inbound relations in the unresolved
+  state the indexer already produces for a link to a note that does not exist yet --
+  `to_id` null, `to_name` holding the source's original link text -- so a
+  broken-link report stops coming back clean over a vault full of them, and
+  recreating the target re-links the references through the existing
+  forward-reference pass. `from_id` keeps `CASCADE`: an entity does own the
+  relations it declares. Rows already lost to the old behavior stay lost; re-index
+  the affected notes to bring them back as unresolved.
+
 ## v0.23.2 (2026-08-25)
 
 Patch release fixing case-duplicate folder creation.
--- a/docs/DOMAIN_MODEL.md
+++ b/docs/DOMAIN_MODEL.md
@@ -101,6 +101,13 @@ A relation is a directed semantic statement owned by its source entity.
 Incoming graph navigation does not transfer ownership to the target. Re-indexing the source note
 may recreate, resolve, or remove its outgoing relations.
 
+Deleting the target is not one of those events. When a target entity is deleted, its inbound
+relations survive with `to_id` cleared and `to_name` intact -- the same unresolved state the
+indexer produces for a link to a note that does not exist yet. The source note's markdown still
+carries the reference, so the graph must keep carrying it too; a graph that quietly drops the edge
+reports a clean bill of health over a broken link. Forward-reference resolution re-links the row
+if the target is ever recreated.
+
 ### NoteContent
 
 `NoteContent` is the operational record of accepted Markdown bytes and their file materialization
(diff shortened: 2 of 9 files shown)


phernandez:
@codex review

PR #1362, opened 2026-08-29
fix(core): make case-only renames stick on case-insensitive filesystems

### LINKED ISSUE #1281: [BUG] Case-only renames on macOS (case-insensitive APFS) fail via move_note and create duplicate index entries
> **Note:** This issue was written with AI assistance, based on a real usage session.

## Bug Description
On macOS with the default **case-insensitive** APFS volume, renaming notes or folders so that only the **letter case** changes misbehaves when done through the MCP tools. Several distinct problems: a case-only file rename is rejected, a case-only *folder* rename silently doesn't change the folder's case on disk, directory moves leave empty orphan directories behind, and case-only renames leave duplicate entries in the index that a normal sync won't clear.

## Steps To Reproduce
1. Install version `0.22.1`, vault on default case-insensitive APFS.
2. **File rename:** call `move_note` to rename `config.md` → `Config.md` (same folder, only case differs).
3. **Folder rename:** call `move_note` (directory) to rename a folder `reviews` → `Reviews`.
4. Let the server sync / restart it, then `list_directory`; also inspect the folder on disk.

## Expected Behavior
- A case-only rename succeeds via `move_note` (a plain `mv config.md Config.md` / Finder does it fine).
- A case-only *folder* rename actually changes the folder's case on disk.
- Directory moves remove the emptied source directory.
- No duplicate/phantom index entries result.

## Actual Behavior
1. **`move_note` refuses the case-only file rename** with `Move Failed - Destination Already Exists` — its destination-exists check treats the case-insensitive-equal path as an existing file. Requires a two-step rename via a temporary name to work around.
2. **The case-only folder rename reports success but doesn't change the folder's case on disk** — the files end up back in the pre-existing lowercase directory; only the index briefly reflects the new case. (A `mv` on the folder does work.)
3. **Directory moves leave the emptied source directory behind on disk.** After a directory move, the original (now empty) folder remains — e.g. an empty `debuggng/` is left after its contents move to `Debugging/` — so renames accumulate orphan empty folders.
4. **Duplicate index entries:** after renaming, every affected note appears twice in `list_directory` — once under the new-cased path, once under the old-cased path — and an incremental sync / server restart does **not** remove the stale ones (they even get re-dated). Running **`bm reindex --full`** rebuilds from the filesystem and clears them.

## Environment
- OS: macOS **26.6.1** (default case-insensitive APFS)
- Python version: **3.14.7**
- Basic Memory version: **0.22.1**
- Installation method: **Homebrew**
- Client: MCP (streamable-http transport)

## Additional Context
- `mv`/Finder perform these case-only renames fine, so this is about the tool/index behavior, not a filesystem limitation.
- Workaround that fully recovers a duplicated index: `bm reindex --full -p <project>` (incremental sync alone does not). Orphan empty directories must be removed manually (e.g. `rmdir`).
- Possibly related: #1275 (NFC/NFD filename variants duplicating entities on macOS) — may share a cause, though that's Unicode form rather than letter case.

## Possible Solution
Treat a case-insensitive same-path destination as the same file in `move_note` (allow the rename), remove the emptied source directory on a directory move, and make sync reconcile case-only path changes so duplicates aren't created / are pruned — rather than relying on a full reindex.

phernandez:
Fixes #1281.

## What actually happens today (re-checked on `main`)

The 0.22.1 "already exists" rejection is gone since #1002's samefile guards, but the rename still doesn't land:

1. `move_note docs/config.md → docs/Config.md` returns 202 and the row becomes `docs/Config.md`.
2. Materialization writes through `temp_path.replace(target)`; on APFS/NTFS that replaces the bytes through the existing case-variant entry and **keeps the old casing** — disk stays `config.md`.
3. Old-path cleanup is (correctly) skipped by the same-inode guard.
4. The next scan misses `config.md` byte-wise, checksum move-detection rewrites the row back to `docs/config.md`, and the permalink churns `config → config-1 → config` because the move's collision check counts the entity itself.

No duplicate rows anymore (those came from the pre-#1002 flow); the reporter's duplicates were on 0.22.1.

## Fix

- `InlineNoteFileDeleteEnqueuer` already recognizes the same-inode alias; when the two spellings differ only by case it now renames the directory entry to the accepted casing (`FileService.move_file` → `Path.rename`). On a case-sensitive filesystem the alias check only passes for hard links, where a rename onto itself is a POSIX no-op — the existing hard-link P0 test still passes.
- `resolve_permalink` takes `current_file_path` (the row being re-resolved) and does not suffix a permalink that row already owns. `prepare_move_entity_content` and the scan-detected move planner pass it; create paths are unchanged.

## Tests

- `test_inline_delete_adopts_accepted_casing_for_case_only_rename` — on a case-insensitive filesystem the entry ends up spelled `Config.md` with the bytes intact (skips on case-sensitive FS).
- `test_resolve_permalink_keeps_the_permalink_the_entity_already_owns` — `docs/Config.md` resolves to `docs/config-1` without the hint and to `docs/config` with it.
- Touched suites (move tool, move content updates, materialization, permalink formatting, entity service, knowledge router) pass; `ruff` / `ty` clean.

Out of scope, as discussed on the issue: directory case-only renames and empty-dir cleanup.

### DIFF
--- a/src/basic_memory/index/local_dependencies.py
+++ b/src/basic_memory/index/local_dependencies.py
@@ -201,6 +201,7 @@ class LocalIndexEntityService(Protocol):
         file_path: Path | str,
         markdown: EntityMarkdown | None = None,
         skip_conflict_check: bool = False,
+        current_file_path: str | None = None,
         session: AsyncSession | None = None,
     ) -> str: ...
 
--- a/src/basic_memory/index/local_moves.py
+++ b/src/basic_memory/index/local_moves.py
@@ -70,6 +70,7 @@ class LocalMoveEntityService(Protocol):
         file_path: Path | str,
         markdown: EntityMarkdown | None = None,
         skip_conflict_check: bool = False,
+        current_file_path: str | None = None,
         session: AsyncSession | None = None,
     ) -> str: ...
 
@@ -125,6 +126,7 @@ class LocalProjectIndexMoveContentUpdater(ProjectIndexMoveContentUpdater):
         permalink = await self.entity_service.resolve_permalink(
             Path(moved_file.new_path),
             skip_conflict_check=True,
+            current_file_path=moved_file.old_path,
             session=session,
         )
         if permalink == moved_file.old_permalink:
--- a/src/basic_memory/index/note_content_materialization.py
+++ b/src/basic_memory/index/note_content_materialization.py
@@ -579,6 +579,17 @@ class InlineNoteFileDeleteEnqueuer:
                 file_path=request.file_path,
                 live_file_path=request.live_file_path,
             )
+            # Trigger: the two spellings differ only by case.
+            # Why: the atomic write replaced the file's bytes through the existing
+            #      directory entry, so the entry still carries the old casing; the
+            #      next scan would read that as a move back and undo the rename (#1281).
+            # Outcome: rename the entry to the accepted casing so disk agrees with the
+            #          row. On a case-sensitive filesystem the alias check above only
+            #          passes for hard links, where a rename onto itself is a no-op.
+            if request.file_path != request.live_file_path and (
+                request.file_path.casefold() == request.live_file_path.casefold()
+            ):
+                await self.storage.file_service.move_file(request.file_path, request.live_file_path)
             # No separate source object exists: the old path aliases the live destination, so the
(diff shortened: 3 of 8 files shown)


phernandez:
@codex review

phernandez:
@codex review

PR #1364, opened 2026-08-29
fix(core): sanitize quoted PostgreSQL Boolean searches

### LINKED ISSUE #1359: PostgreSQL search parser generates invalid tsquery for quoted OR expressions
## Summary

A production hosted-MCP search accepted quoted Boolean-looking input such as quoted phrases joined by OR, then constructed invalid PostgreSQL tsquery syntax. The request emitted repeated PostgresSyntaxError warnings and one search execution reached the 30-second timeout.

This is a Core search parsing/correctness bug. Cloud rate limiting occurred in the same trace but is a separate expected backpressure outcome.

## Production evidence

- Representative Logfire trace: https://logfire-us.pydantic.dev/basic-memory/basic-memory-cloud/?q=trace_id%3D%276a92126f00000000d7ebdd4d9e278f87%27
- Observed Cloud release: bcdb30feac9a92939c390a96eb3a0af03972eb76
- Input shape: multiple quoted phrases joined with OR.
- Core emitted PostgreSQL syntax errors for generated tsquery strings containing unmatched quote/token boundaries.
- The same trace contains a 30-second TimeoutError under search.execute and the enclosing MCP search path.

No customer filenames or note contents are included here.

## RCA

Trigger: a caller supplied a search term containing quoted phrases and Boolean-looking OR separators.

Root cause: Core's PostgreSQL FTS query construction does not construct a valid typed tsquery for this accepted input shape. It lets quote/operator syntax cross into the database expression, then the fallback path repeats work and can exhaust the request timeout.

Propagation: the syntax failure is logged as a warning, search continues through fallback/hybrid work, and the enclosing hosted MCP call can time out. Concurrent Cloud rate limiting makes the trace noisy but does not cause the invalid tsquery.

Impact: customer-visible search failure or timeout. Retrying the same query shape is not a valid recovery because it reconstructs the same invalid expression.

## Fix scope

- Parse accepted user search syntax into a safe internal term representation before constructing PostgreSQL tsquery expressions.
- Treat quote characters and Boolean-looking words as user text unless the public search contract explicitly supports Boolean syntax.
- Ensure fallback does not repeat an identically invalid database expression.
- Preserve ordinary phrase, prefix, and hybrid search behavior.

## Acceptance criteria

- Quoted phrases joined by OR produce valid SQL/tsquery input and return a normal search result or an explicit no-match outcome.
- Unbalanced quotes and operator-like input cannot produce PostgresSyntaxError.
- A regression test proves the production query shape does not reach the 30-second timeout.
- Existing phrase, prefix, and hybrid-search tests remain green.
- Invalid search construction remains actionable; the fix must not merely lower telemetry severity.

## Related

Cloud observability and expected-outcome work: basicmachines-co/basic-memory-cloud#1661.

mikemikimike:
## Summary

Fix PostgreSQL search parsing for quoted Boolean-looking input such as quoted phrases joined by `OR`.

The PostgreSQL Boolean preparer now sanitizes each operand with the existing single-term rules while preserving supported `AND`, `OR`, `NOT`, and grouping behavior. This prevents quote and punctuation syntax from reaching `to_tsquery()` as invalid expressions.

## Testing

- `uv run --frozen pytest -p pytest_mock -q tests/repository/test_search_repository.py --no-cov` (48 passed, 1 skipped)
- `uv run --frozen ruff check src/basic_memory/repository/postgres_search_repository.py tests/repository/test_search_repository.py`
- `uv run --frozen ty check src/basic_memory/repository/postgres_search_repository.py tests/repository/test_search_repository.py`

Fixes #1359

### DIFF
--- a/src/basic_memory/repository/postgres_search_repository.py
+++ b/src/basic_memory/repository/postgres_search_repository.py
@@ -394,7 +394,7 @@ class PostgresSearchRepository(SearchRepositoryBase):
         # Split on Boolean keywords first, then sanitize each operand using the
         # same rules as ordinary terms. This keeps user punctuation (including
         # quotes and unbalanced grouping characters) from reaching to_tsquery.
-        parts = re.split(r"\b(AND|OR|NOT)\b", query, flags=re.IGNORECASE)
+        parts = re.split(r"\b(AND|OR|NOT)\b", query)
         result: list[str] = []
         for index, part in enumerate(parts):
             if index % 2:
--- a/src/basic_memory/repository/postgres_search_repository.py
+++ b/src/basic_memory/repository/postgres_search_repository.py
@@ -391,16 +391,27 @@ class PostgresSearchRepository(SearchRepositoryBase):
             "(pour OR french) AND press" -> "(pour | french) & press"
             "coffee NOT decaf" -> "coffee & !decaf"
         """
-        # Replace Boolean operators with tsquery operators
-        # Keep parentheses for grouping
-        result = query
-        result = re.sub(r"\bAND\b", "&", result)
-        result = re.sub(r"\bOR\b", "|", result)
-        # NOT must be converted to "& !" and the ! must be attached to the following term
-        # "Python NOT Django" -> "Python & !Django"
-        result = re.sub(r"\bNOT\s+", "& !", result)
+        # Split on Boolean keywords first, then sanitize each operand using the
+        # same rules as ordinary terms. This keeps user punctuation (including
+        # quotes and unbalanced grouping characters) from reaching to_tsquery.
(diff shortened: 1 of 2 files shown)


mikemikimike:
Addressed the review feedback in commit `199fcb6`. Boolean operator recognition is now case-sensitive, and the regression test covers `foo OR not` remaining an operand (`foo | not`). Focused tests, Ruff, and ty pass.

mikemikimike:
Follow-up: the branch history was cleaned up so the fix is now represented by a single signed commit `a5d75c7` on top of the original PR commit. The earlier intermediate commits are no longer part of this PR.

Issue #952, opened 2026-06-11
Create Unix-style Basic Memory manual pages as notes


phernandez:
## Motivation

Basic Memory should have a documentation set written in the style of Unix man pages, but implemented as Basic Memory notes.

This is related to #610, which focuses on shipping traditional man pages for `bm` CLI commands. This issue is broader: create a `manual` or `manpages` Basic Memory project/folder that documents commands, file formats, concepts, schemas, and advanced memory patterns using man-page structure and Basic Memory graph links.

The product idea fits the Unix/Git framing:

- Unix man pages are compact, structured, durable documentation.
- Git became powerful by introducing new nouns and verbs.
- Basic Memory is developing its own nouns and verbs: notes, observations, relations, schemas, activity, versions, materialized views, and memory routines.
- If Basic Memory docs are also Basic Memory notes, the documentation becomes searchable, linkable, semantically indexed, and useful to both humans and agents.

## Proposal

Create a Basic Memory manual page documentation set.

Example note frontmatter:

```yaml
title: bm-write(1)
type: manpage
section: 1
name: bm-write
summary: write or update a Basic Memory note
tags: [manual, cli, notes]
```

Example body:

```markdown
# bm-write(1)

## NAME
bm-write - write or update a Basic Memory note

## SYNOPSIS
bm write [PROJECT] [TITLE] [OPTIONS]

## DESCRIPTION
Creates or updates a markdown note, parses semantic observations, extracts wikilinks, and updates the knowledge graph.

## OPTIONS
--type
--tags
--folder
--overwrite

## EXAMPLES
bm write "Decision - Use Postgres"

## SEE ALSO
[[bm-search(1)]], [[bm-read(1)]], [[bm-schema(5)]], [[bm-routine(5)]]
```

## Manual Sections

Use traditional Unix-style section numbers adapted for Basic Memory:

- `1` - Commands: `bm(1)`, `bm-search(1)`, `bm-read(1)`, `bm-write(1)`, `bm-sync(1)`, `bm-cloud(1)`
- `3` - APIs / MCP tools: `search-notes(3)`, `read-note(3)`, `write-note(3)`, `build-context(3)`
- `5` - File formats / schemas: `bm-note(5)`, `bm-observation(5)`, `bm-relation(5)`, `bm-schema(5)`, `bm-routine(5)`
[...]

Issue #610, opened 2026-02-26
Add man pages for bm CLI commands


bm-clawd:
## Motivation

Man pages are the original machine-readable documentation — structured, parseable, and now the ideal format for AI agents to understand CLI tools. With agents increasingly using CLIs directly (see [Karpathy's take](https://x.com/karpathy/status/2026360908398862478)), shipping proper man pages makes `bm` a first-class citizen in agent workflows.

The `bm` CLI has enough subcommands and complex flag combinations (especially `search-notes` with `--hybrid`, `--min-similarity`, `--metadata-filters`, `--project`, etc.) that `--help` output alone isn't sufficient. Man pages provide the EXAMPLES section that both humans and agents need.

References:
- [Unix man pages: AI-friendly documentation since 1971](https://www.nutrient.io/blog/unix-man-pages-ai-friendly-documentation/)
- [Unix.com: man pages, LLMs, and why this old site still lives](https://community.unix.com/t/unix-com-man-pages-llms-and-why-this-old-site-still-lives/397390)

### 1. Ship man pages with `bm`

Start with the most complex commands:
- `bm-search-notes(1)` — hybrid search, metadata filters, similarity thresholds, project scoping
- `bm-build-context(1)` — depth, memory URLs, graph traversal
- `bm-project(1)` — add, list, remove, reindex
- `bm(1)` — overview with SEE ALSO to subcommands

Each man page should follow the standard structure:
```
NAME
SYNOPSIS
DESCRIPTION
OPTIONS
EXAMPLES      ← this is the key section for agents
EXIT STATUS
SEE ALSO
```

The EXAMPLES section is critical — real-world queries that agents can adapt:
```
EXAMPLES
    Search by semantic similarity:
        bm tool search-notes "machine learning" --hybrid

    Filter by metadata:
        bm tool search-notes --metadata '{"status": "active"}' --project claw

    Cross-project search:
        bm tool search-notes "API design" --project specs

    Lower similarity threshold for broader results:
        bm tool search-notes "coffee" --min-similarity 0.3 --page-size 20
```

### 2. Implementation options

- **click-man**: Auto-generate man pages from Click/Typer CLI definitions. Gets us 80% there.
- **Hand-written mdoc/roff**: Better EXAMPLES sections, but more maintenance.
- **Hybrid**: Auto-generate structure, hand-write EXAMPLES.

Install location: `man bm-search-notes` should work after `uv tool install basic-memory`.

[...]

PR #1380, opened 2026-08-30
docs(mcp): add downloadable MCP tool reference artifact

### LINKED ISSUE #404: [DOCS] create downloadable tool usage doc artifact

FBISiri:
Closes #404

## What

Adds a generated MCP tool reference document (`docs/mcp-tools.md`) and
the script that produces it (`scripts/generate_tool_docs.py`).

## Why

Issue #404 requested a downloadable tool usage documentation artifact.
This PR provides a comprehensive reference for all 20+ registered MCP tools,
auto-generated from source to stay current as tools evolve.

## How

- `scripts/generate_tool_docs.py` uses AST parsing (zero runtime deps) to
  extract tool names, docstrings, parameters, types, and defaults from
  `src/basic_memory/mcp/tools/`
- Only documents tools registered in `__all__` (no internal helpers)
- Idempotent: running twice produces identical output
- To regenerate: `uv run scripts/generate_tool_docs.py`

## Testing

- Ran the generator; verified all registered tools appear in output
- Verified idempotency (second run produces zero diff)
- No existing tests affected (new files only)

### DIFF
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -1,0 +1,626 @@
+# Basic Memory MCP Tool Reference
+
+> **Auto-generated** by `scripts/generate_tool_docs.py`. Do not edit by hand.
+>
+> Regenerate with: `uv run scripts/generate_tool_docs.py`
+
+This reference documents all **21** MCP tools registered by Basic Memory. Each entry lists the tool's purpose and its parameters (types, whether they are required, defaults, and descriptions).
+
+## Table of Contents
+
+- [Note Management](#note-management)
+  - [`delete_note`](#delete_note)
+  - [`edit_note`](#edit_note)
+  - [`move_note`](#move_note)
+  - [`write_note`](#write_note)
+- [Reading & Navigation](#reading--navigation)
+  - [`build_context`](#build_context)
+  - [`list_directory`](#list_directory)
+  - [`read_content`](#read_content)
+  - [`read_note`](#read_note)
+  - [`recent_activity`](#recent_activity)
+  - [`view_note`](#view_note)
+- [Search](#search)
+  - [`fetch`](#fetch)
+  - [`search`](#search)
+  - [`search_notes`](#search_notes)
+- [Project & Workspace Management](#project--workspace-management)
+  - [`create_memory_project`](#create_memory_project)
+  - [`delete_project`](#delete_project)
+  - [`list_memory_projects`](#list_memory_projects)
+  - [`list_workspaces`](#list_workspaces)
+- [Schema Tools](#schema-tools)
+  - [`schema_diff`](#schema_diff)
+  - [`schema_infer`](#schema_infer)
+  - [`schema_validate`](#schema_validate)
+- [Diagnostics](#diagnostics)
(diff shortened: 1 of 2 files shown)


phernandez:
Thanks @FBISiri — this is the right shape for #404 and I'd like to land it.

To get it through CI and the review bot quickly I've cherry-picked your commit as-is (authorship preserved) onto an in-repo branch and added one follow-up commit addressing the three Codex findings above: #1385.

- The `search_notes("my-project", "keyword")` argument-order finding is really a bug in the `search_notes` docstring in `src/basic_memory/mcp/tools/search.py` (the generator just faithfully reproduced it), so the fix lands there and the reference inherits it.
- Anchor disambiguation now follows GitHub's `-1`/`-2` suffix rule in document order.
- Headings embedded in a docstring are nested one level under the tool heading.

I'll close this PR once #1385 merges so the credit stays attached to your commit. If you'd rather push the fixes here yourself, say so and I'll hold off.

phernandez:
Update, and a change of plan from my earlier comment — sorry for the churn.

While finishing #1385 we realized this overlaps with work already in the repo: Basic Memory has a per-tool manual written as Unix-style man pages (#952 — a `search-notes(3)` page for every MCP tool, with verified examples) and `bm man install` from #610. The maintainers' call is to fold documentation generation into that manual rather than keep a separate flat reference, so #1385 is going to draft instead of merging.

What carries forward from your work:

- The Codex review of your PR exposed two real bugs in the `search_notes` tool description (argument order in all 27 examples, and an escaped-quote example that rendered as invalid Python). Those are fixed in #1386, which credits this PR.
- Your extraction approach and the "regenerate and diff in CI" drift gate are going into the manual's registry generator (see the roadmap in `docs/manual-pages.md`).

Thanks for the contribution — it moved #404 forward even though the file itself won't land in this shape. If you're interested in the manual-side generator, that issue will be linked from #610 once it's filed.

FBISiri:
Thanks for the detailed write-up — no apology needed for the churn,
this is a cleaner outcome than merging a file that'd drift from the
manual.

Good to hear the `search_notes` bugs landed as real fixes in #1386.
That alone makes the extraction work worth doing.

The registry generator feeding man pages is a better home for this
than a standalone reference — same idea, less surface area to
maintain. I'd be interested in picking up a piece of that once the
issue is filed from #610. Happy to start with something small on the
generator side.

phernandez:
The manual-side follow-up is up as #1389 (bundled section-3 pages + `memory://man` resources + `bm man <topic>`), and your docstring fixes are in #1386. Thanks again.

phernandez:
Closing in favor of the manual-based approach: your docstring fixes landed via #1386 and the generator idea lives on as the manual's registry generator (#610 / #1389). Thanks @FBISiri — the review your PR triggered ended up fixing bugs well beyond docs.

Issue #1387, opened 2026-08-30
[FEATURE] support for remote projects in hermes plugin


devlsh:
## Feature Description
Allow the Hermes Basic Memory plugin to connect to a self-hosted remote Basic Memory project through its Streamable HTTP MCP endpoint.

## Problem This Feature Solves
The Hermes plugin currently supports local projects and Basic Memory Cloud, but cannot use a Basic Memory project hosted on another service or machine. This prevents Hermes from sharing an existing self-hosted project with other clients.

## Proposed Solution
Add a remote mode that accepts a Basic Memory MCP endpoint URL.

Example configuration:

```json
{
  "mode": "remote",
  "url": "https://memory.example.com/mcp"
}
```

The plugin would use the remote endpoint for its existing memory tools, retrieval, and capture behavior.

## Alternative Solutions
Current workarounds are maintaining a custom plugin patch or running a separate local Basic Memory project for Hermes. Both add operational complexity or prevent clients from sharing one authoritative project.

## Additional Context
Basic Memory already supports Streamable HTTP transport. This request is specifically for exposing that transport as a supported connection mode in the Hermes plugin.

## Impact
This would let Hermes use centrally hosted, self-managed Basic Memory projects and share the same knowledge base with other MCP clients without requiring Basic Memory Cloud.

phernandez:
Superseded by #1475, which tracks the same remote Hermes user story using the current Hermes native MCP runtime and host-owned transport lifecycle.

PR #1397, opened 2026-08-31
feat(hermes): support an existing Basic Memory HTTP daemon

scarabone:
## Summary

- add an explicit `server_url` mode to the Hermes provider using the MCP SDK's streamable HTTP transport
- preserve the existing local stdio and cloud behavior when `server_url` is absent
- require an explicit project for URL mode and keep provider prefetch, tools, per-turn capture, session summaries, and slash-command routing
- resolve the active Hermes home/profile when discovering provider configuration

## Why

Hermes users may already have one long-running Basic Memory daemon serving a shared vault. The current local provider mode always starts its own `bm mcp` child, so it cannot safely attach to that daemon and use provider-managed recall/capture without introducing a second process.

With this change, a provider can use:

```json
{
  "server_url": "http://127.0.0.1:8766/mcp",
  "project": "main"
}
```

URL mode does not look up/install `bm`, create or register projects, mutate Basic Memory configuration, or spawn a subprocess.

## Reliability and safety

- validates HTTP(S) URLs and rejects credentials, queries, and fragments
- requires HTTPS for non-loopback hosts
- reconnects once after HTTP transport failures
- retries only read/idempotent operations; ambiguous mutations reconnect for the next call but are never replayed
- closes the MCP session before the HTTP transport and fails closed on invalid reconfiguration
- supports both two-item and three-item stream tuples returned by MCP SDK versions in the supported range

## Verification

- `295 passed, 17 skipped`
- targeted Ruff checks pass
- read-only integration suite against a running streamable HTTP daemon: `3 passed, 2 explicitly gated fixtures skipped`

### DIFF
--- a/integrations/hermes/CHANGELOG.md
+++ b/integrations/hermes/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **Existing Streamable HTTP server mode.** Set `server_url` to connect the
+  provider to a running Basic Memory MCP endpoint (for example,
+  `http://127.0.0.1:8766/mcp`) without installing/spawning `bm` or creating a
+  local project. Existing tool, prefetch, capture, cleanup, reconnect, and
+  per-call project-routing behavior is retained. A reconnect retries only
+  read/idempotent operations; ambiguous note mutations are never replayed.
+
 ## [0.3.2] — 2026-05-23
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


phernandez:
Thank you, @scarabone, for putting this together and for demonstrating the remote MCP use case so thoroughly. The underlying feature request is valid and remains tracked in #1387.

Since this PR was opened, newer Hermes releases have added a native plugin-to-MCP bridge through `PluginContext.call_mcp()`, backed by Hermes's shared MCP transport, authentication, reconnect, trust, and circuit-breaker machinery. That gives us a substantially smaller direction: configure Basic Memory as a normal Hermes MCP server, let the memory provider use the host-owned connection for automatic recall and capture, and expose the complete Basic Memory MCP tool surface through normal discovery.

We've captured that updated implementation plan and its Hermes prerequisite in #1475.

Because this changes the transport ownership and removes most of the actor-based implementation, we're going to close this PR rather than ask you to keep revising it into a materially different design. Your work was useful in validating the user story and surfacing the lifecycle cases the replacement needs to cover. We appreciate the contribution.

PR #1405, opened 2026-08-31
fix(mcp): normalize constrained project names

### LINKED ISSUE #1388: [BUG] case sensitivity issue with `--project`
## Bug Description
Basic Memory 0.23.2 normalizes a configured project name to lowercase but compares it case-sensitively with the `--project` constraint used by the MCP indexing and watch services.

For example, a project configured as `ExampleProject` is stored or resolved as `exampleproject`, while `--project ExampleProject` remains mixed-case. The comparison fails, leaving the project without indexing or file-watching tasks even though the service remains healthy.

## Steps To Reproduce
1. Install Basic Memory `0.23.2`.
2. Configure a local project using a mixed-case name such as `ExampleProject`.
3. Start the MCP server with `bm mcp --project ExampleProject --transport streamable-http`.
4. Add files to the configured project directory.
5. Inspect the service logs and try to retrieve the files through MCP.

## Expected Behavior
The project constraint should resolve consistently regardless of how Basic Memory normalizes the configured project name.

The project should be indexed and watched successfully.

## Actual Behavior
The configured project is normalized to `exampleproject`, but the MCP constraint remains `ExampleProject`.

The service stays healthy, but it creates no indexing tasks and repeatedly logs:

`No projects to watch`

Files in the project are therefore unavailable through MCP.

Using the lowercase identifier `exampleproject` consistently restores indexing and file watching.

## Environment
- OS: Linux container
- Python version: Unknown; provided by the official container image
- Basic Memory version: `0.23.2`
- Installation method: Official Basic Memory Docker image
- Claude Desktop version (if applicable): Not applicable

## Additional Context
This appears related to the project-name casing problems previously addressed in #127 and the `--project` watcher filtering introduced or corrected through #434 and #758.

The current workaround is to use the same lowercase project identifier everywhere, including the project configuration, default project, and `--project` argument.

## Possible Solution
Normalize the `--project` constraint using the same canonicalization applied to configured and stored project names before comparing them.

Alternatively, resolve the constraint through the project lookup API rather than comparing project-name strings directly.

mikemikimike:
## Summary

- normalize `BASIC_MEMORY_MCP_PROJECT` before selecting startup indexing projects
- apply the same normalized matching to the local watch service
- add regression coverage for mixed-case constraints in both paths

## Root cause

Project names are normalized to permalinks during configuration reconciliation, but the MCP constraint was compared directly with the persisted project name. A mixed-case constraint therefore selected no project, leaving startup indexing and file watching idle while reporting `No projects to watch`.

## Validation

- `uv run --frozen pytest tests/index/test_watch_service.py tests/services/test_initialization.py -q`
- `uv run ruff format --check ...`
- `uv run ruff check ...`
- `uv run pyright ...`
- `git diff --check`

Closes #1388.

### DIFF
--- a/src/basic_memory/index/watch_service.py
+++ b/src/basic_memory/index/watch_service.py
@@ -34,6 +34,7 @@ from basic_memory.index.local_watch import (
 from basic_memory.index.storage_events import StorageEventIndexRuntime
 from basic_memory.models import Project
 from basic_memory.repository import ProjectRepository
+from basic_memory.utils import generate_permalink
 
 
 class WatchEvent(BaseModel):
@@ -175,7 +176,10 @@ class WatchService:
             projects = await self.project_repository.get_active_projects(session)
 
         if self.constrained_project:
-            projects = [project for project in projects if project.name == self.constrained_project]
+            constrained_permalink = generate_permalink(self.constrained_project)
+            projects = [
+                project for project in projects if project.permalink == constrained_permalink
+            ]
 
         skipped = [
             project.name
--- a/src/basic_memory/services/initialization.py
+++ b/src/basic_memory/services/initialization.py
@@ -19,6 +19,7 @@ from basic_memory.models import Project
 from basic_memory.repository import (
     ProjectRepository,
 )
+from basic_memory.utils import generate_permalink
 
 if TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


phernandez:
Thanks @mikemikimike — this fix is now integrated in #1471 and merged to main. Your signed commit was preserved with authorship and cherry-pick provenance.

PR #1417, opened 2026-08-31
fix(core): let edit_note metadata set a note's type

sammywachtel:
## What this changes

`_METADATA_IDENTITY_FIELDS` in `src/basic_memory/services/note_preparation.py` drops
`title`, `type` and `permalink` from an `edit_note` metadata merge. This removes `type`
from that set and leaves the other two alone.

## Why

The comment on the constant explains the exclusion:

```python
# title/type/permalink already have dedicated resolution paths in
# prepare_edit_entity_content (H1 title reconciliation, permalink resolver). Letting a
# metadata merge touch them would race with those paths and could be silently reverted.
```

That is exactly right for two of the three, and I want to be clear that I am not
proposing to weaken it:

- **`title`** is read out of frontmatter and then handed to
  `reconcile_prepared_edit_title_from_h1(...)`, which can replace it with the note's H1.
  A merged title really can be overwritten.
- **`permalink`** is read and then passed through `resolve_permalink(...)`, which
  suffixes on collision. A merged permalink really can come back as something else.

`type` does not share that property. Its entire handling in
`prepare_edit_entity_content` is:

```python
if "type" in content_frontmatter:
    note_type = _coerce_to_string(content_frontmatter["type"])
```

A plain read, with nothing downstream that can disagree. The `note_type` that reaches
the entity row comes from `_build_entity_fields`, which re-parses the merged markdown
(`entity_markdown.frontmatter.type`) — so writing `type` into the frontmatter *is* how
you feed that path, not a race against it. The one other place the local `note_type`
travels is `_build_frontmatter_markdown(...)` on the permalink branch, where
`resolve_permalink` only ever looks at `frontmatter.permalink` and ignores the type
entirely.

So the guard is correct for title and permalink and over-broad by one member.

## The behavior before this change

`edit_note(..., metadata={"type": "decision"})` returned a successful edit and left the
note's type as it was. Nothing in the response distinguished that from a type change
that worked, so the only way to notice is to read the note back afterwards and compare.
Setting the type through `metadata` is the natural thing to reach for, since every other
frontmatter field works that way.

## Tests

- `test_merge_metadata_into_markdown_writes_type` — the merge writes `type` while still
  dropping `title` and `permalink` from the same payload.
- `test_prepare_edit_entity_content_metadata_sets_note_type` — the prepared write carries
  the new type in both the markdown frontmatter and `entity_fields.note_type`, with title
  and permalink unmoved.
- `test_edit_note_metadata_sets_note_type` (integration) — end to end through the MCP
  tool: the type lands in the file, and `search_notes(note_types=["decision"])` finds the
  note, so the index agrees with the file.
- The existing `..._metadata_ignores_identity_fields` tests, unit and integration, keep
  their `title` and `permalink` hijack attempts and keep asserting those are refused. I
  dropped only the `type` key from their payloads and added an assertion that an
  unsupplied type is still left alone.

All three new tests fail against the current `main` behavior and pass with the change.

Docs updated in the same commit: the `edit_note` tool docstring, the `EditEntityRequest`
comment in `schemas/request.py`, and the `metadata` bullet in `man3/edit-note(3).md`.

## How I tested

- `just test-unit-sqlite` — 5735 passed, 43 skipped.
- `just test-int-sqlite` — 501 passed, 16 skipped.
- Postgres: `tests/services/test_entity_service_prepare.py` and
  `test-int/mcp/test_edit_note_integration.py` with `BASIC_MEMORY_TEST_POSTGRES=1`, all
  passing. This change touches only frontmatter merging, so I ran the affected suites
  against Postgres rather than the whole matrix.
- `just fix`, `just format`, `just typecheck` — clean, apart from four pre-existing
  `unresolved-import` diagnostics for `pymilvus`, which lives in the optional `milvus`
  extra that a plain `uv sync` does not install. They are present on an unmodified
  checkout too.

Three failures in `tests/cli/test_cloud_promo.py` on my machine are environmental: those
tests construct a real `ConfigManager()` with no isolation, so they read
`~/.basic-memory/config.json`, and mine already records the promo as shown. They fail
identically on a clean checkout of `main` here. Nothing to do with this change, and I
left them alone.

### DIFF
--- a/src/basic_memory/man/man3/edit-note(3).md
+++ b/src/basic_memory/man/man3/edit-note(3).md
@@ -68,7 +68,8 @@ permalink, or memory:// URL — there is no fuzzy fallback for edits.
   heading of any level and preserves subsections
 - **metadata** — dict of frontmatter fields merged in alongside any operation;
   given keys overwrite or add, other keys and the body are untouched.
-  `title`, `type`, and `permalink` are ignored; keys cannot be deleted
+  `title` and `permalink` are ignored; `type` is applied like any other
+  frontmatter field; keys cannot be deleted
 - **project** / **project_id** / **workspace** — routing; same semantics as
   [[write-note(3)]]
 
--- a/src/basic_memory/mcp/tools/edit_note.py
+++ b/src/basic_memory/mcp/tools/edit_note.py
@@ -437,9 +437,9 @@ async def edit_note(
         metadata: Optional dict of frontmatter fields to merge, independent of `operation`.
             Provided keys overwrite existing frontmatter values (or are added if new);
             unrelated frontmatter keys and the note body are left untouched. Can be
-            combined with any operation in the same call. `title`, `type`, and `permalink`
-            are ignored since those have their own dedicated handling. Key deletion is
-            not supported.
+            combined with any operation in the same call. `title` and `permalink` are
+            ignored since those have their own dedicated handling; `type` is applied like
+            any other frontmatter field. Key deletion is not supported.
         output_format: "text" returns the existing markdown summary. "json" returns
             machine-readable edit metadata.
         context: Optional FastMCP context for performance caching.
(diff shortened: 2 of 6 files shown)


phernandez:
Finished and landed via #1472. The replacement preserves @samwachtel’s signed original commit, authorship, and cherry-pick provenance, then adds the remaining edit-note type validation, auto-create precedence, BOM-safe canonical frontmatter handling, and explicit MCP integration coverage. Thank you for the contribution.

PR #1466, opened 2026-09-04
fix(cli): validate local sync path before saving config

### LINKED ISSUE #1441: project add --local-path can write an unloadable config and brick every later bm command
## Summary

`bm project add --cloud --local-path <path under a file>` persists a config entry whose path cannot be loaded. Every subsequent `bm` command then dies with "failed to load config" — including the commands a user would reach for to undo it.

Found while implementing #1414 (PR #1440), flagged rather than fixed there because it is a different class from the recovery-path findings that PR was addressing.

## Cause

The config save happens **before** the `mkdir`. So an invalid `--local-path` is written to disk first, and the failure that would have revealed it comes second — by which time the bad entry is already persisted.

## Why it matters more than it looks

The failure is not scoped to the project that was added. A config that cannot load takes out the whole CLI, so the user has no in-tool route back: they cannot list projects, cannot remove the bad one, and cannot inspect what happened. The recovery is a manual edit of the config file, which the error message does not mention.

## Direction

Validate the path before persisting anything, or persist only after the directory exists. Given the ordering is the defect, the second is likely simpler and also fixes any other post-save step that can fail on a path.

Worth checking at the same time:

- Whether config loading should degrade rather than hard-fail on one bad entry, so a single unusable project cannot make the tool unusable. That is arguably the deeper fix — the ordering bug is one way to reach the state, and a hand-edited config is another.
- Whether the same save-before-validate ordering exists on other paths that write config entries.

## Verify whether this is pre-existing

PR #1440 restructured the post-persist error handling in this command. Establish whether the save-before-mkdir ordering predates that change — the reporter's reading is that it does, but it should be confirmed against `main` before the issue is scoped, since that determines whether this is a regression to fix urgently or a long-standing trap.

dawNotPoi:
Fixes #1441.

## What changed

- Create the cloud project local sync directory before mutating or saving its routing entry.
- If directory creation fails, leave the on-disk config unchanged and direct the user to repair the path and rerun project add, which adopts the already-created remote project.
- Add regression coverage proving the failed add does not persist the project and a fresh `bm config list` remains usable.

Degraded loading of already-invalid or hand-edited configs remains out of scope.

## Testing

- 32 passed, 5 skipped across the focused project-add suites
- Ruff check and format check
- Changed-file ty check
- Basic Memory doctor

## AI assistance

Codex was used for implementation, testing, and independent code review. This Draft is pending human review.

### DIFF
--- a/src/basic_memory/cli/commands/project.py
+++ b/src/basic_memory/cli/commands/project.py
@@ -805,15 +805,16 @@ def _abort_after_project_created(
     step: str,
     remedy: str,
     error: Exception,
+    retry_add_after_repair: bool = False,
 ) -> NoReturn:
     """Report a failure that happened after the project row became durable.
 
-    Two things are always true here and stated once: the project was created, and
-    re-running `bm project add` will refuse with "already exists". What must vary
-    is the remedy — it has to name a command that both fits the project's mode and
-    can actually repair the step that failed. A single fixed remedy pointed cloud
-    projects at `bm project index`, which the local-only reindex path rejects
-    outright (#1440 review).
+    The default contract states that the project was created and a blind re-run
+    will refuse with "already exists". When ``retry_add_after_repair`` is true,
+    the failed step prevented local registration, so repairing it makes re-running
+    the command safe: the existing adoption path finishes registration instead of
+    creating a second project. In either case, the remedy must fit the project's
+    mode and repair the step that actually failed (#1440 review).
 
     The project is deliberately not rolled back: the row is the user's, and
     deleting it to tidy up an error message would discard what they asked for.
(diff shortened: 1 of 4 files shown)


phernandez (approved):
Reviewed the current head f3c3a0be. The ordering is correct: the local sync directory is validated before the routing entry is persisted, while the already-created remote project remains recoverable through the existing add/adoption path. Focused project-add coverage passes (32 passed, 5 skipped), and just fast-check passes. No blocking code findings from my review. This still needs the contributor CLA and draft status cleared before it can enter the merge gate.

phernandez:
Thanks for the focused fix. I reviewed the ordering change and recovery path locally. The invalid-path regression exercises real filesystem/config behavior and confirms a subsequent config list remains usable. Confirmed save-before-mkdir predates #1440. Local validation: 32 passed, 5 skipped across the three project-add suites; full typecheck, targeted Ruff, and doctor passed. Local codex review against main found no actionable regressions. Keeping degraded loading of already-invalid configs out of scope.

@codex review

PR #1481, opened 2026-09-05
fix(mcp): round-trip find note-type predicates

### LINKED ISSUE #1428: find --meta results display a note_type value that matches nothing
## The trap

`find --meta note_type=X` aliases to the frontmatter `type:` field, which holds the **authored** casing (`Chapter`). But the result row displays the **snake-cased search-index projection** (`chapter`). So the output shows you a value that, fed back into the same query, returns zero rows — silently, exit 0.

Verified against a live project:

```
$ bm find --meta 'note_type=chapter' --fields chapter_number --plain
exit=0                                       # <- zero rows, no error

$ bm find --meta 'note_type=Chapter' --fields chapter_number --plain
chapters/Chapter 1.md	{"chapter_number":"1"}
...
exit=0

$ bm find --meta 'note_type=Chapter' --json | jq '.results[0].metadata'
{"note_type": "chapter"}                     # <- the value that matches nothing
```

Same shape for multi-word types: `LiteraryDevice` is authored, `literary_device` is displayed.

## Why it matters

An agent reading its own results learns the wrong value and writes a query returning nothing. Zero rows with exit 0 is indistinguishable from "no notes match" — the failure never announces itself. This is the same silent-wrong-answer class as the wrong-project writes (#1415) and the never-indexed empty reads (#1414).

## Options

1. Display the value that queries (authored `type`) rather than the index projection.
2. Match case-insensitively on `note_type` predicates, so both spellings work.
3. Normalize the predicate the same way the projection is normalized, so `chapter` matches `Chapter`.

(2) or (3) seem kinder than making callers care about which representation they are looking at; (1) alone still leaves the two spellings behaving differently.

Found while verifying the literary-analysis skill's documented commands (#1424), where every `note_type` example was written from the displayed value and therefore returned nothing.

https://claude.ai/code/session_014pmKq6bqCi6Zp6BTHuZjrp

phernandez:
## Why

Fixes #1428. `find --meta note_type=Chapter` returned a row displaying `chapter`, but feeding that value back returned no rows. The same mismatch affected `LiteraryDevice` / `literary_device`.

## What Changed

- Route string equality and all-string `in` predicates for `type` / `note_type` through the existing normalized note-type search filter.
- Keep other metadata predicates and file-path scoping intact.
- Document the matching behavior and add three clear real CLI integration tests, without database or platform mocks.

## Implementation Details

The shared CLI/MCP find translation extracts eligible type predicates into `SearchQuery.note_types`. Existing search normalization and legacy spelling compatibility apply on both backends; no new SQL, migrations, canonical note rewrites, or normalization rules are introduced. Null, numeric, range, and array predicates retain their existing metadata comparisons.

## Testing

- With the translation disabled, the two round-trip tests reproduce the exact bug: returned `chapter` and `literary_device` values produce empty second queries.
- `uv run pytest test-int/cli/test_find_note_type_integration.py --no-cov -q`: 3 passed on SQLite.
- `BASIC_MEMORY_TEST_POSTGRES=1 uv run pytest test-int/cli/test_find_note_type_integration.py --no-cov -q`: 3 passed on real Postgres.
- `just fast-check`: passed lint, formatting, and full typecheck.
- `just doctor`: passed.
- `codex review --uncommitted`: no actionable defects; independently ran the three new integration tests successfully.
- Broader existing CLI/MCP suite is running; full CI will validate this head.

## Risks / Follow-ups

Type string equality/membership now intentionally accepts normalized equivalents. Ordinary metadata keys remain case-sensitive. The public metadata-filter API itself is unchanged; this fixes the find translation shared by CLI and MCP.

### DIFF
--- a/src/basic_memory/man/man1/find(1).md
+++ b/src/basic_memory/man/man1/find(1).md
@@ -122,6 +122,13 @@ mis-spelled multi-character operator: `status==active`, `status=>active` and
 and `>3`. An unquoted value may therefore not begin with `=`, `<` or `>`;
 quote one that genuinely does, as in `range=">=5"`.
 
+String equality and `in` predicates on `type` or `note_type` use the same
+normalization as note-type search: `Chapter` and `chapter` match the same
+notes, as do `LiteraryDevice` and `literary_device`. A `metadata.note_type`
+value returned by find can be reused in its next `--meta` predicate.
+Other metadata keys remain case-sensitive; null, numeric, range, and array
+predicates keep their existing frontmatter comparison behavior.
+
 ## OPTIONS
 
 - **--name** — file-name glob, e.g. `"*.md"`; omitted matches everything.
--- a/src/basic_memory/mcp/tools/posix_tools.py
+++ b/src/basic_memory/mcp/tools/posix_tools.py
@@ -949,6 +949,9 @@ async def find(
             with "=" — the other operators compare against the value, and a
             comparison with null is never true. Keys accept dot-paths
             ("review.approved"); "note_type" aliases the frontmatter "type" key.
+            String equality and "in" on either type spelling use normalized
+            note-type matching ("LiteraryDevice" matches "literary_device").
+            Other metadata keys and operators retain their usual comparisons.
             Any other operator fails fast naming the supported set.
         fields: Frontmatter fields to return per hit (dot-paths allowed), e.g.
             ["title", "priority"]. Requires `meta`. A field missing on a hit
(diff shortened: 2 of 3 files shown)


phernandez:
Broader local validation completed: 403 passed across the new round-trip tests, existing search metadata integration test, MCP POSIX tests, and CLI POSIX tests. Command: uv run pytest test-int/cli/test_find_note_type_integration.py test-int/cli/test_search_notes_meta_integration.py tests/mcp/test_tool_posix.py tests/cli/test_cli_posix_verbs.py --no-cov -q . The three new tests also pass on real Postgres; doctor and static checks pass.

PR #1494, opened 2026-09-07
fix(core): report resolved schema identity to validation observers

groksrc:
Schema validation observers currently receive the covered type and the authored schema reference, neither of which identifies the exact schema used. When several schemas cover one type, or an explicit reference misses and validation falls through to type lookup, Cloud's SPEC-87 Mission 5 cannot reliably attribute the passing result and skips credit.

This change carries the selected source and inline/named kind through schema resolution and adds required `schema_external_id` and `schema_kind` fields to `ValidatedNoteOutcome`. Named schemas report the selected schema note's stable ID; inline schemas report the validated note's ID. Single-note, type-wide, and all-types validation use the same resolution adapter. The existing definition-only resolver delegates to the same selection logic, preserving resolution order and parser errors. The sourced resolver and its types are exported from `basic_memory.picoschema`. Public validation response models remain unchanged.

This is the Core prerequisite for Mission 5. Cloud must adopt the merged Core commit and consume the new identity fields before removing its repeated schema lookup and ambiguous-match skip. Cloud mission locks, rollout flags, rewards, and database schemas are unchanged. Existing observers can keep reading the old fields; code constructing outcomes must supply the two new required fields.

Validation:
- 203 SQLite Picoschema/API tests pass, including 17 identity API regressions and 9 source-resolution cases.
- 17 PostgreSQL identity API tests pass after the final changes; the broader 48 schema API tests passed before the review follow-up.
- 241 schema, MCP, CLI, and integration tests passed before the review follow-up.
- Mixed batches contain two distinct named schemas and an inline schema, with both passing and failing notes. Reusing the first note's schema ID failed both type-wide and all-types regressions; both passed after restoration.
- Deliberately swapped the selected schema source to the other candidate: the API regression failed on the mismatched validated field; all 15 identity regressions passed after restoring the implementation.
- `just fast-check`, `just doctor`, and `git diff --check` pass.

Started from Core `26439122a2c656916a39dd55cb44516aaf78869f`. This work can merge independently of Cloud #1993.

### DIFF
--- a/src/basic_memory/api/v2/routers/schema_router.py
+++ b/src/basic_memory/api/v2/routers/schema_router.py
@@ -255,7 +255,7 @@ async def validate_schema(
                     schema_reference=schema_ref if isinstance(schema_ref, str) else None,
                     passed=response.passed,
                     schema_external_id=resolved.source.external_id,
-                    schema_kind="inline" if isinstance(schema_ref, dict) else "named",
+                    schema_kind=resolved.kind,
                 )
             )
 
@@ -509,7 +509,7 @@ async def _validate_note_entities(
                         schema_reference=schema_ref if isinstance(schema_ref, str) else None,
                         passed=response.passed,
                         schema_external_id=resolved.source.external_id,
-                        schema_kind="inline" if isinstance(schema_ref, dict) else "named",
+                        schema_kind=resolved.kind,
                     )
                 )
 
--- a/src/basic_memory/api/v2/routers/schema_router.py
+++ b/src/basic_memory/api/v2/routers/schema_router.py
@@ -34,7 +34,13 @@ from basic_memory.schemas.schema import (
     DriftFieldResponse,
     TypeValidationSummary,
 )
-from basic_memory.picoschema.resolver import SchemaSearchFn, resolve_schema
+from basic_memory.picoschema.resolver import (
+    ResolvedSchema,
+    SchemaCandidate,
+    SchemaSearchFn,
+    resolve_schema,
+    resolve_schema_with_source,
+)
 from basic_memory.picoschema.parser import SchemaDefinition
 from basic_memory.picoschema.validator import validate_note
 from basic_memory.services.schema_validation_hooks import (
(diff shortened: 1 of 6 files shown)


groksrc:
## Review findings

Reviewed against `origin/main` `26439122` (the base this PR names). No blocking defects — the resolver refactor is behavior-preserving and the identity fields are correct in every resolution path I exercised, including a `schema:` value that is neither dict nor str (falls through to type lookup, correctly reports `named` with `schema_reference=None`).

Verification I ran beyond reading the diff:

- Mutated `schema_external_id=resolved.source.external_id` → `entity.external_id`: 9 failures, exactly the named cases, 6 inline/400 cases still green.
- Mutated `candidates[0]` → `candidates[-1]` in both resolver branches: 4 failures, including `test_invalid_selected_schema_does_not_skip_to_another_source`.
- `tests/api/v2 tests/picoschema tests/schemas` → 665 passed on a clean tree.
- `just typecheck` reports 4 diagnostics, all `unresolved-import: pymilvus` in an unrelated test — a missing optional extra in my worktree, nothing on the changed files.

Three findings, none of them merge blockers.

---

```text
medium | tests/api/v2/test_schema_validation_identity.py:109 | test oracle quality | the batch path this PR exists to fix is never tested with more than one note
```

**Why:** Every observer test in the repo asserts `len(outcomes) == 1` — the three in `test_schema_router.py` and the new one here. The parametrization covers all three scopes and all four reference kinds, but always with a single validated note, so nothing pins per-note attribution across a batch. That is precisely the Mission 5 usage: validate-all, then credit each note against the schema it actually resolved to. A regression that crossed attribution between notes in `_validate_note_entities` — a hoisted resolution, a stale `resolved` reused across loop iterations — would pass the entire suite today.

To be clear, this is a coverage gap and not a bug: I wrote the missing test (three notes — two named refs to *different* schemas plus one inline) and it passes against this branch. Each note is attributed to its own schema, the two named ids are distinct, and the inline note is self-attributed.

**Fix:** Add a multi-note batch case. Happy to push the test I wrote onto the branch if you want it.

---

```text
medium | src/basic_memory/api/v2/routers/schema_router.py:258, :512 | knowledge duplication | schema_kind is re-derived by the caller instead of reported by the resolver
```

**Why:** `resolve_schema_with_source` already knows which branch it took — line 99 is the inline decision — but discards it and returns only `definition` and `source`. The router then reconstructs the same rule with `"inline" if isinstance(schema_ref, dict) else "named"`, duplicated verbatim at both construction sites. That is one rule encoded three times across two layers, and the two copies furthest from the decision are the ones the observer actually sees.

It is correct in every branch today; I checked. The risk is drift, not a present defect. If resolution order ever changes — inline with a named fallback, or named overriding inline — the resolver changes and the router's guess silently disagrees, producing a well-formed outcome with the wrong `schema_kind`. Nothing would fail. This is the same class of ambiguity the PR is removing from `schema_entity`/`schema_reference`, so it is worth closing while the API has exactly one consumer.

**Fix:** Carry the kind on `ResolvedSchema` next to `source`, set it at each `return` in `resolve_schema_with_source`, and have both call sites read `resolved.kind`.

---

```text
low | src/basic_memory/picoschema/__init__.py:41 | dependency direction | the new resolution API is absent from the package's public surface
```

**Why:** `__all__` exports `resolve_schema`, but `resolve_schema_with_source`, `SchemaCandidate`, and `ResolvedSchema` are not exported, so Cloud has to reach past the package into `basic_memory.picoschema.resolver` to adopt the seam — while its older, less capable sibling is package-public. The asymmetry reads as "this one is internal," which is the opposite of the intent for a Core prerequisite Cloud is meant to consume.

**Fix:** Add the three names to `picoschema/__init__.py`, or leave a one-line note that the sourced API is deliberately submodule-only.

---

Two notes for the Cloud side, not findings:

- Nothing in `basic-memory-cloud` imports `schema_validation_hooks`, `ValidatedNoteOutcome`, or `SchemaValidationObserver` today, so the new required fields break nothing on that side. The "can merge independently of Cloud #1993" claim holds.
- For inline schemas `schema_external_id == note_external_id`, so Cloud must branch on `schema_kind` before joining that id against schema notes.

Making the two fields required rather than defaulted is the right call and matches the house fail-fast rule — a default id would let wrong attribution flow silently, which is the exact failure this PR removes.

🤖 Agent-assisted review via [Claude Code](https://claude.com/claude-code)

PR #1510, opened 2026-09-08
perf(core): avoid hashing files before the first index

### LINKED ISSUE #1446: First bm status on an unindexed project checksums every file
## Summary

The status observer walks the project and, when a file has no indexed row to compare against, reads it through `compute_checksum`. On a brand-new or never-indexed project **nothing** is indexed, so `_reuse_indexed_checksum` misses on every file and the first `bm status` reads the entire project from disk.

## Why it is avoidable

The answer for a never-indexed project is "not indexed". That is derivable from the walk alone — the checksums are computed to detect *changes* against indexed rows, and there are none to compare against. The work is spent producing information the caller cannot use.

Suggested: when a project has no indexed rows, report the unindexed state from the walk without checksumming. `project.last_indexed_at` (added in #1440) already distinguishes never-indexed from idle, so the condition is available without inferring it from counts.

## Provenance and scope

Found while implementing #1414 (PR #1440), which fixed the closely related case: `bm project add --no-wait` was calling readiness and paying this same full scan, defeating the flag. That is fixed — `--no-wait` now prints the state rather than measuring it.

This one was left deliberately, because for `bm status` the scan *is* the observation the user asked for. It is an optimization rather than a defect, which is why it is filed separately rather than folded in.

## Worth checking alongside

Whether the same reuse-miss cost appears on other first-run paths, and whether the walk could report a cheap "N files present, not yet indexed" without reading contents — that is the wording #1414 introduced for `project add`, and it would be consistent for `status` to answer the same way at the same cost.

phernandez:
## Why

The first `bm status` on a project with no indexed rows read every file to compute checksums, even though every discovered path was already known to be unindexed.

## What Changed

An observation backed by a confirmed empty index now returns the eligible file paths without reading file contents. Status still reports the file count, `never_indexed`, and all files pending.

## Implementation Details

The local observed-file source reuses its existing indexed-stat query to identify the empty-index case. A source without database evidence continues hashing. Projects with indexed rows retain stat-based checksum reuse, changed-file hashing, and move detection. The indexing batch reader computes checksums when it reads new files, so this also avoids redundant reads before a first index pass.

## Testing

- After preserving file-size metadata: `uv run pytest test-int/cli/test_status_wait_integration.py tests/index/test_local_project_index.py tests/index/test_local_project_scan_parity.py tests/api/v2/test_project_router.py -q --no-cov`: 86 passed; `just fast-check`: passed.

- `uv run pytest test-int/cli/test_status_wait_integration.py tests/index/test_local_project_index.py tests/index/test_local_project_scan_parity.py -q --no-cov`: 54 passed.
- `uv run pytest test-int/cli/test_status_wait_integration.py -q --no-cov`: 2 passed, including readiness assertions.
- `BASIC_MEMORY_TEST_POSTGRES=1 uv run pytest test-int/cli/test_status_wait_integration.py tests/index/test_local_project_index.py -q --no-cov`: 46 passed.
- `just fast-check` and `just doctor`: passed.
- The CLI regression makes any checksum call fail and verifies the actual status response, including hidden-file filtering.

## Risks / Follow-ups

An empty-index observation carries null checksums until indexing reads the content; file sizes remain available from stat metadata. The filesystem count remains a snapshot and converges on subsequent observations if files or indexed rows change concurrently.

Closes #1446.

### DIFF
--- a/src/basic_memory/index/local_project.py
+++ b/src/basic_memory/index/local_project.py
@@ -317,6 +317,12 @@ class LocalProjectIndexObservedFileSource(ProjectIndexObservedFileSource):
             if self.indexed_stat_source is not None
             else {}
         )
+        # With a confirmed empty index, every discovered path is new. Hashes
+        # cannot identify changes or moves yet; the batch reader computes them
+        # when indexing content. Status can therefore count files without reads.
+        if self.indexed_stat_source is not None and not indexed_stats:
+            return tuple(RuntimeObservedIndexFile(path=path) for path in file_paths)
+
         observed_files: list[RuntimeObservedIndexFile] = []
         for file_path in file_paths:
             try:
--- a/src/basic_memory/index/local_project.py
+++ b/src/basic_memory/index/local_project.py
@@ -317,18 +317,15 @@ class LocalProjectIndexObservedFileSource(ProjectIndexObservedFileSource):
             if self.indexed_stat_source is not None
             else {}
         )
-        # With a confirmed empty index, every discovered path is new. Hashes
-        # cannot identify changes or moves yet; the batch reader computes them
-        # when indexing content. Status can therefore count files without reads.
-        if self.indexed_stat_source is not None and not indexed_stats:
-            return tuple(RuntimeObservedIndexFile(path=path) for path in file_paths)
-
         observed_files: list[RuntimeObservedIndexFile] = []
         for file_path in file_paths:
             try:
                 metadata = await self.file_service.get_file_metadata(file_path)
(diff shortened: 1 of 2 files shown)


phernandez:
@codex review

893e3ebdc preserves stat metadata for unindexed files while still skipping content checksums. The API/CLI/indexing regression suite passes 86 tests and just fast-check passes.

phernandez:
@codex review

Head 5c61a7326 fixes the Windows-only regression test expectation: status correctly reports on-disk bytes, including CRLF, so the assertion now compares file stat sizes. Both status integration tests and fast-check pass. Production code is unchanged from the reviewed head.

PR #1513, opened 2026-09-08
fix(mcp): retry gateway rate limits

### LINKED ISSUE #1378: MCP client should honor Retry-After and retry on 429 instead of failing the tool call
## Problem

When the cloud gateway rejects a request with `429`, the MCP client surfaces the error to the model and stops. It does not retry, and it discards the `Retry-After` boundary the server sends specifically so that clients can retry correctly.

A transient, often sub-second rate-limit window therefore becomes a visible task failure. The agent has to notice the error, invent its own backoff, and retry by hand — and many will simply report the failure to the user instead.

Observed in the wild (basic-memory 0.23.2), an agent doing ordinary work hit this on two consecutive calls:

```
Too many requests. Retry after the current rate-limit window resets.
```

It reported to the user:

> Rate-limited on Basic Memory, so I can't yet distinguish "dangling reference" from "read failed".

...worked around it by falling back to local `git` inspection, and succeeded on a manual retry ~35 seconds later. Nothing was actually wrong. A single automatic retry would have made the entire episode invisible.

## Current behavior

`src/basic_memory/mcp/tools/utils.py` has five near-identical call wrappers — `call_get`, `call_put`, `call_patch`, `call_post`, `call_query` — each containing the same block:

```python
# Client errors: log as info except for 429 (Too Many Requests)
if status_code == 429:
    logger.warning(f"Rate limit exceeded: GET {url}: {error_message}")
...
response.raise_for_status()
```

and a generic message in `get_error_message`:

```python
elif status_code == 429:
    return "Too many requests: Please slow down and try again later"
```

So: one log line, then `ToolError`. No retry, no backoff, no use of the retry boundary.

## What the server already sends

The gateway emits everything a client needs in order to retry precisely:

- `Retry-After` (seconds)
- `X-RateLimit-Limit-Tenant`, `X-RateLimit-Remaining-Tenant`, `X-RateLimit-Reset-Tenant`
- on the hosted MCP path, a JSON body carrying `code: "rate_limit_exceeded"`, `retry_after`, and `reset_at`

All of it is currently dropped.

## Proposed fix

1. **Honor `Retry-After` with bounded automatic retry.** On `429`, wait the server-provided interval plus a small jitter — jitter matters because concurrent agents sharing a quota will otherwise retry in lockstep and collide again — then retry. Cap both the attempt count and total wait, so a genuinely exhausted long window fails fast instead of hanging a tool call.
2. **Retrying a 429 is safe for every method, including writes.** The limiter rejects the request before it is processed, so there is no partial work to duplicate. Worth an explicit comment, since the instinct is to restrict retries to idempotent verbs.
3. **Put the boundary in the message when retries are exhausted.** `"Too many requests: Please slow down and try again later"` gives the model nothing to act on. `"Rate limit exceeded. Retry after 43s."` lets an agent schedule around it rather than abandoning the task or hot-looping.
4. **Implement once.** The five wrappers already share this branch by copy-paste; a shared helper (or an httpx transport / event hook) keeps them from drifting.

## Scope

Client-side only. This issue is strictly about the client being a well-behaved consumer of a `429` — honoring a retry boundary the server already publishes.

Whether the underlying quotas are the right shape is a separate question and is deliberately out of scope here.

dawNotPoi:
## Why

The shared MCP API helpers currently surface a gateway `429` immediately, even when the server
provides a valid `Retry-After` boundary and a later attempt would succeed. This exposes transient
Basic Memory gateway rate limits to every MCP operation instead of honoring the gateway's
pre-processing rejection contract.

## What Changed

- Route all six Basic Memory MCP HTTP wrappers (`GET`, `PUT`, `PATCH`, `POST`, `DELETE`, and
  `QUERY`) through one rate-limit retry helper.
- Honor valid `Retry-After` delay-seconds and HTTP-date values, with at most three total attempts,
  positive jitter, and a 30-second cumulative sleep budget.
- Retry writes only when their request bodies are replayable. Streaming content and file uploads
  remain one-shot.
- Preserve the final server detail and `Retry-After` boundary when retrying is skipped, exceeds
  the wait budget, or exhausts all attempts.

## Implementation Details

- Retries are limited to HTTP `429` responses with a valid standard `Retry-After` header; missing
  or malformed headers, transport failures, timeouts, and non-429 responses are not retried.
- The server delay is never shortened to fit the local budget. A next attempt is skipped when its
  server delay plus jitter would exceed the remaining 30-second cumulative sleep budget.
- Each completed 429 response is closed before sleeping. Cancellation continues to propagate,
  and each attempt retains the caller's existing HTTP timeout.
- Existing public wrapper signatures, request parameters and workspace headers, telemetry
  attributes, and ordinary error paths—including `PATCH`—are preserved.

## Testing

- `uv run pytest --no-cov -q tests/mcp/test_tool_utils_rate_limit.py` — **26 passed in 0.24s**
  on final HEAD `c004638a1d713be8517409fa8af6c19437942f84`.
- `COVERAGE_FILE=/home/admindawn/auto_pr/repros/basic-memory-1378/.coverage-focused-final/.coverage uv run coverage run --branch -m pytest -q tests/mcp/test_tool_utils.py tests/mcp/test_tool_utils_cloud_auth.py tests/mcp/test_tool_telemetry.py tests/mcp/test_tool_utils_rate_limit.py`
  — **65 passed in 12.41s**. A line/arc comparison against base
  `c50ed118809e0ffff5bfb2e65698f05913148ae2` found all **78/78 added executable lines** and
  **20/20 added branch arcs** covered; this is changed-code coverage, not whole-module coverage.
- `just check` — passed on final HEAD (`ruff check`, `ruff format`, and full `ty check`).
- `just doctor` — passed on production-equivalent HEAD `9cfee2136c33452e06d6b04e0f453d2dae23b895`;
  the final commit only adds tests.
- `TZ=UTC RAYON_NUM_THREADS=1 just test` on final HEAD:
  - SQLite unit: **7,223 passed, 49 skipped**.
  - SQLite integration: **552 passed, 15 skipped, 34 deselected**.
  - PostgreSQL unit: **7,204 passed, 68 skipped**.
  - PostgreSQL integration: **557 passed, 10 skipped, 34 deselected, 1,347 warnings in
    1,478.38s**. Its original pytest process continued after the 600-second outer wrapper had
    reported `error: recipe test-int-postgres failed with exit code 1`, and later produced this
    passing summary. All four pytest stages therefore have passing summaries, but the overall
    `just test` command itself exited unsuccessfully and is not reported green.
- FastMCP 4.0.3 compatibility was checked with an isolated overlay containing exactly the three
  packages changed by upstream #1506 (`fastmcp==4.0.3`, `fastmcp-slim==4.0.3`, and
  `uncalled-for==0.4.0`) over the existing locked environment:
  - Both scripts were run from the repository after setting the required overlay environment:

    ```bash
    export PYTHONPATH=/home/admindawn/auto_pr/repros/basic-memory-1378/fastmcp-4.0.3-compat-c004638/overlay:/home/admindawn/auto_pr/repos/basic-memory/src:/home/admindawn/auto_pr/repos/basic-memory/tests
    export BM1378_COMPAT_METADATA=/home/admindawn/auto_pr/repros/basic-memory-1378/fastmcp-4.0.3-compat-c004638/metadata
    export BM1378_COMPAT_OVERLAY=/home/admindawn/auto_pr/repros/basic-memory-1378/fastmcp-4.0.3-compat-c004638/overlay
    export TZ=UTC
    export RAYON_NUM_THREADS=1
    cd /home/admindawn/auto_pr/repos/basic-memory
    ```

  - `.venv/bin/python /home/admindawn/auto_pr/repros/basic-memory-1378/fastmcp-4.0.3-compat-c004638/run_focused_pytest.py`
    — **59 passed, 2 warnings in 0.23s** for the rate-limit and existing utils tests, with the
    overlay versions and loaded module paths asserted in-process.
  - `.venv/bin/python /home/admindawn/auto_pr/repros/basic-memory-1378/fastmcp-4.0.3-compat-c004638/compat_runtime_check.py`
    — in-process FastMCP calls made exactly three HTTP attempts for each of the default
    `ToolError` and `raise_on_error=False` paths, preserving the gateway detail and retry boundary.
  - This was a three-package overlay check, not a complete fresh environment or full-suite run on
    upstream's new lock. The branch's full-suite run uses locked FastMCP 4.0.0b1.
- A prior native Asia/Shanghai MCP suite on implementation HEAD `20999286` produced **1,385 passed
  and 2 failed**; both naive-datetime failures reproduce unchanged on base `c50ed118`.
- One PDF subprocess test initially hit a host Rayon resource-limit error and failed identically on
  base `c50ed118`; the isolated test passed with command-local `RAYON_NUM_THREADS=1`.

## Risks / Follow-ups

- No live model/provider API or external Basic Memory gateway was called. HTTP retry behavior was
  exercised with `httpx.MockTransport` plus the in-process FastMCP client smoke.
- A full suite has not been completed under upstream's FastMCP 4.0.3 lock; the bounded overlay
  result does not replace that validation.
- The four pytest stages produced passing summaries, but the PostgreSQL integration process
  outlived the recipe's 600-second wrapper. Because that wrapper exited with code 1, the overall
  `just test` command is not counted as green despite the later passing child-process summary.

Fixes #1378

### DIFF
--- a/src/basic_memory/mcp/tools/utils.py
+++ b/src/basic_memory/mcp/tools/utils.py
@@ -256,7 +256,7 @@ def _parse_retry_after(response: Response) -> tuple[str, float] | None:
             len(normalized_delay) == len(max_wait_text) and normalized_delay > max_wait_text
         ):
             return raw_retry_after, float("inf")
-        return raw_retry_after, float(int(retry_after))
+        return raw_retry_after, float(int(normalized_delay))
 
     try:
         retry_at = parsedate_to_datetime(retry_after)
--- a/src/basic_memory/mcp/tools/utils.py
+++ b/src/basic_memory/mcp/tools/utils.py
@@ -4,7 +4,12 @@ These functions provide a consistent interface for making HTTP requests
 to the Basic Memory API, with improved error handling and logging.
 """
 
+import asyncio
+import random
 import typing
+from collections.abc import Awaitable, Callable
+from datetime import datetime, timezone
+from email.utils import parsedate_to_datetime
 from typing import Any, Optional
 
 import logfire
@@ -33,6 +38,13 @@ from loguru import logger
 from fastmcp.exceptions import ToolError
 
 from basic_memory.config import ConfigManager
+
+_RATE_LIMIT_MAX_ATTEMPTS = 3
+_RATE_LIMIT_MAX_TOTAL_WAIT_SECONDS = 30.0
+_RATE_LIMIT_JITTER_MIN_SECONDS = 0.05
+_RATE_LIMIT_JITTER_MAX_SECONDS = 0.25
+
+type _RequestCall = Callable[[], Awaitable[Response]]
 
 
 def _classify_http_outcome(status_code: int) -> str:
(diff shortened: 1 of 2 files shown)


phernandez:
Verified current head c004638a1d713be8517409fa8af6c19437942f84 locally: the rate-limit, existing HTTP helper, cloud-auth, and telemetry suites pass together (65 tests). Hosted checks are passing, Codex has completed review of c004638 with a fresh thumbs-up, and the review-thread query is empty. This covers the #1378 milestone work without a duplicate implementation. No live gateway call was made in this verification.

Issue #993, opened 2026-06-13
Use frontmatter to enforce validation and set conditional (read-only / team-scoped) permissions


groksrc:
## Motivation

Basic Memory already treats frontmatter as a first-class control surface: `type` drives schema resolution, and any custom key is captured as `entity_metadata` and indexed. The schema system (Picoschema validation, inference, drift) reads from there too. Two capabilities are a natural extension of that surface and are worth landing — or at least designing the frontmatter contract for — in the v1.0 window, since both touch the write path:

1. **Tighter validation** — make schema validation enforce-able from frontmatter, not just advisory.
2. **Conditional permissions** — let frontmatter mark a note (or type) as read-only / locked so agents can't silently edit or delete it.

Both are "the file is the source of truth" features: the rule travels with the note, survives moves, and works in git without a side database.

## Part 1 — Frontmatter-driven validation

Today validation is advisory (`schema_validate` reports warnings/errors but writes still go through). Proposal:

- A note's schema (inline `schema:` dict, a `schema:` reference, or the implicit type schema) can declare **required fields** and **enforcement level**.
- On `write_note` / `edit_note`, the write path consults that schema. At minimum surface violations back to the agent in the tool result; optionally **reject** writes that violate a schema marked `enforce: true`.
- Keep the default non-breaking (warn, don't block) so existing notes/imports don't start failing; enforcement is opt-in per type/note.

This turns the existing `ValidationReport` machinery into something that can guard the write path instead of only being run on demand.

## Part 2 — Conditional permissions (write/edit protection)

Let frontmatter declare that a note is protected, e.g.:

```yaml
---
title: Release Process
type: runbook
locked: true          # or: permissions: read-only
---
```

When a note is `locked` / `read-only`, `edit_note` and `delete_note` (and overwriting `write_note`) refuse the operation and return a clear, actionable message ("this note is marked read-only in frontmatter; remove `locked: true` to edit"). This is the single most useful case: stable runbooks, canonical references, and decision records that an agent should not casually rewrite.

Open questions to settle in design:
- Note-level vs type-level (schema) declaration — likely support both, note-level wins.
- Whether the lock can be lifted by an agent editing the frontmatter itself (escape hatch vs. genuine guard) — probably a config/flag decision.
- Interaction with imports and `move_note` (move should still be allowed; content edit should not).

## Part 3 — Teams extension (forward-looking, may be a separate issue)

[...]

PR #1521, opened 2026-09-08
fix(mcp): unify note writes with typed outcomes

### LINKED ISSUE #1479: [BUG] write_note(overwrite=True) creates a second note instead of replacing one whose file has been moved
## What happens

`write_note(..., overwrite=True)` creates a second note instead of replacing the existing one, when the existing note's file has been renamed or moved. It reports `# Created note` and no error.

The result is two notes with the same title in the same directory. Everything that reads the note by path, or by listing the directory, keeps reading the moved file, so the edit appears to have been made and changes nothing. The caller is told the write succeeded.

## Reproduction

`main` at `6d6e7fe53c53d2c1d8e034953ef3a82be25251c8` (2026-09-05), SQLite, default config — specifically `update_permalinks_on_move: false`, which is the shipped default. Only public MCP tools are used.

```python
await write_note(title="song-sketching", directory="app/probe",
                 content="# original\nfirst body")

await move_note(identifier="song-sketching",
                destination_path="app/probe/SKILL.md")

await write_note(title="song-sketching", directory="app/probe",
                 content="# replacement\nsecond body", overwrite=True)
```

Literal output:

```
--- STEP 1 write_note (create) ---
# Created note
file_path: app/probe/song-sketching.md
permalink: test-project/app/probe/song-sketching

--- STEP 2 move_note ---
Note moved successfully
song-sketching -> app/probe/SKILL.md
Permalink: test-project/app/probe/song-sketching

--- STEP 3 write_note(overwrite=True) ---
# Created note
file_path: app/probe/song-sketching.md
permalink: test-project/app/probe/song-sketching-1

--- STEP 4 list_directory ---
Contents of '/app/probe' (depth 1):
Page 1 (page size 10, 2 total items)

SKILL.md            app/probe/SKILL.md | song-sketching | id: 85fba0a0-...
song-sketching.md   app/probe/song-sketching.md | song-sketching | id: 0ebd2edd-...

Total: 2 items (2 files)

--- STEP 5 files on disk ---
['SKILL.md', 'song-sketching.md']

--- STEP 6 read_note("app/probe/SKILL.md") ---
---
title: song-sketching
type: note
permalink: test-project/app/probe/song-sketching
---

# original
first body
```

Step 2 is the interesting one. `move_note` was given the bare string `song-sketching` and found the note — the API log records `resolution_method: 'title'`. So resolving this note by title, after the move, already works and is already in the codebase. `write_note` just doesn't ask.

Step 6 is the part that costs a user something: the note anyone actually reads still holds the original body. The replacement went into a file nothing points at.

## Cause

Traced on the commit above.

1. `write_note` does an optimistic create and treats a 409 as "exists, switch to update" (`src/basic_memory/mcp/tools/write_note.py`). The 409 originates in `prepare_create_entity_content`, from `file_service.exists(file_path)` (`src/basic_memory/services/note_preparation.py:391`). `file_path` is derived from the title and directory, not looked up by identity.
2. After the move, `app/probe/song-sketching.md` does not exist, so there is no 409, so the `overwrite` branch is never entered at all. `overwrite=True` and `overwrite=False` produce the same thing here.
3. The new note's permalink then collides with the moved note's, which still holds `test-project/app/probe/song-sketching` because `update_permalinks_on_move` defaults to false. The suffix loop in `resolve_permalink` (`note_preparation.py:274-285`) appends `-1`.

That loop already guards the neighboring case, and its comment names it:

```python
# A case-only rename resolves to the slug the entity already holds;
# suffixing it would churn `config` -> `config-1` on every move (#1281).
if owner is None or owner == current_file_path:
    break
```

This case slips past because on the create branch there is no `current_file_path` to compare against — the caller never identified an existing row.

The suffixing itself may well be intentional, and I am not arguing against it. Two notes cannot hold one permalink, and `-1` is a reasonable way to break the tie once you have decided to create a second note.

What does not look defensible is that `overwrite=True` silently does not overwrite.

- **The parameter's whole job is this decision.** `#818` established that an explicit `overwrite=True` must not be dropped, and `write_note`'s own docstring says it replaces an existing note. Here the flag is read, resolved against the config default, and then never consulted, because the code path that consults it is gated on a filesystem check that the move defeats.
- **The end state the caller asked for is already reachable.** `move_note` resolved the same note from the same title one call earlier. Nothing about the note is unfindable; the create path just uses a different key.
- **There is no signal.** No error, no warning, no changed wording in the response — the tool says `# Created note` in both the ordinary create case and this one. `#1077` made the same complaint about a log-only warning that MCP callers never see; here there isn't even a log line, because `detect_file_path_conflicts` looks for filename-convention variants of the same title and a rename to an unrelated name is not one.
- **A duplicate wins nothing.** If the goal were to protect the existing note, refusing would do that and would tell the caller. Creating a shadow copy protects the note while hiding the fact that it was protected.

## Existing issues I checked

- `#860` (merged) fixes the resolution *inside* the overwrite branch — resolve the conflicting entity by `file_path` strictly rather than by fuzzy permalink. Different problem: it assumes the 409 fired. Here it does not.
- `#1077` / `#1078` cover a duplicate created when the existing file uses a different filename convention for the same title (kebab-case vs Title Case). Same symptom, different trigger — that one is about how the path is derived, this one is about the file having been moved out from under the derived path. `#1078` was closed unmerged.
- `#1281` is the case-only rename the suffix guard above was written for.
- `#1326` is directory-name case resolution.
- `#818` is the opposite symptom: `overwrite=True` refusing a write it should have made.
- `#1259` proposes semantic dedup before a write, which is a different question from identity.

I could not find this reported. Apologies if I missed it.

## Two ways to fix it, and I would rather ask than assume

**A — replace by identity.** Before taking the create branch, look the note up the way `move_note` already does. If a note with that title exists in that directory under a different filename, and `overwrite=True`, update it where it now lives.

Sharper for the caller, but it means `write_note` can write to a path it was not given. If someone moved a note deliberately, a later write silently following it may not be what they wanted either.

**B — refuse, and say where it went.** Detect that a note with this title and directory exists at a different path, and return an error naming that path, in the shape of the existing `_format_overwrite_error` message. The caller decides: write to the moved file, or create a genuinely new note.

Safer and much smaller. Costs a round trip in the common case, where following the move is probably what was meant.

There is a plausible third position — that the duplicate is correct because the title-and-directory pair no longer names that note once it has been renamed — but then `overwrite=True` should say so rather than reporting a plain create.

I have not written a patch. Happy to, once you say which behavior you want; I did not want to arrive with a design.

phernandez:
## Why

`write_note` inferred an existing-note conflict from exception text, then resolved an identifier and attempted a second write. That masked locked-note failures and made moved-note behavior depend on the general link resolver. This consolidates #1525 into one explicit write operation.

Fixes #1479. Related to #993.

## What Changed

- A path-addressed write returns typed created, updated, already-exists, target-moved or locked outcomes. The MCP tool consumes validated variants with an exhaustive match.
- The exact current file owner is the overwrite target. When the path is vacant and a supplied frontmatter permalink or generated path alias identifies a moved note, return `NOTE_PATH_CONFLICT` with its current path. `overwrite=False` can deliberately create a separate note.
- Preserve actual replacement refusals and canonical content. Remove exception-text matching and the MCP resolver round trip; no general LinkResolver changes are needed.
- Document moved-note behavior and retain the real MCP/CLI lock and moved-note regressions.

## Implementation Details

`NoteContentMutationService.write_note` owns target selection. Internal frozen dataclasses carry the outcomes; `/knowledge/write` serializes them as a Pydantic discriminated union; `KnowledgeClient` validates that response once.

The new operation dispatches through public create/update methods so runtime overrides retain their pre-acceptance work. Those methods share actor resolution, transaction ownership, acceptance hooks, cache invalidation and post-commit publication. Existing create/update HTTP contracts and the core accepted-write guards remain intact. Legacy structured service errors are translated once into recoverable outcomes; other refusals retain their status and detail.

## Testing

- `uv run pytest tests/mcp/test_tool_write_note.py tests/mcp/test_tool_write_note_kebab_filenames.py tests/mcp/test_tool_write_note_metadata.py test-int/mcp/test_write_note_integration.py test-int/mcp/test_moved_note_overwrite.py test-int/mcp/test_locked_note_results.py test-int/cli/test_locked_note_refusals.py --no-cov -q` — 121 passed.
- `uv run pytest test-int/test_note_write_outcomes.py tests/indexing/test_accepted_note_mutation_runner.py tests/cloud/test_cloud_services.py --no-cov -q` — 65 passed before the final runtime-override regression additions.
- `uv run pytest test-int/test_note_write_outcomes.py test-int/mcp/test_moved_note_overwrite.py --no-cov -q` — final 9 passed, including runtime override refusals and transaction-hook rollback.
- PostgreSQL: the write/MCP/CLI integration suite passed 29 tests. `BASIC_MEMORY_TEST_POSTGRES=1 uv run pytest test-int/test_note_write_outcomes.py test-int/mcp/test_moved_note_overwrite.py test-int/mcp/test_locked_note_results.py --no-cov -q` passed all 15 final runtime-dispatch cases.
- Review regression: custom-frontmatter moved-note cases failed in both output formats before the fix. `uv run pytest tests/mcp/test_tool_write_note.py test-int/test_note_write_outcomes.py test-int/mcp/test_moved_note_overwrite.py test-int/mcp/test_locked_note_results.py --no-cov -q` then passed 81 tests; the added malformed-frontmatter API case passed separately. `BASIC_MEMORY_TEST_POSTGRES=1 uv run pytest test-int/mcp/test_moved_note_overwrite.py --no-cov -q` passed all 5 cases.
- CI telemetry correction: the full Python 3.14 SQLite unit run had one failure (7,132 passed): its old create-endpoint span assertion. Added API/client spans for the new write operation and updated the telemetry contract. `uv run pytest tests/mcp/test_tool_telemetry.py tests/api/v2/test_knowledge_router_telemetry.py test-int/test_note_write_outcomes.py --no-cov -q` passed all 15 cases.
- `just fast-check` and `just doctor` passed.

## Risks / Follow-ups

Cloud must deploy the Core version containing `/knowledge/write` before clients using this endpoint are released. Existing HTTP create/update callers remain compatible. This is not a Cloud deployment or a change to the concurrency protocol. PR #1525 is superseded by this implementation and its behavioral tests.

### DIFF
--- a/src/basic_memory/api/v2/routers/note_write_router.py
+++ b/src/basic_memory/api/v2/routers/note_write_router.py
@@ -1,0 +1,118 @@
+"""One path-addressed write, with expected outcomes preserved across HTTP."""
+
+from dataclasses import asdict
+from typing import Annotated, assert_never
+
+from fastapi import APIRouter, HTTPException, Path
+
+from basic_memory.deps import (
+    AppConfigDep,
+    EntityRepositoryV2ExternalDep,
+    EntityVectorSyncSchedulerDep,
+    NoteContentMaterializationProviderDep,
+    NoteContentMutationServiceDep,
+    ProjectExternalIdPathDep,
+    ProjectRepositoryDep,
+    RelationResolutionSchedulerDep,
+    SessionMakerDep,
+)
+from basic_memory.runtime.note_content import runtime_note_content_payload_as_dict
+from basic_memory.schemas.v2.entity import EntityResponseV2
+from basic_memory.schemas.v2.note_write import (
+    NoteAlreadyExists,
+    NoteCreated,
+    NoteLocked,
+    NoteTargetMoved,
+    NoteUpdated,
+    WriteNoteRequest,
+    WriteNoteResponse,
+)
+from basic_memory.services.note_content_writes import (
+    NoteContentMutationServiceError,
+    note_content_mutation_error_from_rejection,
+)
+from basic_memory.services.note_write_outcomes import (
+    AlreadyExists,
+    Created,
+    Locked,
+    Rejected,
+    TargetMoved,
+    Updated,
+)
+from basic_memory.utils import build_permalink_resolution_candidates
+from basic_memory.workspace_context import current_workspace_permalink_context
+
+router = APIRouter()
+
+
+@router.post("/write", response_model=WriteNoteResponse)
+async def write_note(
+    data: WriteNoteRequest,
+    project_id: ProjectExternalIdPathDep,
+    project_external_id: Annotated[str, Path(alias="project_id")],
+    project_repository: ProjectRepositoryDep,
+    entity_repository: EntityRepositoryV2ExternalDep,
+    session_maker: SessionMakerDep,
+    note_content_mutation_service: NoteContentMutationServiceDep,
+    note_content_materialization_provider: NoteContentMaterializationProviderDep,
+    vector_sync_scheduler: EntityVectorSyncSchedulerDep,
+    relation_resolution_scheduler: RelationResolutionSchedulerDep,
+    app_config: AppConfigDep,
+) -> WriteNoteResponse:
+    """Create at the requested path or replace the note that currently owns it."""
+    # Release the read connection before the mutation opens its own transaction.
+    async with session_maker() as session:
+        project = await project_repository.get_by_id(session, project_id)
+    if project is None:
+        raise HTTPException(status_code=404, detail="Project not found")
+    workspace = current_workspace_permalink_context()
+    candidates = build_permalink_resolution_candidates(
+        data.note.file_path,
+        project.permalink,
+        include_project=app_config.permalinks_include_project,
+        workspace_permalink=(
+            workspace.workspace_slug if workspace and workspace.should_prefix_permalinks else None
+        ),
+    )
+    try:
+        outcome = await note_content_mutation_service.write_note(
+            project_external_id=project_external_id,
+            data=data.note,
+            overwrite=data.overwrite,
+            entity_repository=entity_repository,
+            permalink_candidates=candidates,
+            user_profile_id=None,
+            source="api",
+        )
+    except NoteContentMutationServiceError as error:
+        raise HTTPException(status_code=error.status_code, detail=error.detail) from error
+    match outcome:
+        case Created(change=change) | Updated(change=change):
+            accepted = await note_content_materialization_provider.materialize_write_change(change)
+            entity = EntityResponseV2.model_validate(
+                runtime_note_content_payload_as_dict(accepted.payload)
+            )
+            # Runtime-injected schedulers preserve local and Cloud publication behavior.
+            if app_config.semantic_search_enabled:
+                vector_sync_scheduler.schedule_entity_vector_sync(
+                    entity_id=entity.id, project_id=project_id
+                )
+            relation_resolution_scheduler.schedule_relation_resolution(project_id=project_id)
+            match outcome:
+                case Created():
+                    return NoteCreated(entity=entity)
+                case Updated():
+                    return NoteUpdated(entity=entity)
+                case _:
+                    assert_never(outcome)
+        case AlreadyExists(file_path=file_path):
+            return NoteAlreadyExists(file_path=file_path)
+        case TargetMoved(note=note):
+            return NoteTargetMoved(**asdict(note))
+        case Locked(message=message):
+            return NoteLocked(message=message)
+        case Rejected(rejection=rejection):
+            error = note_content_mutation_error_from_rejection(rejection)
+            raise HTTPException(status_code=error.status_code, detail=error.detail)
+        case _:
+            assert_never(outcome)
--- a/src/basic_memory/api/v2/routers/note_write_router.py
+++ b/src/basic_memory/api/v2/routers/note_write_router.py
@@ -3,6 +3,7 @@
 from dataclasses import asdict
 from typing import Annotated, assert_never
 
+import logfire
 from fastapi import APIRouter, HTTPException, Path
 
 from basic_memory.deps import (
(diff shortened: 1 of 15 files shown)


phernandez:
@codex review

src/basic_memory/mcp/tools/write_note.py
```diff
@@ -470,31 +470,27 @@ async def write_note(
...
-                        action = "Updated"
-                    except Exception as update_error:  # pragma: no cover
-                        # Re-raise the original error if update also fails
-                        raise e from update_error  # pragma: no cover
+                    # Report the attempted replacement's failure, not the create conflict
+                    # that selected this path (for example, a locked-note refusal).
+                    if not entity.permalink:
+                        raise ValueError(
+                            "Entity permalink is required for updates"
+                        )  # pragma: no cover
```
phernandez: should we remove this permalink check?

phernandez: Removed in 735301d8a. The comment now describes strict file-path resolution followed by update using the returned UUID; no permalink prerequisite is needed. Replacement failures propagate normally. Validation: 71 focused tests, fast-check, and doctor passed.

phernandez:
@codex review

phernandez:
@codex review

src/basic_memory/api/v2/routers/note_write_router.py
```diff
@@ -0,0 +1,118 @@
...
+) -> WriteNoteResponse:
+    """Create at the requested path or replace the note that currently owns it."""
+    # Release the read connection before the mutation opens its own transaction.
+    async with session_maker() as session:
+        project = await project_repository.get_by_id(session, project_id)
+    if project is None:
+        raise HTTPException(status_code=404, detail="Project not found")
+    workspace = current_workspace_permalink_context()
+    candidates = build_permalink_resolution_candidates(
+        data.note.file_path,
```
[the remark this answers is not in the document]
phernandez: Fixed in 1b02c1ceb. Reproduced both text and JSON calls creating custom/song-1 after moving the original; those cases now return the typed TargetMoved outcome and preserve the original bytes. The service checks an explicit content-frontmatter permalink before generated candidates only when the exact path is vacant. This follows preparation semantics: metadata.permalink is discarded by schema_to_markdown, so it does not introduce a separate identity source. Validation: 81 focused SQLite cases, 5 PostgreSQL moved-note cases, the malformed-frontmatter API regression, and fast-check passed.

phernandez:
@codex review

phernandez:
@codex review

PR #1535, opened 2026-09-11
fix(mcp): report never-indexed projects instead of a silent empty search

### LINKED ISSUE #1534: search_notes returns an empty result set on a never-indexed project, indistinguishable from "no such note"
## Summary

On a project that has never been indexed, `search_notes` returns an empty result set for **every** query — including a stopword-grade term like `"the"` — and says nothing about the index being empty. From inside an MCP session that is indistinguishable from an honest "no such note."

That matters because "search before creating" is the standing rule in most agent workspaces. When search cannot fail loudly, that check passes *vacuously*: the agent searches, sees nothing, and concludes the note does not exist. It is also the state of **every fresh install before the first index pass**, so it is the first thing a new user hits.

Reported by a user running a ~24k-file local knowledge base across a bunch of agent workspaces. One of their machines had `bm` installed and never indexed, and it was the machine most of the workspaces booted from. Neither they nor their agents noticed for weeks — the agents just grepped the tree and carried on.

## Reproduce

1. Install Basic Memory on a machine with notes on disk, and never run an index pass (`last_indexed_at IS NULL`).
2. From an MCP session, call `search_notes(query="the", project="<project>")`.
3. Result: `results: []`, `total: 0`, with the standard "No results found for 'the' … Try broader or different terms" copy.
4. Expected: something that names the state — files are present, the index was never built.

## Why it happens

`_format_search_markdown` in `src/basic_memory/mcp/tools/search.py:377-400` deliberately assumes an empty result means "no match for this query," not "empty knowledge base":

```
# Empty search is usually "no match for this query," not "empty knowledge base," so we
# do not repeat the first-note offer here (that would nag established users). Point at
# recent_activity, which owns the getting-started guidance when the base is truly empty.
```

That assumption is right for an indexed project and wrong for a never-indexed one, and nothing on the path distinguishes the two. The zero-results branch at `search.py:1480-1486` only logs and returns the empty response as normal.

The handoff to `recent_activity` does not rescue it either: `recent_activity` does not consult readiness, so on a never-indexed project it is empty for the same reason. The two surfaces agree, and are both wrong together.

## The signal already exists

#1414 added exactly what is needed and wired it into the CLI only:

- `Project.last_indexed_at` (`models/project.py:76-82`) — "the one durable bit that separates 'never indexed' from 'indexed and idle'. Every other readiness signal is a count, and a count of zero cannot tell 'nothing to do' from 'nothing was ever started'."
- `ProjectIndexPhase.NEVER_INDEXED` and `ProjectIndexReadiness.describe()` (`schemas/project_readiness.py:112-143`), which already renders `"N files present, not yet indexed — run 'bm project index <name>'"`.
- `report_project_readiness` (`cli/commands/command_utils.py:102`) renders it for `bm project add` / `bm status`.

`grep -rn "last_indexed_at" src/basic_memory/mcp/` returns nothing. The MCP surface — the one agents actually read through — never asks.

This is the same class of bug as #1414, one surface over: #1414 fixed vacuous readiness on the CLI, this is vacuous *negative results* on MCP.

## Suggested fix

When a search returns zero results, check project readiness before formatting the miss. If the phase is `NEVER_INDEXED`, say so instead of reporting a negative — reuse `ProjectIndexReadiness.describe()` so the wording cannot drift from `bm status`. Roughly:

> Search returned no results because project '<name>' has never been indexed (N files present on disk). This is not a negative result — run `bm project index <name>`.

Notes on scope:

- The readiness read should only happen on the **zero-result** path, so an indexed project pays nothing.
- Worth applying the same guard to `recent_activity`, since search currently delegates its "truly empty" guidance there.
- Structured (`output_format="json"`) callers need the signal too, not just the markdown string — an agent parsing JSON sees `results: []` with no hint. A field on the response (e.g. `index_phase`) beats prose for that path.
- Adjacent but distinct prior art, for whoever picks this up: #1184 (`reindex --embeddings` silently no-ops on an unsynced project), #1236 (vector search silently degrades with no ready manifest rows), #1320 (zero-results reporting `total_is_exact: true`). Same theme — a search path that cannot distinguish "nothing matched" from "nothing was searched."

Filed at `b04d1b6d`.

wangzhengzhuo05:
## What

On a project that has never been indexed (`project.last_indexed_at IS NULL`), the MCP tools reported an ordinary empty result that is indistinguishable from an honest "no match" on an indexed project:

- `search_notes` returned `results: []` plus `No results found for '<q>' in project '<p>'. Try broader or different terms, ...`
- `recent_activity` (project-scoped) was empty and said nothing about the index.

Both are vacuous negatives: files exist on disk, but nothing was ever searched. This defeats the "search before creating" convention agents rely on, and it is the state of every fresh install before the first index pass.

## Why

The signal already exists — `ProjectIndexPhase.NEVER_INDEXED` and `ProjectIndexReadiness.describe()` — and the CLI already uses it (`bm status`, `bm project add`). Nothing under `src/basic_memory/mcp/` read `last_indexed_at`, so the MCP surface — the one agents actually read through — never asked.

## How

- `src/basic_memory/mcp/tools/search.py`: on the **zero-result path only**, read readiness through the existing typed client (`ProjectClient.get_status`). When the phase is `NEVER_INDEXED`, `_format_search_markdown` returns an honest sentence that reuses `ProjectIndexReadiness.describe(...)` (so it cannot drift from `bm status`). The structured path carries `index_phase` in the returned payload.
- `src/basic_memory/mcp/tools/recent_activity.py`: the same guard for the project-scoped empty page, replacing the "you have no notes yet" onboarding (which would misread "never looked" as "no notes") with the never-indexed state.
- If readiness cannot be read, both paths fall back to today's copy — this is a reporting courtesy, not a hard failure.
- No schema, router, client, or CLI changes; no new dependencies. An indexed project pays nothing (the readiness read happens only on the empty path).

## Tests

Added 3 tests (`tests/mcp/test_tool_search.py`, `tests/mcp/test_tool_recent_activity.py`) that set up a project whose `last_indexed_at` is `None` with a note on disk:

- `search_notes` text output names the never-indexed state and the `bm project index` remedy
- `search_notes(output_format="json")` carries `index_phase == "never_indexed"`
- `recent_activity` text reports the never-indexed state

Results:

```
uv run pytest tests/mcp/test_tool_search.py tests/mcp/test_tool_recent_activity.py -q --no-cov
88 passed, 2 failed
```

Both failures (`test_search_with_date_filter`, `test_recent_activity_format_relative_time_and_truncate_helpers`) are **pre-existing on clean `main`** — confirmed by stashing this change and re-running them (same 2 failures), and they are unrelated to the touched code.

`uv run ruff check` and `uv run ruff format --check` pass on all changed files.

## Mutation check

Disabling the readiness read (`index_readiness = None`) makes all 3 new tests fail, and restoring it makes them pass — the tests genuinely guard the fix.

## Notes

- Commits carry a DCO `Signed-off-by` line.
- This is my first PR here, so the `license/cla` check will need my one-time CLA acceptance from the GitHub bot link; the DCO check should pass on its own.

Fixes #1534

### DIFF
--- a/tests/mcp/test_tool_recent_activity.py
+++ b/tests/mcp/test_tool_recent_activity.py
@@ -695,3 +695,53 @@ async def test_recent_activity_entity_rows_include_external_id(client, test_grap
     assert any(re.search(uuid_pattern, line) for line in entity_lines), (
         f"entity rows missing external_id: {entity_lines!r}"
     )
+
+
+@pytest.mark.asyncio
+async def test_recent_activity_never_indexed_says_so(
+    client, test_project, session_maker, config_home
+):
+    """A never-indexed project must not report an ordinary empty activity feed (#1534)."""
+    from sqlalchemy import text as sa_text
+
+    from basic_memory import db
+
+    # Trigger: files exist on disk but no index pass has ever completed.
+    notes_dir = config_home / "notes"
+    notes_dir.mkdir(exist_ok=True)
+    (notes_dir / "unindexed-note.md").write_text("# Unindexed Note\n\nNot yet indexed.\n")
+    async with db.scoped_session(session_maker) as session:
+        await session.execute(
+            sa_text("UPDATE project SET last_indexed_at = NULL WHERE id = :id"),
+            {"id": test_project.id},
+        )
+
+    result = await recent_activity(project=test_project.name, timeframe="7d")
+
+    assert isinstance(result, str)
+    assert "never been indexed" in result
+    assert "bm project index" in result
+
+
+@pytest.mark.asyncio
+async def test_recent_activity_indexed_empty_keeps_onboarding(
+    client, test_project, session_maker, config_home
+):
+    """An indexed-but-quiet project keeps the original onboarding copy (#1534)."""
+    from datetime import datetime
+
+    from sqlalchemy import text as sa_text
+
+    from basic_memory import db
+
+    async with db.scoped_session(session_maker) as session:
+        await session.execute(
+            sa_text("UPDATE project SET last_indexed_at = :now WHERE id = :id"),
+            {"now": datetime.now(), "id": test_project.id},
+        )
+
+    result = await recent_activity(project=test_project.name, timeframe="7d")
+
+    assert isinstance(result, str)
+    assert "No recent activity" in result
+    assert "never been indexed" not in result
--- a/tests/mcp/test_tool_search.py
+++ b/tests/mcp/test_tool_search.py
@@ -2281,3 +2281,112 @@ def test_search_notes_parse_str_list_rejects_non_string_list_elements_in_place()
     # All-string lists still work correctly.
     assert parse_str_list(["note", "task"]) == ["note", "task"]
     assert parse_str_list(["note,task"]) == ["note", "task"]
+
+
+@pytest.mark.asyncio
+async def test_search_never_indexed_text_says_so(client, test_project, session_maker, config_home):
+    """A never-indexed project must not report an ordinary empty result (#1534)."""
+    from sqlalchemy import text as sa_text
+
+    from basic_memory import db
+
+    # Trigger: files exist on disk but no index pass has ever completed.
+    notes_dir = config_home / "notes"
+    notes_dir.mkdir(exist_ok=True)
+    (notes_dir / "unindexed-note.md").write_text(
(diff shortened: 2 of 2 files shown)


wangzhengzhuo05:
Superseded by #1545, which preserves the signed-off commit and extends it with implementation/test adaptations. Closing to keep the queue clean — thanks @phernandez.

PR #1542, opened 2026-09-13
fix(sync): treat a markdown path with a non-markdown type as dirty

### LINKED ISSUE #1538: [BUG] Checksum-keyed reconciliation can never repair an entity row whose type is wrong — sync reports it “unchanged” forever
## Bug Description

If an `entity` row's *type* fields are wrong but its `checksum` matches the bytes on disk, no amount of re-syncing or re-indexing will repair it. Change detection compares checksums only, so the row is reported `unchanged` and is never re-parsed. There appears to be no mechanism anywhere that notices a row whose **type** disagrees with its file.

## How we hit it

A markdown note's entity row was replaced by a plain-file row — `note_type='file'`, `content_type='text/plain'`, `permalink=NULL`, no `note_content` row, title taken from an Emacs autosave (filed separately as #1537).

The important part for *this* issue is the reason it was unrepairable. The autosave's bytes were byte-identical to what the note became 16 seconds later, so the corrupt row's `checksum` and `size` already matched the real file exactly. Every subsequent scan therefore said:

```
[CHANGE] Move detection: found 0 moved files
[CHANGE] Delete detection: found 0 deleted files
[CHANGE] Detection complete: 0 new, 0 modified, 0 deleted, 0 moved, N unchanged
```

Repeated incremental rebuilds changed nothing. This has now happened twice, weeks apart, and on the first occurrence the rebuild attempts were abandoned as ineffective without the cause ever being found.

## Steps To Reproduce

1. Obtain an `entity` row whose `file_path` is an existing `.md` note, but with `note_type='file'`, `content_type='text/plain'`, `permalink=NULL`, no `note_content` row — and a `checksum` that matches the file on disk. (Either via #1537, or by hand.)
2. Run any sync / rebuild, however many times.
3. Observe the row reported as `unchanged`; `edit_note <permalink>` still does not resolve.
4. Now append a single byte to the file, and it repairs itself completely.

## Expected Behavior

A sync or rebuild should be able to repair a row that disagrees with its own file. At minimum, a row whose `file_path` is a markdown path but whose `content_type` is not `text/markdown` — or a markdown entity with no `note_content` row — should be treated as dirty regardless of checksum.

## Actual Behavior

The row is permanently invisible to sync. The only repair is to change the file's bytes so the checksum differs.

Worth noting that the repair machinery itself is entirely healthy — it just cannot be *triggered*. Once the checksum differed, `index_markdown_file` ran and `_apply_markdown_entity_fields()` rewrote `title`, `note_type`, `permalink` and `content_type` on the **existing** row, in place:

```
title         #note.md#   ->  note
note_type     file        ->  note
content_type  text/plain  ->  text/markdown
permalink     NULL        ->  proj/notes/note
observations  0           ->  N
note_content  absent      ->  present, synced
```

Same entity id, graph intact, no delete and no cascade. A one-byte append did what an unbounded number of rebuilds could not.

## Possible Solution

1. Add a consistency predicate to change detection alongside the checksum comparison — mark a row dirty when `runtime_file_path_is_markdown_note(file_path)` is true but `content_type != 'text/markdown'`, or when a markdown entity has no `note_content` row.
2. Offer a `bm sync --force` / `--repair` that ignores checksums for a project or path. Either of these would have let us fix this without resorting to modifying the user's note to provoke a reindex.
3. Consider a `bm doctor`-style read-only consistency check, since this class of corruption is silent: `read_note` kept working the whole time, so nothing surfaced until a write was attempted.

## Environment

- **OS:** macOS 26.6.2 (Darwin 25.6.0), arm64
- **Python:** 3.14.7
- **basic-memory:** 0.23.2
- **Install method:** pipenv virtualenv; MCP servers launched by Claude Desktop and Claude Code
- **Config:** local SQLite, `semantic_search_enabled: true`, `index_changes: true`, `kebab_filenames: false`

wangzhengzhuo05:
## What

Fixes #1538. A row whose `file_path` is a markdown note but whose indexed `content_type` is not `text/markdown` can never be repaired, because change detection compares checksums only: the row's checksum already matches the file on disk, so every scan reports it `unchanged` and it is never re-parsed.

Reported shape of the corrupt row (an Emacs autosave took over the real note's `file_path`):

```
file_path     notes/note.md
note_type     file
content_type  text/plain
permalink     NULL
note_content  no row
checksum      matches the bytes on disk exactly
```

Appending a single byte to the file repairs the row completely, which shows the repair machinery is healthy — it just cannot be triggered.

## How

`EntityRepository.get_by_file_paths()` already masks an indexed checksum to SQL `NULL` when the indexed projection is known to be incomplete even though the checksum matches (`publication_pending`, `legacy_relation_pending`, `legacy_markdown_identity`). A `NULL` checksum makes change detection re-read the canonical bytes.

This adds one more predicate to that same mask: the path is a markdown note path (`.md` / `.markdown`, case-insensitive) **and** the row's `content_type` is not `RUNTIME_MARKDOWN_CONTENT_TYPE`. Such a row is now treated as dirty regardless of checksum, so the next scan re-reads the file and `_apply_markdown_entity_fields()` repairs `title`, `note_type`, `permalink`, `content_type` and the note-content row in place — the same repair the manual one-byte append provokes.

Scope is deliberately the first acceptance clause of the issue ("a row whose `file_path` is a markdown path but whose `content_type` is not `text/markdown`"). No API, protocol, or row-shape change: `get_by_file_paths()` still returns `(file_path, checksum)`.

## Tests

`tests/repository/test_entity_repository.py::test_get_by_file_paths_masks_markdown_path_without_markdown_type`

- a corrupt row (`notes/note.md`, `content_type="text/plain"`, current checksum) is masked to `None`;
- a healthy row (`notes/note.md`, `content_type="text/markdown"`) keeps its real checksum, guarding against over-masking.

## Verification (local)

```
uv run pytest tests/repository/test_entity_repository.py -q    # 37 passed
uv run pytest tests/indexing/ -q                               # 636 passed, 1 skipped
uv run ruff check / ruff format --check <changed files>        # clean
```

Mutation check: with `markdown_path_without_markdown_type` removed from the mask's `or_(...)`, the new test fails (`1 failed`); restored, it passes. The test therefore guards this exact predicate and not incidental behaviour.

---

*Implemented with AI assistance (OpenCode + muse-spark); reviewed, re-run and mutation-checked by the author before submission.*


wangzhengzhuo05:
Superseded by #1543, which preserves the contributor commit (35c3a5b) and addresses the P2 review finding. Closing to keep the queue clean — thanks @phernandez.

